PR #831, opened 2015-01-06
Add support for Rails 4.2

cbeer:


cbeer:
@jkeck I justed added 7128491 (which uses `#prepend` from Ruby 2.0..). If we want to support Ruby 1.9, we'd probably need to copy/paste those `search_*_url` methods around.

PR #848, opened 2015-01-10
Add square and thumb derivatives for non-repository images.

### LINKED ISSUE #839: Create derivatives from uploaded non-repo images
This should be something that we make configurable by the implementing app (somehow, maybe through a proc or something), but we'll probably want defaults.

My initial thoughts are full (original), thumb (400px long-edge), and square (100x100?).

We also have a large version that we index into our fixtures but I'm not sure we use that anywhere or what the dimensions relative to the original are.

jkeck:
Closes #839 

I have a pending spec about configuring additional derivatives.  I wasn't able to think of a good way to do that after experimenting with a few approaches, but I'm submitting anyway as that seemed like a lower priority.


app/uploaders/configurable_uploader_versions.rb
```diff
@@ -0,0 +1,30 @@
+module ConfigurableUploaderVersions
```
cbeer: I should have noticed this before, but should this and the ItemUploader be Spotlight-namespaced?

jkeck: :+1: wasn't sure if it was a conscious decision for the uploaders or not.  Just namespaced it and will push another commit to namepace the other uploaders.

PR #857, opened 2015-01-14
Show uploaded non-repository images in OSD viewer on record view...

### LINKED ISSUE #838: How do we handle viewing non-repository images on the full record view
We currently attempt to load an OSD instance using indexed IIIF info.

jkeck:
...using legacy image pyramid.

Closes #838 

osd-viewer


cbeer:
:+1:

PR #876, opened 2015-01-16
Update sir-trevor and swap out sir trevor markdown for github markdown

### LINKED ISSUE #864: List markup doesn't work in markdown fields

### LINKED ISSUE #875: Update to Sir Trevor 0.5.0.beta1

cbeer:
@jkeck I think this is ready to go. There's one test that seems to intermittently fail more than others, but if it gets too annoying we can just disable it. Can you give this a thorough manual testing before merging?

Fixes #864 
Fixes #875


app/assets/javascripts/spotlight/compat.js
```diff
@@ -0,0 +1,625 @@
+/**
```
jkeck: Should we put this file in `vendor/assets`?

jkeck:
@cbeer looks pretty good so far.  One thing I noticed is that this branch appears to break the nestable drag-and-drop re-ordering, not only in the sir-trevor widgets but also everywhere else.  It looks like a javascript error is now getting thrown from the nestable library (not really sure how that's even possible, with this PR maybe lo-dash?).

It's getting late so I can't look into it too much more this evening but can help look at it next week.

cbeer:
Ok, I think that's fixed up now.  Let me know if you see any more trouble.

app/assets/javascripts/spotlight/compat.js
cbeer: I'm not sure. Happy to, but it's our own assembly of shims for our use of phantomjs, not the full upstream distribution of polyfills.

PR #879, opened 2015-01-17
Add admin-togglable published flag for controlling exhibit access

### LINKED ISSUE #392: Administrator needs to restrict exhibit access
We have feedback that some exhibit-builders will want their exhibit to be private or restricted in some way. 

Use case: a librarian and a faculty member collaborate to create an exhibit for a course. The exhibit is intended to be used just for the course and the faculty member doesn't want anyone outside of the course to have access. This might not be a common use case until we add user personalization features (annotating, user galleries, commenting, other potential end user personalization things) but when Spotlight does have those, restricted access will definitely be a need. I think there are probably other use cases even without user personalization features involved.

cbeer:
screen shot 2015-01-16 at 4 35 41 pm

Fixes #392

### DIFF
--- a/app/controllers/spotlight/concerns/application_controller.rb
+++ b/app/controllers/spotlight/concerns/application_controller.rb
@@ -13,7 +13,15 @@ module Spotlight::Concerns
       helper Spotlight::ApplicationHelper
 
       rescue_from CanCan::AccessDenied do |exception|
-        redirect_to main_app.root_url, :alert => exception.message
+        if current_exhibit and !can? :read, current_exhibit
+          # Try to authenticate the user
+          authenticate_user!
+  
+          # If that fails (and we end up back here), offer a 404 error instead
+          raise ActionController::RoutingError.new('Not Found')
+        else
+          redirect_to main_app.root_url, :alert => exception.message
+        end
       end
     end
   end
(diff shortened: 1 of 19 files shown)


app/controllers/spotlight/concerns/application_controller.rb
```diff
@@ -13,7 +13,11 @@ module ApplicationController
       helper Spotlight::ApplicationHelper
 
       rescue_from CanCan::AccessDenied do |exception|
-        redirect_to main_app.root_url, :alert => exception.message
+        if current_exhibit and !can? :read, current_exhibit
+          redirect_to main_app.new_user_session_url, :alert => exception.message
```
jkeck: I think this will be problematic for logged in users who are not admins/curators of the current group.  It will try to log them in, and then once it realizes they're already logged in it will try to redirect them back to the place that raised the `AccessDenied` error.

cbeer: Hm. I guess we could just give those users a 403 and call it good.

cbeer: Or just a 404, since that's baked into Rails.

jkeck: Either that or redirect them to the home page with a stern warning.

cbeer: We can't send them home here. If the exhibit is unpublished and they don't have access, they can't see the home page either.

jkeck: Heh, I was kinda joking (hence the stern warning part).

Anyhow, this ends up being problematic if it is the default exhibit that is unpublished because a logged in user could get into the situation where they get to an error page and are never able to navigate back to their exhibit unless they know the URL by heart.

I understand that this is really an application concern, and they can decide what the correct behavior is for them in this case, but it may be something we should document down the road.

PR #881, opened 2015-01-17
Make configured non-repository item upload fields editable

### LINKED ISSUE #841: Titles of non-repository items should be editable
The way we have things setup we are mapping the title that is submitted to the configured title field.  However; this isn't a field that the user can edit once they've uploaded the file.

I guess a related question would be if these objects should be delete-able from the UI.

jkeck:
Closes #841

edit-uploaded-fields


ggeisler:
:+1:

app/views/spotlight/catalog/_edit_default.html.erb
```diff
@@ -5,7 +5,7 @@
       <% current_exhibit.custom_fields.each do |field| %>
         <div class="form-group">
           <%= d.label field.field, field.label %>
-          <%= d.text_field_without_bootstrap field.field, value: document[field.field], class: 'form-control' %>
+          <%= d.text_field_without_bootstrap field.field, value: @document[field.field], class: 'form-control' %>
```
cbeer: I'm not sure why we're using `@document` instead of just `document`. If it works, I think we should prefer the local variable.

jkeck: :+1: yeah, I noticed inconsistent use in that file already.  I'll push up a change to make them all local vars.

PR #891, opened 2015-01-22
Store the exhibit slug in solr documents, and optionally filter on it

### LINKED ISSUE #883: Solr Documents and Exhibit Multitenancy.
Should an item uploaded to an exhibit (or an imported repository item) be exclusive to that Exhibit?

Currently if I were to upload several items to my exhibit they would be available for other exhibits within the same Spotlight instance (and the same is true w/ items imported from the repository).

This isn't really an issue in the exhibits.stanford.edu context (for now at least), but probably will be for other consumers given that we have some multitenancy features.

This won't be technically too difficult to achieve, but it then begs the question does there need to be a way to import items into a spotlight instance that would be available to all exhibits (which is how it is now), and if so, how do we present the distinction.

### LINKED ISSUE #893: Improve resource indexing machinery
When indexing a Spotlight::Resource, we should:

 - parallelize the operation (if necessary, e.g. for large collections)
 - use the default #to_solr implementation
 - include exhibit-specific solr data
 - avoid hard-coding things we should pull out of the blacklight config

cbeer:
e.g for multi-tenant solr indexes

Fixes #883 (or, at least provides a way to do it.)
Fixes #893

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ env:
   global:
     - JRUBY_OPTS="-J-Xms512m -J-Xmx1024m"
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
-  matrix:
-    - "RAILS_VERSION=4.1.8"
-    - "RAILS_VERSION=4.2.0"
 
 sudo: false
 language: ruby
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -153,9 +153,7 @@ class Spotlight::CatalogController < ::CatalogController
 
   def uploaded_resource_params
     if @document.uploaded_resource?
-      return Spotlight::Resources::Upload.fields(current_exhibit).collect do |_, config|
-        config.solr_field
-      end
+      return Spotlight::Resources::Upload.fields(current_exhibit).collect(&:solr_field)
     end
     []
   end
--- a/app/controllers/spotlight/resources/upload_controller.rb
+++ b/app/controllers/spotlight/resources/upload_controller.rb
@@ -1,3 +1,5 @@
+require 'csv'
+
 module Spotlight::Resources
   class UploadController < ApplicationController
     helper :all
@@ -5,7 +7,7 @@ module Spotlight::Resources
     before_filter :authenticate_user!
 
     load_and_authorize_resource :exhibit, class: Spotlight::Exhibit
-    before_filter :build_resource, only: [:new, :create]
+    before_filter :build_resource, only: [:new, :create, :template]
 
     load_and_authorize_resource class: 'Spotlight::Resources::Upload', through_association: "exhibit.resources", instance_name: 'resource'
 
@@ -25,9 +27,25 @@ module Spotlight::Resources
       end
     end
 
+    def csv_upload
+      file = csv_params[:url]
+      csv = CSV.parse(file.read, {headers:true, return_headers: false}).map(&:to_hash)
+      Spotlight::AddUploadsFromCSV.perform_later(csv, current_exhibit, current_user)
+      flash[:notice] = t('spotlight.resources.upload.csv.success', file_name: file.original_filename)
+      redirect_to :back
+    end
+
+    def template
+      render text: CSV.generate { |csv| csv << data_param_keys.unshift(:url) }, content_type: 'text/csv'
+    end
+
     private
     def build_resource
       @resource ||= Spotlight::Resources::Upload.new exhibit: current_exhibit
+    end
+
+    def csv_params
+      params.require(:resources_csv_upload).permit(:url)
     end
 
     def resource_params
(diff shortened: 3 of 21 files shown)


app/models/spotlight/exhibit.rb
```diff
@@ -96,6 +96,10 @@ def import hash
     update hash
   end
 
+  def solr_data
```
cbeer: I'd love a better name  for this method.

app/models/spotlight/resource.rb
```diff
@@ -30,10 +30,10 @@ def self.class_for_resource r
     end
 
     def to_solr
-      {
-        spotlight_resource_id_ssim: "#{(type.tableize if type) || self.class.to_s.tableize }:#{id}",
-        spotlight_resource_url_ssim: url
-      }
+      exhibit.solr_data.merge({
```
jkeck: I think we'll need to do this for the `Spotlight::Resources::Upload` class as well.  Should that be part of this PR or should we just do that separately?

cbeer: For some reason, I thought you were calling `super` there. I'll take a look at it tomorrow and try to reconcile the two.

jkeck: I started out down that path but removed it because the `url` in the context of the uploaded image is a CarrierWave object or something like that.  It may be worth reevaluating that and just deleting out the fields that we don't want coming from the parent class.

cbeer: :+1:

cbeer:
This has spawned #893 while I try to address your comment, fix some problems in the original implementation, and fix some behavior when working with really large collections.

spec/features/javascript/multi_image_select_spec.rb
```diff
@@ -40,7 +40,7 @@
       expect(page).to have_content(/Image \d of \d/)
       click_link("Change")
     end
-
+    sleep 1
```
cbeer: This test is failing consistently for me, but I don't think I broke it..

app/models/spotlight/resource.rb
```diff
@@ -30,10 +29,23 @@ def self.class_for_resource r
...
+        :"#{Spotlight::Engine.config.solr_fields.prefix}spotlight_resource_id#{Spotlight::Engine.config.solr_fields.string_suffix}" => "#{(type.tableize if type) || self.class.to_s.tableize }:#{id}",
```
cbeer: I realized we were hard-coding `_ssim` here, so I've fixed that. I also updated `spotlight_resource_url_ssm` to be `spotlight_resource_url_ssim` to avoid introducing another field type. I don't think that's a problem, right?

app/models/concerns/spotlight/solr_document/atomic_updates.rb
```diff
@@ -1,14 +1,19 @@
 module Spotlight::SolrDocument::AtomicUpdates
 
-  def reindex
-    data = hash_for_solr_update
+  def reindex merge = true
```
cbeer: I'm not thrilled with this, but for really large collections of objects, I needed a way to send individual updates instead of a large array. Maybe that suggests that indexing resources should be treated different than indexing `SolrDocument` objects. Thoughts?

app/models/spotlight/resource.rb
```diff
@@ -1,21 +1,20 @@
...
-    after_save if: :data_changed? do
-      unless performing_reindex
-        performing_reindex = true
-        reindex
-        performing_reindex = false
-        update_index_time!
-      end
-    end
+
+    after_save :reindex
```
cbeer: Better late than never.

app/controllers/spotlight/dashboards_controller.rb
```diff
@@ -26,7 +26,7 @@ def load_recent_solr_documents count
       solr_params = { sort: "#{blacklight_config.index.timestamp_field} desc" }
       @response = query_solr({}, solr_params)
       @response.docs.take(count).map do |doc|
-        ::SolrDocument.new(doc, @response)
+        blacklight_config.solr_document_model.new(doc, @response)
```
cbeer: I probably started the bad habit of using `SolrDocument` directly, but we should pull the class name out of the blacklight config, in case our downstream consumers are doing stuff strangely.

app/models/concerns/spotlight/solr_document.rb
```diff
@@ -42,16 +42,20 @@ def update current_exhibit, new_attributes
...
+    def update_index data = [], merge = false
+      # no-op update index implementation
+    end
+
     def sidecar exhibit
       sidecars.find_or_initialize_by exhibit: exhibit
     end
 
     def to_solr
-      { id: id }.reverse_merge(sidecars.inject({}) { |result, sidecar| result.merge(sidecar.to_solr) }).merge(tags_to_solr)
```
cbeer: Thanks for doing `id` correctly in `Upload`. Hopefully I've fixed all the places I did it wrong before..

PR #897, opened 2015-01-23
Improve exhibit resource index machinery

### LINKED ISSUE #893: Improve resource indexing machinery

cbeer:
Fixes #893 
Fixes #891


cbeer:
(@mejackreed gave this a :+1:, so I've shipped it)

PR #898, opened 2015-01-23
Image versions field

cbeer:
@jkeck can you look at  90c8b54? @snydman was running into this problem in -prod because we hadn't configured a square field. Should we just invent a default? Should we define derivates in some other way?

### DIFF
--- a/app/models/spotlight/resources/upload.rb
+++ b/app/models/spotlight/resources/upload.rb
@@ -66,7 +66,8 @@ module Spotlight
 
     def add_file_versions solr_hash
       Spotlight::ItemUploader.configured_versions.each do |config|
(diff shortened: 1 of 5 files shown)


jkeck:
Hmmm, maybe we should configure the field in `Spotlight::Engine.config` instead of the Blacklight config.  I guess it doesn't really belong there anyhow.  It's similar to the thumbnail url (which is why I put it there), but since it's something that is provided by Spotlight it should probably be in the spotlight config (and that way we can depend on it being there).

cbeer:
Sounds good. I'll let this sit out here, in case we want to do it right.

jkeck:
Just submitted #899 to this PR.  Let me know if you think that'll solve the issue.

PR #899, opened 2015-01-23
Move full, thumb, and square image field configs to Spotlight::Engine...

jkeck:
...config.

### DIFF
--- a/app/uploaders/spotlight/configurable_uploader_versions.rb
+++ b/app/uploaders/spotlight/configurable_uploader_versions.rb
@@ -10,7 +10,7 @@ module Spotlight
       @configured_versions ||= [
         {
           version: :thumb,
-          blacklight_config_field: :thumbnail_field,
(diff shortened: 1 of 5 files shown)


app/uploaders/spotlight/configurable_uploader_versions.rb
```diff
@@ -19,7 +19,7 @@ def configured_versions
         },
         {
           version: :square,
-          blacklight_config_field: :square_image_field,
+          field: Spotlight::Engine.config.square_image_field,
```
cbeer: Want to move the check for the field into here?

jkeck: Okay, I tried something.  Seems a little cleaner than a conditional appending configurations to the array so I'm rejecting any nil fields.  Let me know if you had something better in mind.

cbeer: :+1: works for me

PR #906, opened 2015-01-29
Allow for search results and records to be accessed w/o exhibit context.

### LINKED ISSUE #896: undefined method `apply_permissive_visibility_filter' for #<CatalogController:0x007f53045dc970>
e.g.

```
GET https://exhibits.stanford.edu/maps-of-africa/catalog/citation?id=mj183kc2482
```

jkeck:
I think this fixes #896 

Once I fixed the `#apply_permissive_visibility_filter` issue there were some helpers that were throwing errors.

Not sure if I'm doing the right thing with the `#link_back_to_catalog` override or not, so let me know if you have other ideas.

### DIFF
--- a/app/helpers/spotlight/main_app_helpers.rb
+++ b/app/helpers/spotlight/main_app_helpers.rb
@@ -27,4 +27,18 @@ module Spotlight::MainAppHelpers
     field.enabled && field.send((:show if controller.is_a? Blacklight::Catalog and ["edit", "show"].include?(action_name)) || document_index_view_type)
   end
 
+  def link_back_to_catalog(opts={:label=>nil})
+    opts[:route_set] ||= spotlight if current_exhibit
+    super
+  end
+
+  def render_save_search
+    render('save_search') if render_save_this_search?
+  end
+
+  def render_save_this_search?
+    (current_exhibit and can?( :curate, current_exhibit)) &&
+    (params[:controller] != "spotlight_catalog_controller" && params[:action] != "admin")
+  end
+
 end
--- a/app/helpers/spotlight/main_app_helpers.rb
+++ b/app/helpers/spotlight/main_app_helpers.rb
@@ -28,17 +28,9 @@ module Spotlight::MainAppHelpers
   end
 
   def link_back_to_catalog(opts={:label=>nil})
-    opts[:route_set] ||= spotlight if current_exhibit
+    if (current_search_session.try(:query_params) || {}).fetch(:controller, "").starts_with? "spotlight"
+      opts[:route_set] ||= spotlight
+    end
     super
   end
-
-  def render_save_search
-    render('save_search') if render_save_this_search?
-  end
-
-  def render_save_this_search?
-    (current_exhibit and can?( :curate, current_exhibit)) &&
-    (params[:controller] != "spotlight_catalog_controller" && params[:action] != "admin")
-  end
-
 end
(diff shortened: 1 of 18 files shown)


jkeck:
One question I have when thinking about this is what are we expecting the interaction to be when not in the spotlight/exhibit context?

I *think* we would just be in an interface that has access to the entire solr index (documents existing in all exhibits) and w/o respecting the private/public setting (as that is exhibit specific).

If that's not what we want we should probably re-think what the expected user behavior should be for somebody who is dropped out of the exhibit/spotlight context.

cbeer:
Without an exhibit, I'd expect to get whatever blacklight would normally offer. I don't see us presenting that to the user ourselves, but in Spotlight we shouldn't break that option for others.

app/helpers/spotlight/main_app_helpers.rb
```diff
@@ -27,4 +27,18 @@ def field_enabled? field, *args
     field.enabled && field.send((:show if controller.is_a? Blacklight::Catalog and ["edit", "show"].include?(action_name)) || document_index_view_type)
   end
 
+  def link_back_to_catalog(opts={:label=>nil})
```
cbeer: I assume you ran into these problems after fixing the permissive visibility one?

app/helpers/spotlight/main_app_helpers.rb
```diff
@@ -27,4 +27,18 @@ def field_enabled? field, *args
...
+    opts[:route_set] ||= spotlight if current_exhibit
+    super
+  end
+
+  def render_save_search
```
cbeer: I think with https://github.com/projectblacklight/blacklight/commit/6e5583e525c38ebf215663ea9923c1bc29a9d909, we should be able to do this better (and, in fact, the override we're doing right now needs to be updated too..)

cbeer: Yeah, off-hand, I don't understand why `app/views/catalog/_sort_and_per_page.html.erb` isn't `app/views/spotlight/catalog/_sort_and_per_page.html.erb`.. It seemed to work as expected after moving it.

app/helpers/spotlight/main_app_helpers.rb
jkeck: Yes, exactly.

jkeck:
I'm noticing an error still cropping up when going to catalog#show w/o exhibit context (but exhibit context in your session history).  Maybe I can bounce some ideas off of you next week.

cbeer:
Fixes #922

cbeer:
@jkeck can you take a look? I think this fixed /catalog, includes @bcail's changes in #922 (with blacklight master, at least), and fixes the different views problem you mentioned.

cbeer:
I'm not seeing the test failure travis is reporting.

jkeck:
I'll take a look at this once I come up for air on #914

jkeck:
This all looks good to me @cbeer.  I can't reproduce the error on poltergeist 1.5.1 or 1.6.0 so I don't think it's that (even though the test is about a javascript `confirm()` and there seems to have been some changes there).  I say we make that spec pending on travis and :shipit:.

cbeer:
Will do. I have another PR and i'll just make sure it's the same error.

I also noticed that yesterday was "travis-ci build environment update day", so maybe that had something to do with it.

cbeer:
Hm. The test passed that time. I guess I'll :shipit: for now and we can mark it pending if it becomes annoying again.

PR #908, opened 2015-01-30
Use explicit named routes when linking to the singular HomePage resource

jkeck:
This PR makes the edit and view links for home pages `/:exhibit_id/home/edit` and `/:exhibit_id` respectively.

We need to have separate logic for home pages because they are a singular resource as opposed to the about and feature pages (see rails/rails#1769).

### DIFF
--- a/app/views/spotlight/dashboards/_page.html.erb
+++ b/app/views/spotlight/dashboards/_page.html.erb
@@ -2,8 +2,13 @@
   <td>
     <h4 class="panel-title"><%= page.title %></h4>
     <div class="page-links">
-      <%= exhibit_view_link page, :class => 'btn btn-link' %> &middot;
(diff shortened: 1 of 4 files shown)


app/views/spotlight/dashboards/_page.html.erb
```diff
@@ -2,8 +2,13 @@
   <td>
     <h4 class="panel-title"><%= page.title %></h4>
     <div class="page-links">
-      <%= exhibit_view_link page, :class => 'btn btn-link' %> &middot;
-      <%= exhibit_edit_link page, :class => 'btn btn-link' %>
+      <% if page.is_a?(Spotlight::HomePage) %>
```
jkeck: I couldn't think of a good way to do this w/o having a lot of extra logic for this one case.  I'm open to suggestions.

PR #922, opened 2015-02-03
update a test; fix a hard-coded :id field in solr document

bcail:
...talog controller, and fix the test

### DIFF
--- a/app/models/concerns/spotlight/solr_document/active_model_concern.rb
+++ b/app/models/concerns/spotlight/solr_document/active_model_concern.rb
@@ -10,7 +10,7 @@ module Spotlight::SolrDocument::ActiveModelConcern
   module ClassMethods
 
     def primary_key
-      :id
+      self.unique_key
     end
     
     # needed for Rails 4.1 + act_as_taggable
--- a/spec/controllers/spotlight/catalog_controller_spec.rb
+++ b/spec/controllers/spotlight/catalog_controller_spec.rb
@@ -190,8 +190,9 @@ describe Spotlight::CatalogController, :type => :controller do
     end
     describe "PATCH update" do
       it "should be successful" do
-        patch :update, exhibit_id: exhibit, id: 'dq287tq6352', solr_document: {tag_list: 'one, two'}
+        patch :update, exhibit_id: exhibit, id: 'dq287tq6352', solr_document: {exhibit_tag_list: 'one, two'}
         expect(response).to be_redirect
+        expect(exhibit.owned_taggings.last.tag_id).to eq 2
       end
     end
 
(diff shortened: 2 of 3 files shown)


spec/controllers/spotlight/catalog_controller_spec.rb
```diff
@@ -190,8 +190,9 @@
     end
     describe "PATCH update" do
       it "should be successful" do
-        patch :update, exhibit_id: exhibit, id: 'dq287tq6352', solr_document: {tag_list: 'one, two'}
+        patch :update, exhibit_id: exhibit, id: 'dq287tq6352', solr_document: {exhibit_tag_list: 'one, two'}
         expect(response).to be_redirect
+        expect(exhibit.owned_taggings.last.tag_id).to eq 2
```
cbeer: This should probably be, e.g.:

```
expect {
 patch ...
}.to change { exhibit.owned_taggings.length }.by(2)
```

app/models/concerns/spotlight/solr_document/active_model_concern.rb
```diff
@@ -10,7 +10,7 @@ module Spotlight::SolrDocument::ActiveModelConcern
   module ClassMethods
 
     def primary_key
-      :id
+      self.unique_key
```
cbeer: I'm actually a little surprised at this change. You need to do this even after configuring your `SolrDocument` correctly by telling it what the unique key column is?

https://github.com/projectblacklight/blacklight/blob/master/lib/blacklight/solr/document.rb#L174-177

bcail: I've got the following line in app/models/solr_document.rb in my app:
self.unique_key = 'pid'
Is there anything else that needs to be set?

spec/controllers/spotlight/catalog_controller_spec.rb
bcail: updated.

app/models/concerns/spotlight/solr_document/active_model_concern.rb
bcail: related note: there seems to be a similar problem in atomic_updates.rb. When I put the following in hash_for_solr_update, the solr update works. Otherwise, the solr update fails because it's trying to set: pid.
Hash[doc.map do |k,v|
        val = if k.to_sym == :pid

bcail:
I added a commit yesterday for the unique key in AtomicUpdates. Are you thinking that these changes will start to be unnecessary (or need to be reworked) when your primary_key PR for blacklight is merged?

bcail:
Update: I tested my app with the latest blacklight master (with the primary_key PR merged) and spotlight master (without this PR), and the error is back. Seems like this PR is still needed.

bcail:
Closing - these changes are handled in another PR. Thanks!

PR #929, opened 2015-02-06
Use and check fine-grained permissions for admin and curator actions

### LINKED ISSUE #904: Add finer grain permissions for administrator/curator actions
At the very least, we want the ability (for ourselves) to restrict import/export to only superadmins, but there's probably some benefit to gating some of the other types of actions (maybe add: :import, :configure_blacklight (?), .. ?)

cbeer:
Fixes #904.

### DIFF
--- a/app/controllers/spotlight/appearances_controller.rb
+++ b/app/controllers/spotlight/appearances_controller.rb
@@ -20,7 +20,7 @@ class Spotlight::AppearancesController < Spotlight::ApplicationController
   protected
 
   def load_and_authorize_appearance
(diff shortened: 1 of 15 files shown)


cbeer:
Do you think any of that needs to be explicitly tested?

jkeck:
I could see the value in testing that the links aren't showing up in the view. I think the individual controllers (typically) test that an admin can access and an anonymous user cannot.  I don't think we have anything setup to do that yet so if it ends up being more pain than it's worth then let's :shipit: and make a tech-debt ticket for later.

cbeer:
Filed #931. It's a pain because our testing is at the role level, not the abilities.

PR #930, opened 2015-02-07
Add Featured Browse Categories Widget

### LINKED ISSUE #914: New widget: Highlight browse categories

jkeck:
Closes #914 

Still working out some kinks w/ the tests.  Also, @ggeisler can you take a look at the 5 category w/ no-sidebar on different view-ports.  Under certain circumstances I see the categories overlapping.

TODO:
- [x] Use variables for height, width, colors, etc.
- [x] Pick and set a good fallback size for the browse categories (falling back on smallest square).
- [x] Store browse category slug instead of ID.
- [x] Make `Show item counts` enabled by default.

## Edit form
edit

## With Sidebar
sidebar-2

sidebar-3

## Without Sidebar
no-sidebar-2
no-sidebar-3
no-sidebar-4
no-sidebar-5

### DIFF
--- a/app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
+++ b/app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
@@ -1,0 +1,90 @@
+$featured-browse-category-border-color: $legend-border-color;
+$featured-browse-category-caption-color: $gray-lighter;
+$smallest-featured-browse-category-width: 220px;
+$smallest-featured-browse-category-height: $smallest-featured-browse-category-width;
+$largest-featured-browse-category-height: 255px;
+$no-sidebar-medium-image-width: 300px;
+$no-sidebar-large-image-width: 360px;
+$with-sidebar-medium-image-width: 340px;
+$with-sidebar-large-image-width: $no-sidebar-large-image-width;
+
+.browse-category {
+  background-size: cover;
+  background-position: center;
+  border: 2px solid $featured-browse-category-border-color;
+  border-radius: $border-radius-large;
+  position: relative;
+  float: left;
+  background-repeat: no-repeat;
+  .category-caption {
+    color: $featured-browse-category-caption-color;
+    position: absolute;
+    bottom: $line-height-computed;
+    text-align: center;
+    text-shadow: 0 1px 0 #000000;
+    width: 100%;
+  }
+  .category-title {
+    font-size: $font-size-large;
+    line-height: 1.2;
+    margin: 0;
+    padding: $table-condensed-cell-padding;
+  }
+  .item-count {
+    font-size: $font-size-base;
+    text-transform: uppercase;
+  }
+}
+
+[data-sidebar="false"] {
+  &.categories-1, &.categories-2, &.categories-3 {
+    .browse-category {
+      height: $largest-featured-browse-category-height;
+      @media (max-width: $screen-md-max) {
+        width: $no-sidebar-medium-image-width;
+      }
+      @media (min-width: $screen-lg-min) {
+        width: $no-sidebar-large-image-width;
+      }
+    }
+  }
+  &.categories-4, &.categories-5 {
+    .browse-category {
+      width: $smallest-featured-browse-category-width;
+      height: $smallest-featured-browse-category-height;
+    }
+    .category-4 {
+      @extend .hidden-sm;
+    }
+    .category-5 {
+      @extend .hidden-md;
+      @extend .hidden-sm;
+    }
+  }
+}
+
+[data-sidebar="true"] {
+  &.categories-1, &.categories-2 {
+    .browse-category {
+      height: $largest-featured-browse-category-height;
+      @media (max-width: $screen-md-max) {
+        width: $with-sidebar-medium-image-width;
+      }
+      @media (min-width: $screen-lg-min) {
+        width: $with-sidebar-large-image-width;
+      }
+    }
+  }
+  &.categories-3, &.categories-4, &.categories-5 {
+   .browse-category {
+      width: $smallest-featured-browse-category-width;
+      height: $smallest-featured-browse-category-height;
+    }
+  }
+  .category-3 {
+    @extend .hidden-sm;
+  }
+  .category-4, .category-5 {
+    display: none;
+  }
+}
--- a/app/assets/javascripts/spotlight/blocks/featured_browse_categories_block.js
+++ b/app/assets/javascripts/spotlight/blocks/featured_browse_categories_block.js
@@ -1,0 +1,170 @@
+/*
+  Block to get allow a user to choose browse
+  categories to feature on various pages.
+*/
+
+SirTrevor.Blocks.FeaturedBrowseCategories =  (function(){
+
+  return Spotlight.Block.extend({
+
+  searches_key: "featured-browse-categories",
+
+  blockGroup: 'Exhibit item widgets',
+
+  description: "This block visually highlights up to five browse categories, which are linked to the browse category results page.",
+
+  maxCategoriesAllowed: 5,
+
+  template: [
+    '<div class="featured-browse-categories-block-admin clearFix">',
+      '<div class="widget-header">',
+        '<%= description %>',
+      '</div>',
+      '<div class="col-sm-12">',
+        '<label for="<%= formId(searches_key) %>">Browse categories</label>',
+        '<div id="<%= formId(searches_key) %>" data-featured-browse-categories="true">',
+        '</div>',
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,69 @@
+.browse-category {
```
cbeer: This is more a note for later, but we should use bootstrap variables (or calculations based on those variables..) for the magic numbers below.

I also vaguely wonder if this is another use case for `flexbox`.

cbeer:
:+1: Nice

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
jkeck: Good call w/ the variables.  I had was copy-pasting from the example code in the issue and had intended on moving it to variables.  I've added it as a TODO in this PR.

This is using `flexbox` (we're using `.spotlight-flexbox` in the DOM).  I admittedly don't have a full understanding of `flexbox` so we may be doing something redundant or against what `flexbox` gains us here.

ggeisler:
Looking good, @jkeck! Regarding the overlapping image problem you see with 5-up, in the original ticket I suggested we hide the 5th image at the medium breakpoint (992px-1199px). It looks like the code currently doesn't hide any images below 1200px, so that's why the overlap. However, even hiding the 5th image isn't enough to prevent overlap at the small breakpoint, so here is what I suggest:

For the 4- or 5-up version (220px image version):

- give the `box category-5` div the equivalent of the Bootstrap `hidden-sm hidden-md` classes (hide 5th div between 768px and 1199px) 
- give the `box category-4` div the equivalent of the Bootstrap `hidden-sm` classes (hide 4th div between 768px and 991px) 

For the 2- or 3-up version (larger image version):

- give the `box category-3` div the equivalent of the Bootstrap `hidden-sm` class (hide
 3th div between 768px and 991px) 

That would give us this at the small viewport (this is the worst case, hiding 1 of 3 large and 2 of 5 small images):

default_exhibit_-_blacklight

There is still some overlap and general unattractiveness below the small viewport, but that's the case with other widgets too, so perhaps we should push worrying about that to another ticket I can create to focus on improving extra-small responsiveness more broadly?

ggeisler:
Also, can you make the "Include item counts?" checkbox `true` as the default?

jkeck:
@ggeisler okay, check out what I've done in 0b80e1009ed626369eb2290593743b9bfae1b0d6 and let me know if that works for you.

jkeck:
I'll rebase before we :shipit:, just leaving as is now for readability.

cbeer:
:+1:

ggeisler:
@jkeck this is great. One minor suggestion is to add `background-position: center;` to the `.browse-category` class, which will center the images. It's not a huge improvement with the maps (see below for a simple example) since they have the black borders, but for a photography collection like Fitch I think it'll be a noticeable improvement to center the images.

maps_of_africa__an_online_exhibit_-_blacklight

jkeck:
@ggeisler: 
background-center

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
+$featured-browse-category-border-color: #b3b3b3;
```
cbeer: Do we need our own special grays, or can we use bootstrap's (at least out-of-the-box)? These seem close-ish:

$legend-border-color c5c5c5
$gray-lighter fff

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
...
+    width: 100%;
+  }
+  .category-title {
+    font-size: 18px;
+    line-height: 1.2;
+    margin: 0;
+    padding: 6px;
+  }
+  .item-count {
+    font-size: 14px;
```
cbeer: $font-size-base?

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
...
+  .category-caption {
+    color: $featured-browse-category-caption-color;
+    position: absolute;
+    bottom: 20px;
+    text-align: center;
+    text-shadow: 0 1px 0 #000000;
+    width: 100%;
+  }
+  .category-title {
+    font-size: 18px;
```
cbeer: $font-size-large?

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
...
+$no-sidebar-medium-image-width: 300px;
+$no-sidebar-large-image-width: 360px;
+$with-sidebar-medium-image-width: 340px;
+$with-sidebar-large-image-width: $no-sidebar-large-image-width;
+
+.browse-category {
+  background-size: cover;
+  background-position: center;
+  border: 2px solid $featured-browse-category-border-color;
+  border-radius: 7px;
```
cbeer: $border-radius-large is 6px.

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
+$featured-browse-category-border-color: #b3b3b3;
+$featured-browse-category-caption-color: #f7f7f7;
+$smallest-featured-browse-category-width: 220px;
```
cbeer: Do you think we can calculate these from $container-tablet / $container-desktop / $container-large-desktop?

ggeisler:
@jkeck looks good!

$legend-border-color is fine to use for $featured-browse-category-border-color (currently #b3b3b3).

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
ggeisler: $border-radius-large would be fine

ggeisler: $featured-browse-category-caption-color could be $gray-lighter (just trying to avoid pure white so the text isn't too dominant on the page)

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
cbeer: e.g.:

```
$featured-browse-category-margin: 14px;
$no-sidebar-large-image-width: floor(($container-large-desktop)/ 3) - 2*$border-radius-large - 4 - $featured-browse-category-margin;

```

jkeck:
Okay, updated the font sizes, border radius, and colors to use existing bootstrap variables.

new-borders-and-colors

WRT the image sizes proportional to the bootstrap container size variables, I'm just not confident enough in my knowledge of the rational behind the image sizes in the first place and whether there would be unintended consequences to pinning to those sizes.

What would you think about me creating a ticket to investigate using bootstrap's container sizes to calculate the image sizes dynamically in the future?

cbeer:
:+1:

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
...
+  background-position: center;
+  border: 2px solid $featured-browse-category-border-color;
+  border-radius: $border-radius-large;
+  position: relative;
+  float: left;
+  background-repeat: no-repeat;
+  .category-caption {
+    color: $featured-browse-category-caption-color;
+    position: absolute;
+    bottom: 20px;
```
cbeer: $line-height-computed?

app/assets/stylesheets/spotlight/_featured_browse_categories_block.scss
```diff
@@ -0,0 +1,90 @@
...
+    text-align: center;
+    text-shadow: 0 1px 0 #000000;
+    width: 100%;
+  }
+  .category-title {
+    font-size: $font-size-large;
+    line-height: 1.2;
+    margin: 0;
+    padding: 6px;
```
cbeer: $table-condensed-cell-padding (5px)?

cbeer:
That's it, I promise. Ship at will.

cbeer:
:shipit:

jkeck:
Last two variables added and commits squashed.

PR #940, opened 2015-02-18
After adding an item to the exhibit, force a solr commit and return the ...

cbeer:
...user to the admin page sorted by timestamp.

Hopefully this will make the flow of add item => admin catalog page make more sense.

untitled

### DIFF
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -196,4 +196,12 @@ class Spotlight::CatalogController < ::CatalogController
       end
     end
   end
+
+  # calls setup_previous_document then setup_next_document.
+  # used in the show action for single view pagination.
+  def setup_next_and_previous_documents
+    super
+  rescue RSolr::Error::Http => e
+    Rails.logger.warn "Unable to setup next and previous documents: #{e}"
+  end
 end
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -51,7 +51,10 @@ class Spotlight::CatalogController < ::CatalogController
   def admin
     self.blacklight_config.view.select! { |k,v| k == :admin_table }
     self.blacklight_config.view.admin_table.partials = [:index_compact]
-    self.blacklight_config.sort_fields.reject! { |k,v| true }
+
+    unless self.blacklight_config.sort_fields.has_key? :timestamp
+      self.blacklight_config.add_sort_field :timestamp, sort: "#{blacklight_config.index.timestamp_field} desc"
+    end
 
     add_breadcrumb t(:'spotlight.curation.sidebar.header'), exhibit_dashboard_path(@exhibit)
     add_breadcrumb t(:'spotlight.curation.sidebar.items'), admin_exhibit_catalog_index_path(@exhibit)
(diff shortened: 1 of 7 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -51,7 +51,10 @@ def autocomplete
   def admin
     self.blacklight_config.view.select! { |k,v| k == :admin_table }
     self.blacklight_config.view.admin_table.partials = [:index_compact]
-    self.blacklight_config.sort_fields.reject! { |k,v| true }
+
+    unless self.blacklight_config.sort_fields.has_key? :timestamp
+      self.blacklight_config.add_sort_field :timestamp, sort: "#{blacklight_config.index.timestamp_field} desc"
```
jkeck: We can't have this sort field configured only on catalog#admin because if you click on an item from a timestamp sorted result set the Previous/Next pagination on the record page will attempt to sort via a config that doesn't exist (which will default to the `timestamp` key, which will fail because it doesn't exist in solr).

app/controllers/spotlight/resources/upload_controller.rb
```diff
@@ -14,16 +14,16 @@ class UploadController < ApplicationController
...
       @resource.attributes = resource_params
 
-      if @resource.save
+      if @resource.save_and_commit
         flash[:notice] = t('spotlight.resources.upload.success')
         if params["add-and-continue"]
           redirect_to new_exhibit_resources_upload_path(@resource.exhibit)
         else
-          redirect_to admin_exhibit_catalog_index_path(@resource.exhibit)
+          redirect_to admin_exhibit_catalog_index_path(@resource.exhibit, sort: :timestamp)
```
jkeck: What would you think of changing https://github.com/sul-dlss/spotlight/blob/master/config/routes.rb#L28 to be `get 'admin', defaults: {sort: 'timestamp'}` instead?

app/controllers/spotlight/catalog_controller.rb
cbeer: Oh, boo. I'm about ready to build our own AdminCatalogController to support these pages..

app/controllers/spotlight/resources/upload_controller.rb
cbeer: I'm just not sure what we want the default behavior to be, so I tried to leave the existing behavior alone.

app/controllers/spotlight/catalog_controller.rb
jkeck: I think that still may have this issue because the previous and next link will be rendered under the context of the app's `CatalogController` and it will attempt to lookup a sort configuration (from the session's searches), which wouldn't have a timestamp sort configured. 

Unless you mean the `AdminController` either blacklists the sort key from being added to the session or just doesn't store its searches at all. (or some other way that I'm sure I'm missing)

app/controllers/spotlight/resources/upload_controller.rb
jkeck: Fair enough.

app/controllers/spotlight/catalog_controller.rb
cbeer: Yeah, maybe not passing the search context would be easiest. I'll see how possible that is.

cbeer: Or making Blacklight not break when the sort isn't available. Maybe that's even better.

cbeer:
and https://github.com/projectblacklight/blacklight/pull/1096

PR #961, opened 2015-02-21
limit page analytics to the top 5 results

cbeer:


app/models/spotlight/exhibit.rb
```diff
@@ -92,19 +92,20 @@ def appearance
...
-  def page_analytics start_date = 1.month
+  def page_analytics start_date = 1.month, path = nil
     return [] unless analytics_provider and analytics_provider.enabled?
-
-    @page_analytics ||= begin
-      analytics_provider.page_data(self, start_date: start_date.ago)
+    
+    @page_analytics ||= {}
+    @page_analytics[start_date] ||= begin
+      analytics_provider.page_data(path || self, start_date: start_date.ago, limit: 5)
```
jkeck: Do you think we should make the limit configurable?  Perhaps a `Spotlight::Engine#config`?

cbeer: Sure thing. See 54c0451.

PR #965, opened 2015-02-24
Move default main navigations list to the engine configuration

cbeer:


jkeck:
This looks great, I do have one question which is more from a future documentation standpoint.

The main purpose of this configuration is to remove items from that navigation (and not necessarily to add items), correct? (e.g. I want to create published browse categories to add to my narrative but may want to hide the `Browse` menu item all-together)

I think there would be some more plumbing necessary to dynamically add menu items to the nav.

cbeer:
Removing items is easy; as you say, adding items will take more work on the implementation side.

Issue #909, opened 2015-01-30
Add option for site masthead to have background image


ggeisler:

ggeisler:
Below are details for displaying a background image in the masthead. The other part of this feature is giving the user a way to select and crop an image to use as the background image -- I'll add details about that once we resolve some decisions about our general approach to image selection and cropping.

The example SCSS below should work with the existing HTML used for the masthead (though I'd like to see the default height of the masthead be slightly smaller -- 120px instead of 131px) so I'm just providing the SCSS. The approach adjusts to any height masthead (assuming the source background image is tall enough, though there will be problems with the text fitting below 60px or so), and any amount of image blur (within reason; single digits). So in theory we should be able to provide the user with configuration options for both the blur amount and the masthead height.

Some example output:

Using the current default masthead height and 1px blur:

![maps-default](https://cloud.githubusercontent.com/assets/101482/6153606/9dffe1a4-b1da-11e4-8d84-3495b52a4547.png)

Same image but using a larger masthead height:

![maps-large](https://cloud.githubusercontent.com/assets/101482/6153629/cb34e7aa-b1da-11e4-859e-78b2c18bc247.png)

Different image:

![walters-default](https://cloud.githubusercontent.com/assets/101482/6153643/e5a93d8e-b1da-11e4-8e50-47a5680c8f02.png)

Different image with higher blur value:

![fitch-default-3px-blur](https://cloud.githubusercontent.com/assets/101482/6153773/e74ee912-b1db-11e4-8d23-36f2344385cc.png)

Extra small viewport, no blur:

![maps-xs-no-blur](https://cloud.githubusercontent.com/assets/101482/6153982/2de7120e-b1dd-11e4-8bca-2637bd1d47c5.png)

The SCSS code:

````scss
$masthead-height: 131px;
$masthead-image-blur: 1px;

#exhibit-masthead > .container {
  // Maintains the site title and subtitle at slightly below
  // vertically-centered, even if height of masthead is changed.
  position: relative;
  top: 60%;
  -webkit-transform: translateY(-50%);
  transform: translateY(-50%);
}

#exhibit-masthead {
  margin-bottom: 0;
  padding-top: 0;
  position: static;
  height: $masthead-height;
  width: 100%;
  // To handle potential text blurring caused by the "transform: translateY"
  // property in #exhibit-masthead > .container
  -webkit-transform-style: preserve-3d;
  -moz-transform-style: preserve-3d;
  transform-style: preserve-3d;

[...]

PR #975, opened 2015-02-26
Add ability to upload and crop a site masthead image

jkeck:
Begins work on #909 by adding just local file upload.

adding-masthead


jkeck:
Have yet to scale the crop image because of a bug in jcrop tapmodo/Jcrop#40.

We may need to end up forking jcrop and merging in tapmodo/Jcrop#152 but I thought I would plug away at feature now w/ the full-sized image.

mejackreed:
:+1: Looks good to me. One jarring thing was when uploading an image, where it ended up and where the resize box was kind of jarring. Not sure if that has anything to do with the jcrop issue. I say :shipit:

jkeck:
@mejackreed I'm going to look into forking Jcrop and applying at least one pending PR that will allow us to scale the image to a max with.  I'm hoping that will address that issue.

jkeck:
And thanks for the :eyes:

PR #976, opened 2015-02-28
Add ability to crop a site masthead from an image in the exhibit

jkeck:
More work on #909 

adding-exhibit-masthead


jkeck:
Will submit a new one on-top of master

PR #979, opened 2015-02-28
Add ability to crop a site masthead from an image in the exhibit

### LINKED ISSUE #909: Add option for site masthead to have background image

jkeck:
Closes #909 

adding-exhibit-masthead


app/models/spotlight/blacklight_configuration.rb
```diff
@@ -170,7 +171,13 @@ def default_blacklight_config
     end
 
     protected
-    
+
+    def spotlight_image_version_fields
+      @@spotlight_image_derivatives.map do |version|
```
cbeer: Does this need to call `@@spotlight_image_derivatives`, or could it use the accessor method?

app/views/spotlight/appearances/edit.html.erb
```diff
@@ -1,8 +1,9 @@
+<% jcrop_options = {croppable: true, selector: 'masthead_image', bg_color: 'black', bg_opacity: '.4', max_size: '[1800,131]', min_size: '[1200,120]'} %>
```
cbeer: Does it make sense to push these options into a helper and/or engine configuration?

app/views/spotlight/appearances/edit.html.erb
```diff
@@ -32,22 +33,23 @@
           <% end %>
           <%= field_set_tag(t(:'.site_masthead.source.header')) do %>
             <div>
-              <%= m.radio_button(:source, :exhibit, disabled: 'disabled', label: t(:'.site_masthead.source.exhibit.label')) %>
+              <%= m.radio_button(:source, :exhibit, label: t(:'.site_masthead.source.exhibit.label')) %>
               <div>
-                <%= m.text_field_without_bootstrap(:document_global_id, class: "form-control", disabled: 'disabled') %>
+                <%= m.hidden_field(:document_global_id) %>
+                <%= m.hidden_field(:remote_image_url, data: jcrop_options) %>
+                <%= text_field_tag(:document_title, presenter(m.object.document || SolrDocument.new).document_show_html_title, class: "form-control", data: {input_select_target: '#appearance_masthead_source_exhibit', masthead_typeahead: true, target_panel: '#masthead-target-panel', :'id-field' => '#appearance_masthead_document_global_id', remote_url_field: '#appearance_masthead_remote_image_url' }) %>
```
cbeer: Don't hard-code `SolrDocument` here. Pull it out of the exhibit config instead (`current_exhibit.blacklight_configuration.solr_document_model` in Blacklight 5.9). Maybe you could get away without needed it, even.

app/views/spotlight/appearances/edit.html.erb
```diff
@@ -32,22 +33,23 @@
...
               <%= m.radio_button(:source, :remote, label: t(:'.site_masthead.source.remote.label')) %>
-              <%= m.file_field_without_bootstrap(:image, data: {
-                croppable: true,
-                selector:  'masthead_image',
-                bg_color:   'black',
-                bg_opacity: '.4',
-                max_size:   '[1800,131]',
-                min_size:   '[1200,120]'
-              }) %>
+              <%= m.file_field_without_bootstrap(:image, data: jcrop_options.merge!(input_select_target: '#appearance_masthead_source_remote')) %>
```
cbeer: Do you need `#merge!`?

I also wonder if there's a rails-y way to get the id of that field, but I doubt it's worth spending time on.

app/models/spotlight/masthead.rb
```diff
@@ -6,6 +6,11 @@ class Spotlight::Masthead < ActiveRecord::Base
     recreate_image_versions if image.present?
   end
 
+  def document
+    return unless document_global_id && source == 'exhibit'
+    @document ||= GlobalID::Locator.locate document_global_id
```
cbeer: :+1:

app/assets/javascripts/spotlight/croppable.js
```diff
@@ -47,38 +47,48 @@ Spotlight.onLoad(function() {
...
+          var reader = new FileReader();
+
+          reader.onload = (function(aImg) { return function(e) {
+            jcrop_api.setImage(e.target.result);
+            cropbox.css({width: "", height: ""});
+            cropbox[0].src = e.target.result;
+            if(previewbox.length > 0) {
+              previewbox[0].src = e.target.result;
+            }
+            jcrop_api.setSelect([0,0,200,200]);
```
cbeer: I'm not sure it's worth tackling now, but we should probably make this default selection configurable. I think this size came from the contacts' avatar crop, right?

spec/models/spotlight/solr_document_images_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    expect(subject.spotlight_image_versions.thumb).to eq ['http://lorempixel.com/400/240/']
+    expect(subject.spotlight_image_versions.square).to eq ['http://lorempixel.com/200/200/']
+  end
+  it 'should include the version keys in the versions array' do
+    [:full, :thumb, :square].each do |version|
+      expect(subject.spotlight_image_versions.versions).to include version
+    end
+  end
+  it 'should include newly configured image versions' do
+    Spotlight::ImageDerivatives.spotlight_image_derivatives << {
```
cbeer: When I've had to mutate global state like this, I try used stubbing instead, e.g.:

```ruby
allow(Spotlight::ImageDerivatives).to receive(:spotlight_image_derivatives).and_return([{version: :tiny, field: :new_image_field}]
```

app/models/concerns/spotlight/solr_document/spotlight_images.rb
```diff
@@ -0,0 +1,41 @@
...
+    @spotlight_image_versions ||= Versions.new(self)
+  end
+
+  class Versions
+    include Spotlight::ImageDerivatives
+    attr_reader :versions
+
+    def initialize(document)
+      @versions = []
+      @@spotlight_image_derivatives.each do |derivative|
```
cbeer: Can we `map` instead of building up the array?

```ruby
@versions = @@spotlight_image_derivatives.map ...
```

app/models/concerns/spotlight/image_derivatives.rb
```diff
@@ -0,0 +1,59 @@
...
+###
+module Spotlight
+  module ImageDerivatives
+    mattr_accessor :spotlight_image_derivatives
+
+    # Extend Spotlight::ImageDerivatives in a CarrierWave uploader
+    # then you can call this as a class method and all of the
+    # configured versions will be available
+    def apply_spotlight_image_derivative_versions
+      @@spotlight_image_derivatives.each do |version_config|
```
cbeer: Same question about using the accessor instead of the `@@`. Maybe we should even have an instance-level accessor too.

app/assets/javascripts/spotlight/croppable.js
jkeck: :+1: I was experimenting w/ that but tossed it out for some reason.  I'll look at it again.

app/models/concerns/spotlight/image_derivatives.rb
jkeck: I believe `mattr_accessor` gives us both class and instance level accessors already.

app/views/spotlight/appearances/edit.html.erb
jkeck: Ah! I was searching around to try to figure out where the `SolrDocument` class was configured.  Thanks!  I'll change it.

app/views/spotlight/appearances/edit.html.erb
jkeck: Yeah, I think I'll push it into a helper.  Some of the stuff doesn't make sense to be configured in the Engine config (like selector/croppable).

I'll probably do some experimenting w/ the Jcrop stuff and the masthead size/proportions once this is merged in, so may end up doing some more stuff w/ this config that could end up lending itself to an Engine configuration.

app/views/spotlight/appearances/edit.html.erb
jkeck: Will take out `#merge!` (it's faster, but in the context of the view it doesn't really matter).

It seems like getting the ID from the form object itself is kinda hacky and requires nasty regexs on the form's `#object_name`.

spec/models/spotlight/solr_document_images_spec.rb
jkeck: I wasn't able to get that to work via stubbing, perhaps I'm doing something wrong...

jkeck:
Updated gif!

aspect-ratio-masthead

PR #980, opened 2015-02-28
Use Jcrop fork to fix boxWidth + minSize/maxSize issues.

jkeck:
## Before
jcrop-before
---

## After
jcrop-after

### DIFF
--- a/app/assets/javascripts/spotlight/jcrop.js
+++ b/app/assets/javascripts/spotlight/jcrop.js
@@ -1,0 +1,1696 @@
+/**
+ * jquery.Jcrop.js v0.9.12
+ * jQuery Image Cropping Plugin - released under MIT License 
+ * Author: Kelly Hallman <khallman@gmail.com>
+ * http://github.com/tapmodo/Jcrop
+ * Copyright (c) 2008-2013 Tapmodo Interactive LLC {{{
+ *
+ * Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/spotlight/jcrop.js
```diff
@@ -0,0 +1,1696 @@
...
+ * OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ *
+ * }}}
+ *
+ * This is a fork of the above mentioned code which is available at https://github.com/sul-dlss/Jcrop
```
cbeer: Do you think we should add the git sha to this comment, or should it be obvious enough?

jkeck: I think this is obvious enough.  If we start getting into modifying this codebase in anyway more significant way then maybe, yeah.

PR #984, opened 2015-03-03
pre-escape the document title, or #truncate will double-escape the title

cbeer:


cbeer:
Probably needs tests.. although I'd like to remember why we're doing such heavy-handed truncation here. Off-hand, I didn't see anything wrong with using the full title or doing some css truncation.

jkeck:
Would you say it's worth just removing the truncation for now?  This is on the admin page, idk why we would truncate on the admin page and not anywhere else.

ggeisler:
If this is the titles on the dashboard or Curation > Items page, I think the rationale for truncating is to try to keep the lists of items compact and easy to scan. The paragraph-long titles items sometimes have reduce scanability. The curator most likely doesn't need to see an entire long title to identify an item, and on the Admin side the most important thing is just enabling the curator to quickly find an item, not understand its full title. Longer-term I think the dashboard and items page will be more usable if they are even more compact than they are now.

That said, if we can use CSS truncation, that would be fine.

PR #992, opened 2015-03-04
Move and group items differently in user utility links menu.

### LINKED ISSUE #718: Menu bar could be more intuitive
It took me a little while to remember how to use Spotlight. At the very least, I think the dashboard link and the create exhibit link need to be grouped together, but maybe the exhibit-building features need to be separated from the user features?

screen shot 2014-11-12 at 9 00 02 am

jkeck:
Fixes #718

### User with permissions
authed-menu

### User without permissions
no-auth-menu


cbeer:
:+1:

PR #997, opened 2015-03-04
Add masthead and thumbnails to pages, searches, and update the exhibit masthead to use common code

cbeer:
screen shot 2015-03-05 at 10 20 37 am
screen shot 2015-03-05 at 10 20 54 am
screen shot 2015-03-05 at 10 21 13 am
screen shot 2015-03-05 at 10 53 24 am
screen shot 2015-03-05 at 10 53 13 am


jkeck:
Looking good so far.  I'll have to look a little closer tomorrow because I'm noticing a couple of oddities (some of which may not be related to this PR at all).

One thing that is totally non-critical is that the document that we get the default thumb from is not pre-selected for browse categories (you may have also mentioned that was more trouble than it's worth to put the default document's global-id in there automatically).

no-item-selected

app/models/spotlight/search.rb
```diff
@@ -52,22 +50,30 @@ def images
   end
 
   def as_json(*args)
-    super.merge(featured_image: featured_image, count: count)
-  end
-
-  def default_featured_item_id
-    images.first.first if images.present?
+    super.merge(count: count)
```
jkeck: Can we ditch this override/super in favor of adding #count to the `methods` options passed to #as_json in https://github.com/sul-dlss/spotlight/pull/997/files#diff-dd6db6608a4fa90bcf92593e2296f214R23

I can't remember if that's the only place this is used or not.

cbeer: :+1: makes sense to me, and I can't imagine where else we'd serialize these to JSON..

spec/features/javascript/browse_category_admin_spec.rb
```diff
@@ -4,16 +4,5 @@
...
-      visit spotlight.edit_exhibit_search_path exhibit, exhibit.searches.first
-      autocomplete_field = find("input#featured-item-title")
-      fill_in_typeahead_field "search[featured_item_id]", with: "gt736xf9712"
-      click_button "Save changes"
-      expect(page).to have_content "The search was successfully updated."
-      within(".pic.thumbnail") do
-        expect(page).to have_css('img[src="https://stacks.stanford.edu/image/gt736xf9712/gt736xf9712_05_0001_thumb"]')
-      end
-    end
-  end
```
jkeck: Is there a test we should have here any longer, or just delete the spec file all together?

app/models/spotlight/search.rb
```diff
@@ -10,24 +10,22 @@ class Spotlight::Search < ActiveRecord::Base
...
-    self.featured_item_id ||= default_featured_item_id
+    set_default_featured_image
   end
 
   include Blacklight::SolrHelper
   include Spotlight::Catalog::AccessControlsEnforcement
 
-  def featured_item
-    if self.featured_item_id.present?
-      @featured_item ||= get_solr_response_for_doc_id(self.featured_item_id).last
```
cbeer: I've realized that we're no longer restricting the thumbnail to the browse category. We should tackle that as part of future work..

jkeck: Want to create a ticket for that just so we don't forget?

jkeck:
Okay, all looks good except for the couple of (non-blocker) things I mentioned (some of which you've already addressed).  :+1: great work.

spec/features/javascript/browse_category_admin_spec.rb
cbeer: Yeah, let me fix that.

app/models/spotlight/search.rb
cbeer: Scratch that. In the UI, we're limiting the autocomplete to members of the browse category. I guess that's probably good enough.

spec/features/javascript/browse_category_admin_spec.rb
cbeer: Thanks, I've moved the tests into `spec/featured/browse_category_admin_spec.rb` and deleted this file.

cbeer:
Good point on the item context. I've tried to address it, so we'll see if it works..

PR #1002, opened 2015-03-05
Try to force UTF-8 encoding on CSV data and replace invalid and...

jkeck:
...undefined characters with a '?'.

### Example indexed data w/ replacement characters.
example-replaced-data

### DIFF
--- a/app/jobs/spotlight/add_uploads_from_csv.rb
+++ b/app/jobs/spotlight/add_uploads_from_csv.rb
@@ -1,3 +1,4 @@
+# encoding: utf-8
 module Spotlight
   class AddUploadsFromCSV < ActiveJob::Base
     queue_as :default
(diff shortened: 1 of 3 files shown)


jkeck:
Fair warning, I don't really get character encodings that well, so this is just something that seems to work.

app/jobs/spotlight/add_uploads_from_csv.rb
```diff
@@ -19,5 +20,13 @@ def perform(csv_data, exhibit, user)
       end
 
     end
+    private
+    def encoded_csv(csv)
+      csv.map do |row|
+        row.map do |label, column|
+          [label, column.encode('UTF-8', invalid: :replace, undef: :replace, replace: "?")] if column.present?
```
cbeer: I'd quibble with using `?`. We should be able to use `\uFFFD` (unicode replacement character) instead.

Also, check out `String#scrub` and https://github.com/jrochkind/scrub_rb.

jkeck: Replaced w/ `\uFFFD`.  I tried String#scrub (under ruby 2.1) but it seemed to still throw the same error.

PR #1007, opened 2015-03-06
Extract previewable and titleable block mixins

cbeer:


app/assets/javascripts/spotlight/block_mixins/previewable.js
```diff
@@ -0,0 +1,101 @@
+(function ($){
+  SirTrevor.PreviewButton = function() {
+    this._ensureElement();
+    this._bindFunctions();
+  }
+
+  SirTrevor.PreviewButton.prototype = Object.create(SirTrevor.BlockDeletion.prototype);
```
cbeer: This is annoying, but SirTrevor doesn't expose the base class we actually want..

PR #1009, opened 2015-03-06
Update to blacklight 5.10

cbeer:


cbeer:
There's still quite a few deprecation warnings, but they need extra effort to resolve.

PR #1011, opened 2015-03-07
Sir trevor widget refactor

### LINKED ISSUE #981: Unify all multi-up widgets admin and display logic
- [ ] Use a common base block class
- [ ] Store GlobalIDs in SirTrevor widget serialization (We currently store SolrDocument ids or FriendlyID slugs.  Using GlobalIDs may simplify the object fetching a bit.)

cbeer:
Fixes #981 

- [x] rendering new style 
- [x] add various display types (multi-up, carousel, featured list, etc)
- [x] populate caption options
- [x] use slugs instead of primary key ids
- [x] configure `typeahead.js` to suggest pages and browse categories  without a minimum character count
- [x] add item limit to the search results block
- [x] trigger javascript behaviors on preview (for carousel; featured list; )
- [ ] use `select_related_input `
screen shot 2015-03-06 at 4 18 43 pm


app/assets/javascripts/spotlight/blocks/block.js
```diff
@@ -1,132 +1,16 @@
...
+    beforeBlockRender: function() {
+      this.availableMixins.forEach(function(mixin) {
+        if (this[mixin] && SirTrevor.BlockMixins[_.capitalize(mixin)].preload) {
+          this.withMixin(SirTrevor.BlockMixins[_.capitalize(mixin)]);
         }
-      });
+      }, this);
     },
-
-    addCaptionSelectFocus: function(){
```
jkeck: Not a huge deal but we have lost this behavior (this would check the checkbox next to the caption field when selected and focus on the select when the checkbox is checked).

We have something that does 1/2 of that in a new jquery plugin at https://github.com/sul-dlss/spotlight/blob/master/app/assets/javascripts/spotlight/select_related_input.js 

Up to you if we :shipit: w/o that.

jkeck:
Looked through.  This all looking great.  Excellent work @cbeer!

One minor comment (that's up to you) and I'm ready to :shipit:

app/assets/javascripts/spotlight/blocks/block.js
cbeer: Oh, yes. I meant to replace it with `select_related_input`. I'll fix that tomorrow.

cbeer: I actually couldn't get the old behavior to work for me at all, but I have connected it so when you make a selection, it will check the checkbox for you. I left the markup for what I think is the equivalent to `addCaptionSelectFocus`, so it you want to give it a go..

PR #1018, opened 2015-03-09
Move CarrierWave storage mechanism into a Spotlight::Engine config.

jkeck:


app/uploaders/spotlight/masthead_uploader.rb
```diff
@@ -2,7 +2,7 @@
 module Spotlight
   class MastheadUploader < CarrierWave::Uploader::Base
     include CarrierWave::MiniMagick
-    storage :file
+    storage Spotlight::Engine.config.uploader_storage || :file
```
cbeer: Do we actually need the `|| :file`?

jkeck: Probably not.  I was worried about the affect of somebody setting `Spotlight::Engine.config.uploader_storage = nil` but I guess that would be their own fault.

I'll update the PR.

PR #1039, opened 2015-03-10
Move import/export/delete into tabs on the Settings page

### LINKED ISSUE #886: Make Admin > Settings into tabbed pages with import/export added
I wonder if we have an issue w/ the wording of some of the sidebar links.

There is an `Import/Export` link which I keep wanting to click to "import" things into the exhibit, however I actually "add/import" things using the `Items` link.  I understand it because I work on it internally but I wonder if we could do better about the wording for first-time users.

import-confusion

Just a thought

jkeck:
Close #886 

*Sneaking* in a Jcrop and spec fix.

basic

import

export

delete


cbeer:
:+1:

PR #1040, opened 2015-03-11
Don't memoize the Spotlight::Engine.catalog_controller, which breaks class reloading

cbeer:


cbeer:
Maybe we can be smarter (say, cache it if `cache_classes` is true..), but I'm ready to have class reloading again.

PR #1042, opened 2015-03-11
Add sir-trevor document embed widget

cbeer:
screen shot 2015-03-11 at 10 54 48 am

Now built as a generic way to invoke the blacklight rendering pipeline.

### DIFF
--- a/app/models/sir_trevor_rails/blocks/solr_documents_block.rb
+++ b/app/models/sir_trevor_rails/blocks/solr_documents_block.rb
@@ -20,11 +20,20 @@ module SirTrevorRails::Blocks
     end
 
     def documents?
-      !documents.blank?
+      documents.present?
     end
 
     def text?
-      !text.blank?
+      text.present?
+    end
+
+    def text
+      val = super
+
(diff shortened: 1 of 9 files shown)


app/models/sir_trevor_rails/blocks/solr_documents_block.rb
```diff
@@ -24,7 +24,7 @@ def documents?
     end
 
     def text?
-      !text.blank?
+      !text.blank? and text != "<p><br></p>"
```
jkeck: Ah, I'm needing to do the exact same thing on the oEmbed block.  I'm using `#present?` instead of `!#blank?` though.  Would it be better to use `text.present? && text != '<p><br></p>'`?  I'll defer to and use what you decide.

cbeer: Yeah, I can never remember to use `#present?`.

cbeer: Actually, I can do one better too..

PR #1052, opened 2015-03-12
Fix spacing on embed + text widget.

### LINKED ISSUE #782: Layout of Embed + Text widget
Adding a purl to embed works great.  But when text is added to the widget the layout (percentages) is off.

jkeck:
Also abstracting text behaviors into a SirTrevorRails::Blocks::Textable mixin.

Closes #782 

### Before
widget-before

### After
widget-after

### DIFF
--- a/app/views/spotlight/sir_trevor/blocks/_oembed_block.html.erb
+++ b/app/views/spotlight/sir_trevor/blocks/_oembed_block.html.erb
@@ -1,6 +1,10 @@
-<div class="content-block item-text">
-  <div class="oembed-block <%= "pull-#{ oembed_block.text_align }" if oembed_block.text.present? %>">
+<div class="content-block item-text row">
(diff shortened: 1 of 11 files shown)


cbeer:
:+1:

app/views/spotlight/sir_trevor/blocks/_oembed_block.html.erb
```diff
@@ -1,6 +1,10 @@
-<div class="content-block item-text">
-  <div class="oembed-block <%= "pull-#{ oembed_block.text_align }" if oembed_block.text.present? %>">
+<div class="content-block item-text clearfix">
```
cbeer: I think now that those blocks are wrapped in `col-md`, this should be `row` instead of `clearfix`.

jkeck: :+1: good catch.  Pushed the row class up.

###  Before
before-row

### After
after-row

cbeer:
:+1:

PR #1055, opened 2015-03-13
Make the default selection area for thumbs really large so when forms...

jkeck:
...are updated w/o updating the default crop we get the larges 4:3 image possible.

This was causing an issue when you would edit a browse category without updating the thumbnail (and would automatically crop the top left 400px by 300px as the thumbnail).  This thumb was particularly bad for large images.

Not sure if this is the best solution as it would probably be ideal to keep the original image when the user hans't cropped, but this is a simple solution for now.

## Before
initial-selection-before

## After
initial-selection-after

### DIFF
--- a/app/helpers/spotlight/jcrop_helper.rb
+++ b/app/helpers/spotlight/jcrop_helper.rb
@@ -20,7 +20,7 @@ module Spotlight
         bg_opacity: '.4',
         box_width: '600',
         aspect_ratio: 4.0 / 3.0,
-        initial_set_select: '[0, 0, 400, 300]'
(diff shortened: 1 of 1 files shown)


cbeer:
I idly wonder if we can do better (or, at least, center-crop by default). But, yeah, this is a clear improvement.

jkeck:
Yeah, I wondered that, but I don't think we can w/o knowing the full size of the image.  I suppose we could call setSelect again once the image has loaded and we can calculate the true dimensions, but if we're going to try and go down the IIIF route I'm not sure how much time we should invest in that sort of thing.

PR #1057, opened 2015-03-13
Use sir-trevor's i18n framework for translations of blocks

### LINKED ISSUE #1021: Use i18n strings for javascript

cbeer:
I'm going to say, fixes #1021. There's a handful of random other strings (form observer, multiimage selector) outside of ST that we should figure out to deal with, but this was low-hanging fruit.

### DIFF
--- a/app/views/spotlight/pages/_sort_and_per_page.html.erb
+++ b/app/views/spotlight/pages/_sort_and_per_page.html.erb
@@ -1,7 +1,0 @@
-<div id="sortAndPerPage" class="clearfix">
-  <div class="search-widgets pull-right">
-    <%= render :partial => 'sort_widget' %>
-    <%= render :partial => 'per_page_widget' %>
-    <%= render :partial => 'view_type_group', locals: { block: block } %>
-  </div>
-</div>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/spotlight/blocks/resources_block.js
```diff
@@ -51,7 +51,7 @@ Spotlight.Block.Resources = (function(){
                     '<div data-panel-image-pagination="true"></div>',
                   '</div>',
                   '<div class="remove pull-right">',
-                    '<a data-item-grid-panel-remove="true" href="#">Remove</a>',
+                    '<a data-item-grid-panel-remove="true" href="#"><%= i18n.t("blocks:resources:panel:remove") %></a>',
```
jkeck: It don't think this is using the (lodash? underscore?) templates so these are just getting printed out as strings:

### Remove link
remove

### Display? label
display-label

cbeer: Thanks for catching that. I'll try to get them parsed.

cbeer:
See #1059. Screwed up pushing to the branch.

PR #1062, opened 2015-03-16
Catalog admin styling

### LINKED ISSUE #1045: Update Dashboard and Curation > Items item lists
We could improve the look and scanability of the table that shows recent items (Dashboard) and the exhibit items (Curation > Items) if we update things to display small, regular-sized thumbnail images, plus update the columns class of the Title field. (I think the same partial is used in these two contexts.)

Below is an example of what I'm after. It's mostly CSS; I'm just not sure the best way to deal with updating the columns class of the Title field since it appears we're bring that in from Blacklight.

#### Current

curation_-_curation_-_items___bob_fitch_photography_archive_-_blacklight
-----

#### Improved

curation_-_curation_-_items___bob_fitch_photography_archive_-_blacklight 2
----

The update to the Title field is to either remove the `col-sm-9 col-lg-10` classes from the `h5` tag below, or change them to `col-sm-12`. This is to enable the title to extend the full width of its containing `td`, which will reduce the height of rows with longer titles.

````html
<h5 class="index_title col-sm-9 col-lg-10">
````

#### CSS used for "Improved" mockup

````css
#documents.table {
  font-size: $font-size-base;

  .document-thumbnail {
    border: none;
    margin-bottom: 0;
  }

  .document-thumbnail a > img {
    border: 1px solid $gray-light;
    width: 60px;
    height: 60px;
    object-fit: cover;
  }

  .document-counter {display: none;}

  .index_title {
    margin-top: $padding-small-vertical;
    margin-bottom: 0;
  }
}
````

cbeer:
Fixes #1045 

screen shot 2015-03-16 at 9 32 38 am

### DIFF
--- a/app/assets/stylesheets/spotlight/_blacklight_configuration.scss
+++ b/app/assets/stylesheets/spotlight/_blacklight_configuration.scss
@@ -64,10 +64,6 @@
   }
 }
 
-.index_title {
-  padding-left: 0;
-}
-
 .disabled-sort-option {
   .panel {margin-bottom: 0;}
 }
--- a/app/assets/stylesheets/spotlight/_catalog.scss
+++ b/app/assets/stylesheets/spotlight/_catalog.scss
@@ -1,31 +1,29 @@
 #documents.table {
-  .documentHeader {
-
-  }
-
-  .checkbox-toggle {
-    text-align: center;
-    form {
-      display: inline-block;
-    }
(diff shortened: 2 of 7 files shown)


cbeer:
(and I've now done something about the vertical alignment of those checkboxes..)

ggeisler:
:+1:

PR #1085, opened 2015-03-18
Add iframe widget for sir-trevor

### LINKED ISSUE #1079: iframe widget

cbeer:
Fixes #1079


app/views/spotlight/sir_trevor/blocks/_iframe_block.html.erb
```diff
@@ -0,0 +1 @@
+<%= sanitize iframe_block.code, tags: %w(iframe), attributes: %w(id width height src frameborder allowfullscreen sandbox seamless srcdoc) %>
```
jkeck: Other possible attributes we may want: `marginwidth`, `marginheight`, and `scrolling`.  There are probably more, but those will be a good start.

I know marginheight and marginwidth are not HTML5, but I think it's still used from time-to-time.

cbeer: Yeah, I only took the HTML5 list and validated it against a couple `<iframe>` providers. I'll add them if you think we should.

PR #1086, opened 2015-03-18
Render placeholder content when the page is empty

### LINKED ISSUE #85: Default homepage should include some curator-friendly content
Since every exhibit will come with an existing homepage, we should include some sample content on the default homepage that gives the curator a good idea of what is possible in terms of customizing the page for his or her exhibit.

Some mix of widgets and explanatory text (using the text widget).

cbeer:
Fixes #85 

screen shot 2015-03-18 at 11 01 29 am

### DIFF
--- a/app/models/spotlight/home_page.rb
+++ b/app/models/spotlight/home_page.rb
@@ -15,10 +15,6 @@ module Spotlight
     end
 
     private
-    def self.default_content_text
-      "This is placeholder content for the exhibit homepage. Curators of this exhibit can edit this page to customize it for the exhibit."
-    end
-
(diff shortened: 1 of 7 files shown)


jkeck:
Wanted to :shipit: but also wanted to ask @ggeisler if we should put anything in there about adding items (now that we know the default state of the template installer will not have any data).  I think this text was envisioned before we made that change, but not sure.

If you decide that it's a good idea I think you can just feel free to update https://github.com/sul-dlss/spotlight/blob/master/app/views/spotlight/home_pages/_empty.html.erb as you see fit.

PR #1096, opened 2015-03-20
Update installation section in README.

jkeck:


README.md
```diff
@@ -29,46 +29,28 @@ or
...
-$ rake spotlight:initialize
-```
-
-If you installed jettywrapper with Blacklight (above), install Spotlight's demo solr configuration and catalog controller configuration:
-
-```
-$ rake jetty:configure_solr
-```
 
 ## Configuration
```
cbeer: I think we can drop `Blacklight configuration` and `Solr Indexing`, which should be covered in the wiki.

jkeck: :+1:

PR #1097, opened 2015-03-20
Add requirements section to README and move test section to bottom.

jkeck:


README.md
```diff
@@ -7,13 +7,12 @@ Spotlight is open source software that sits on top of a digital repository and e
...
 
-Run tests:
-
-```
-$ rake
-```
+1. Ruby (2.0.0 or greater)
+2. Rails (4.2.0 or greater)
+3. Java (7 or greater) *for solr*
+4. ImageMagick
```
jkeck: Do you think we should add anything about Active Job queuing backends?  It's not a requirement but may be worth mentioning.  I'll definitely add a wiki page about it.

cbeer: I think the wiki page is enough. The default synchronous implementation works fine, so any configuration there is just a environment-specific performance boost.

PR #1106, opened 2015-03-23
Updated Spotlight::FeaturedImage and Masthead to work with uploaded reso...

### LINKED ISSUE #1105: Uploaded resources cannot be used as masthead / page thumbnail images
Uploaded resources store root relative URLs to the image, which breaks `remote_image_url=`, e.g.:

```
Masthead image trying to download a file which is not served over HTTP
```

cbeer:
...urces with root-relative URIs to images

Fixes #1105 (I guess..)


cbeer:
@jkeck any better ideas?

PR #1118, opened 2015-03-30
Provide API access to get or enrich a solr document with exhibit-specific data

cbeer:


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -201,4 +201,13 @@ def current_page_context
       end
     end
   end
+
+  def additional_export_formats document, format
+    super
+
+    format.solr_json do
+      authorize! :update_solr, @exhibit
+      render json: document.to_solr.merge(@exhibit.solr_data)
```
cbeer: It'd be nice if you could get this data from a global endpoint, e.g.  `/catalog/x.solr_json`, but we don't have a list of all the exhibits a document should be in, so I guess we have to settle for this.

app/controllers/spotlight/catalog_controller.rb
```diff
@@ -201,4 +201,13 @@ def current_page_context
       end
     end
   end
+
+  def additional_export_formats document, format
+    super
+
+    format.solr_json do
+      authorize! :update_solr, @exhibit
```
jkeck: Just out of curiosity, what's the rationale behind requiring this level of permission for viewing this json?

cbeer: I don't think we want to expose the raw solr document widely, and with the `update_solr` ability someone can come along and grant a service account access to just the solr json.

PR #1129, opened 2015-04-06
Update resources controller to whitelist resource-type-specific data parameters

cbeer:


app/models/spotlight/resource.rb
```diff
@@ -61,9 +61,7 @@ def becomes_provider
       klass = Spotlight::Resource.class_for_resource(self)
 
       if klass
-        z = self.becomes klass
-        z.type = z.class.to_s
-        z
+        self.becomes! klass
```
cbeer: @jkeck thanks for finding `#becomes!`, it seems like it's exactly what we needed.

app/controllers/spotlight/resources_controller.rb
```diff
@@ -44,7 +44,7 @@ def reindex_all
 
     protected
     def resource_params
-      params.require(:resource).permit(:url)
+      params.require(:resource).permit(:url, data: params[:resource][:data].try(:keys))
```
cbeer: @jkeck what do you think of this approach (from http://edgeguides.rubyonrails.org/action_controller_overview.html#outside-the-scope-of-strong-parameters)?

jkeck: :+1:

PR #1137, opened 2015-04-20
Missing derivatives

cbeer:
@ebenenglish, @ggeisler I'm optimistic this will fix the problems you both reported by email with items without thumbnails.


jkeck:
:+1:

PR #1177, opened 2015-10-06
README updates for development

ndushay:
@cbeer @peetucket 

I may have gone a bit overboard ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@ Read more about what Spotlight is, our motivations for creating it, and how to i
 
 1. Ruby (2.2.0 or greater)
 2. Rails (4.2.0 or greater)
-3. Java (7 or greater) *for solr*
-4. ImageMagick
+3. Java (7 or greater) *for Solr*
+4. ImageMagick (http://www.imagemagick.org/script/index.php) due to [carrierwave](https://github.com/carrierwaveuploader/carrierwave#adding-versions)
 
 ## Installation
 
@@ -67,16 +67,30 @@ Spotlight introduces functionality that depends on being able to send emails to 
 
 See the [Spotlight wiki](https://github.com/sul-dlss/spotlight/wiki) for more detailed information on configuring Spotlight.
 
-## To start the development/test application
+# Developing Spotlight
 
- 1. Clone this repo, `cd` in and run `$ bundle install`
- 2. Start jetty and the dev server: `$ bundle exec rake spotlight:server` (this task will build a Spotlight-based application, start Solr, and run the built-in rails server)
- 3. Visit [http://localhost:3000](http://localhost:3000)
+Spotlight:
+
+* is a Rails engine and needs to be used in the context of a Rails application. We use [engine_cart](https://github.com/cbeer/engine_cart) to create an internal test application at spec/internal.
+* uses Solr as part of its integration tests. We use [jettywrapper](https://github.com/projecthydra/jettywrapper) to manage the Solr instance used for development and test.
+
+Our `$ rake ci` and `$ rake spotlight:server` tasks utilize Solr and the testing rails app automatically.
+
+##  More Information for Developers
+
+* [Contributing to Spotlight](https://github.com/sul-dlss/spotlight/wiki/Contributing-to-Spotlight)
+* [Testing](https://github.com/sul-dlss/spotlight/wiki/Testing)
 
 ## Tests
 
-Run tests:
+### Prerequisites
+
+PhantomJS (https://github.com/teampoltergeist/poltergeist#installing-phantomjs) is an addition requirement for testing javascript.
+
+### Run all the tests:
 
 ```
 $ rake
 ```
+
+This utilizes Solr and the testing rails app automatically.
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 capybara-*.html
 .rspec
 .rvmrc
+.ruby-version
 /.bundle
 /vendor/bundle
 /log/*


ndushay:
hold up on merge - I will add info for installing phantomjs and imagemagick

cbeer:
I like the addition of phantomjs and imagemagick (although I'd be tempted to cut some of verbiage about carrierwave, and just link to https://github.com/carrierwaveuploader/carrierwave#adding-versions)

I'm puzzled about my gut reaction to the "Running Spotlight Gem in Development/Test" section. Talking it over with @jkeck, I wonder if most of that content would fit better as a wiki page. As-is, I feel overwhelmed by details that actually aren't important for using Spotlight in my application or doing "happy path" development (which isn't to say it isn't important, but somehow feels out of scope for this readme).

For example, in Blacklight, we pulled out several wiki pages with developer-specific documentation:

https://github.com/projectblacklight/blacklight/wiki#contributing-to-blacklight

I also wonder if some of this information would actually be a great contribution to the README for `engine_cart` or `jettywrapper`. It doesn't seem particularly specific to spotlight, and if we can make the upstream projects have better documentation (or be easier to use), we could collapse a lot of this to something shorter like, e.g.

> Spotlight uses Solr as part of its integration tests. We use [jettywrapper](https://github.com/projecthydra/jettywrapper) to manage the Solr instance used for development and test. Our `rake ci` and `rake spotlight:server` tasks uses it automatically, but you can refer to its documentation to for how to start and stop the jetty instance outside of those scopes.
>
> Spotlight is a Rails engine and needs to be used in the context of a Rails application. We use engine_cart to create an internal test application for you.

ndushay:
I was also thinking about some of the information being generally useful, either for jettywrapper or for engine_cart.   I can work on moving generic info:.

readme stuff for engine_cart and maybe jettywrapper

and a contributing to spotlight wiki  with the specific testing info for spotlight

possible dev playbook stuff:
testing with solr -> jettywrapper with solr only 
testing with fedora + solr -> jettywrapper with both
dev rails engine gems -> engine_cart

does that seem useful?

cbeer:
:+1:

README.md
```diff
@@ -67,16 +67,29 @@ Spotlight introduces functionality that depends on being able to send emails to
...
-Run tests:
+### Prerequisites
+
+PhantomJS (https://github.com/teampoltergeist/poltergeist#installing-phantomjs) is an addition requirement for testing javascript.
+
+### Run all the tests:
 
 ```
-$ rake
+$ rake ci
```
cbeer: We should be able to just run `rake`, no?

ndushay: yes, but I felt it was better to be explicit about the task.  It seemed like better documentation, somehow.  Otherwise there's a "the default rake task is to run all the tests" assumption baked in.

ndushay: oh yeah - and "rake ci" is reference in the previous text 

"Our `$ rake ci` and `$ rake spotlight:server` tasks utilize Solr and the testing rails app automatically."

I think that was the deciding factor for me

cbeer: I guess I disagree. `rake` is the thing people should run. That it happens to run `rake ci` is an implementation detail.

ndushay: I guess I'm surprised this makes a difference to you. I disagree with you, but it seems like such a minor point.  

*Why* is "rake the thing people should run"?   I don't mind being told to run "bundle install" instead of "bundle" ... it makes it clearer what I'm doing, especially when I am less familiar with the tools.  Someone less familiar with spotlight might be happier knowing that "rake ci" is the explicit command for the tests, in the analogous way.

If I were to change it, I might add a sentence to unify the comment preceding about rake ci  automatically taking care of the rails app and Solr, and that rake alone is going to run the default task, which is ci (only worded much better and shorter).   I don't think there's something sacrosanct about the default rake task being the one to run tests (tho I do approve of the practice!), nor do I think that the README needs to make it obvious that the default rake task is to run the tests (developers can figure that out, IMO).

cbeer: The convention in the ruby community is that the default rake task runs the project's tests, and we should reenforce that expectation here.

The reason it makes a difference is it gives us the opportunity to add in additional types of testing without changing the public API. In fact, we've already done that here when we added rubocop to the default task.

peetucket: My opinion FWIW, I like knowing what is happening under the hood as a developer.  I like the last suggestion of breaking out more detail about how you can run things separately. I don't like it when it says little, assumes I know about the convention of "rake" runs the tests (maybe I am newish Ruby developer and don't), and then on top of it also runs style conventions, code coverage and other stuff that is not obvious.  In a way, it kind of reminds me of Rails way back in the earlier days when you could do scaffolding and magic would happen.  It was cool, but bugged me that I didn't know what was going on.

So I guess I like the idea of a TL/DR at the top which basically lists each command you need to get things running, including tests with "rake".  And then a bit of an extended section explaining how to do things like reload fixtures; clear the state of the jetty, run tests independently of things like rubucop, what "rake" actually does, etc.

cbeer: @peetucket I think @ndushay has addressed those really well in the linked wiki pages:

- https://github.com/sul-dlss/spotlight/wiki/Contributing-to-Spotlight
- https://github.com/sul-dlss/spotlight/wiki/Testing

In the README, I think we should keep things clear and simple for new adopters and drive-by contributors. Deep dives and occasionally used information is better relegated to the wiki for the relatively few users who need/want that.

cbeer:
:+1: looks good to me. Can you squash the commits before we merge it?

ndushay:
... and it's ready. (travis passed, squashed)

cbeer:
:+1:

ndushay:
Just as a followup, I revised the engine_cart README, wrote two wiki pages for jettywrapper and revised the jettywrapper README, in addition to writing the two wiki's for spotlight and the README revisions here.

I'm punting on the developers playbook at this time - I'm documented out.

PR #1180, opened 2015-10-12
Set up a framework for having different indexing strategies.

cbeer:
Currently this just has a local indexing strategy, but it should be
simple to add alternatives.


cbeer:
I was just cleaning up old branches and stumbled on this one. I can't figure out, off-hand, if it is still useful. I rebased it just in case.

ndushay:
@jcoyne do you care about this branch?

cbeer:
I doubt @jcoyne cares. I think it's more a question about whether we care about this feature, and I think it's debatable (and would benefit from articulating why you'd use this feature).

ndushay:
perhaps this question will be morphed into "do we need special indexing for SALT, and if so, how should we do it?"   in which case I'd close this PR, and keep the branch until we've gotten past that point.

cbeer:
Wrong issue?

jcoyne:
I haven't really had a chance to look at spotlight lately. So, no I don't care. But this seems like a good change nevertheless. :+1:

PR #1191, opened 2015-10-14
Removes SU list notifications for travis builds

mejackreed:
Do we really need to alert this list for travis builds?


cbeer:
:+1:. Authors will still get notified, and that's good enough for me.

PR #1198, opened 2015-10-16
Fix exhibit serializer handling for normal resources

cbeer:


peetucket:
@cbeer Looks good to us but we don't have sufficient background knowledge to accept.  Do you want us to merge anyway?

PR #1202, opened 2015-10-20
Pull batch indexing and exhibit-specific data logic up from spotlight-dor-resources

cbeer:
I hope this will let us simplify `Spotlight::Resources::DorResource` to just the glue code with `gdor-indexer`.

### DIFF
--- a/app/models/spotlight/resource.rb
+++ b/app/models/spotlight/resource.rb
@@ -28,10 +28,15 @@ module Spotlight
 
     ##
     # @abstract
-    # Convert the resource into a new solr document.
+    # Convert this resource into zero-to-many new solr documents. The data here
+    # should be merged into the resource-specific {#to_solr} data.
+    #
+    # @return [Hash] a single solr document hash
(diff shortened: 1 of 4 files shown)


cbeer:
I think this should simplify Spotlight::Resources::DorResource down to something like:

```ruby
module Spotlight::Resources
  class DorResource < Spotlight::Resource
    def to_solr
      return to_enum :to_solr unless block_given?
      
      base_doc = super
      
      [resource, (resource.items if resource.collection?)].flatten.compact.each do |res|
        yield base_doc.merge(Spotlight::Dor::Resources.indexer.solr_document(res))
      end
    end
    
    private
    
    def resource
      @resource ||= Spotlight::Dor::Resources.indexer.resource doc_id
    end
  end
end
```

ndushay:
@peetucket should there be a ticket for the simplification Chris indicates above for Spotlight::Resources::DorResource

cbeer:
:+1:

peetucket:
Simplication ticket: sul-dlss/spotlight-dor-resources#6

ndushay:
:+1:

PR #1209, opened 2015-11-09
Add benchmarking for reindexing Spotlight::Resources

cbeer:


app/models/spotlight/resource.rb
```diff
@@ -52,16 +45,13 @@ def reindex_with_lock
...
-          blacklight_solr.update params: { commitWithin: 500 },
-                                 data: batch.reject(&:blank?).map { |doc| doc.reverse_merge(existing_solr_doc_hash(doc[unique_key]) || {}) }.to_json,
-                                 headers: { 'Content-Type' => 'application/json' }
+      benchmark "Reindexing #{self} (batch size: #{batch_size})" do
+        run_callbacks :index do
+          documents_to_index.each_slice(batch_size) do |batch|
+            blacklight_solr.update params: { commitWithin: 500 },
+                                   data: batch.to_json,
+                                   headers: { 'Content-Type' => 'application/json' }
+          end
```
ndushay: is there a way to find out the HTTP response code?  Would it help?

cbeer: I think RSolr throw a big exception if this fails.

ndushay: I guess what I'm asking is:  will an error get to the logs?

ndushay: looks like it doesn't in gdor either - meh.   moving on.

PR #1211, opened 2015-11-10
Record additional metadata about resource indexing

cbeer:


.rubocop.yml
```diff
@@ -37,4 +37,8 @@ Metrics/AbcSize:
   Max: 20
 
 Metrics/ModuleLength:
-  Max: 120
\ No newline at end of file
+  Max: 120
+
+Metrics/ClassLength:
+  Exclude:
+    - 'app/models/spotlight/resource.rb'
```
cbeer: Happy to discuss this exception.. I tried a couple different refactorings for `Spotlight::Resource` and none of them seemed to improve readability.

ndushay: I'm fine with the exception.

app/models/spotlight/resource.rb
```diff
@@ -142,5 +153,17 @@ def unique_key
         :id
       end
     end
+
+    def add_to_index(batch)
```
ndushay: would this be clearer as "send_to_solr" ?

app/models/spotlight/resource.rb
```diff
@@ -74,13 +90,9 @@ def needs_provider?
       type.blank?
     end
 
-    def save_and_commit
+    def save_and_index
```
ndushay: so the save is for the db and the index is for Solr?  Comment might help.

app/models/spotlight/resource.rb
```diff
@@ -42,24 +44,38 @@ def reindex_with_lock
...
+             last_indexed_estimate: documents_to_index.size,
+             last_indexed_count: count,
+             last_index_elapsed_time: time_end - time_start)
+    end
+
     ##
     # Index the result of {#to_solr} into the index in batches of {#batch_size}
+    #
+    # @return [Integer] number of records indexed
     def reindex
```
ndushay: if we always call something called "reindex" then isn't it really "index" or "write_to_solr" or something?  If there's no diff between the first write and subsequent updates?

ndushay: and don't we cleverly have code somewhere with a method ".to_solr" which really means "to a hash suitable for sending to solr" ?

app/models/spotlight/resource.rb
```diff
@@ -74,13 +90,9 @@ def needs_provider?
       type.blank?
     end
 
-    def save_and_commit
+    def save_and_index
       save.tap do
```
ndushay: I confess #tap is new to me, and this line is confusing to me.  "save" would be save an ActiveModel object to a database, but now it has .tap on the end?

spec/models/spotlight/resource_spec.rb
```diff
@@ -48,6 +48,30 @@ def to_solr
...
         subject.reindex
       end
+
+      it 'returns the number of indexed objects' do
+        allow(subject.send(:blacklight_solr)).to receive(:update)
+
+        expect(subject.reindex).to eq 1
+      end
+
+      it 'triggers a solr commit' do
```
ndushay: so is this implying that a commit will occur after each record?  If so, could we do a soft commit after each record, and a hard commit at the end of indexing?

spec/models/spotlight/resource_spec.rb
```diff
@@ -60,10 +84,21 @@ def to_solr
...
+    before do
+      allow(subject.send(:blacklight_solr)).to receive(:update)
+    end
+
+    it 'saves the object and commit to solr' do
+      expect(subject).to receive(:save).twice
+      subject.save_and_index
+    end
+
+    it 'reindexs after save' do
```
ndushay: typo: reindexes

app/models/spotlight/resource.rb
cbeer: :ok_hand:

spec/models/spotlight/resource_spec.rb
```diff
@@ -60,10 +84,21 @@ def to_solr
...
-  it 'reindexs after save' do
-    expect(subject).to receive(:reindex)
-    subject.data_will_change!
-    subject.save!
+  describe '#save_and_index' do
+    before do
+      allow(subject.send(:blacklight_solr)).to receive(:update)
+    end
+
+    it 'saves the object and commit to solr' do
```
ndushay: this test doesn't show a solr commit but the text implies it would?

spec/models/spotlight/resource_spec.rb
cbeer: We're only committing after all indexing. Let me update the test to try to demonstrate that.

spec/models/spotlight/resource_spec.rb
cbeer: Thanks; we probably don't need to test if `commit` is triggered here, as it's tested as part of `#reindex`. I'll update the test.

app/models/spotlight/resource.rb
```diff
@@ -74,72 +35,141 @@ def needs_provider?
...
-    end
 
-    protected
+      private
 
-    def blacklight_solr
-      @solr ||= RSolr.connect(connection_config)
-    end
+      ##
+      # Get any exhibit-specific metadata stored in e.g. sidecars, tags, etc
```
ndushay: Did this comment line get separated from the right method (def exhibit_specific_solr_data) ?

app/models/spotlight/resource.rb
```diff
@@ -74,72 +35,141 @@ def needs_provider?
...
+          Spotlight::Engine.config.resource_global_id_field => (to_global_id.to_s if persisted?),
+          Spotlight::SolrDocument.resource_type_field => self.class.to_s.tableize
+        }
+      end
 
-    def exhibit_specific_solr_data
-      exhibit.solr_data if exhibit
+      def document_model
+        exhibit.blacklight_config.document_model if exhibit
+      end
```
ndushay: I'm a huge fan of  end # concern GeneratingSolrDocuments on the approp end statement when it's more than a screen.

app/models/spotlight/resource.rb
```diff
@@ -74,72 +35,141 @@ def needs_provider?
...
+      def add_to_index(batch)
+        blacklight_solr.update params: { commitWithin: 500 },
+                               data: batch.to_json,
+                               headers: { 'Content-Type' => 'application/json' }
+      end
+
+      def commit
+        blacklight_solr.commit
+      rescue => e
+        Rails.logger.warn "Unable to commit to solr: #{e}"
```
ndushay: similarly:   end  # concern Indexing   would be great.  Easy to lose my place in indentation

app/models/spotlight/resource.rb
cbeer: Looks right to me. I think it's just funny in the diff.

app/models/spotlight/resource.rb
```diff
@@ -74,72 +35,141 @@ def needs_provider?
...
+
+      def connection_config
+        Blacklight.connection_config
+      end
+
+      def batch_size
+        Spotlight::Engine.config.solr_batch_size
+      end
+
+      def add_to_index(batch)
```
ndushay: I can live with "add_to_index" but I think "save_to_index" or "write_to_index" would be clearer.

ndushay: especially since this is for reindexing and adding, no?

spec/controllers/spotlight/resources_controller_spec.rb
```diff
@@ -53,8 +53,7 @@
     describe 'POST create' do
       let(:blacklight_solr) { double }
       it 'create a resource' do
-        allow_any_instance_of(Spotlight::Resource).to receive(:reindex)
-        expect(blacklight_solr).to receive(:commit)
+        expect_any_instance_of(Spotlight::Resource).to receive(:reindex)
```
ndushay: here again it's curious that POST create   is  a "reindex"  -- normally, I would think PATCH or PUTS  would be a reindex and  POST would be more of an "add to index"

app/models/spotlight/resource.rb
cbeer: Sure, `write_to_index` is a good name.

spec/controllers/spotlight/resources_controller_spec.rb
cbeer: The `POST` here is creating a resource, and one of the side-effects of creating a resource is to trigger (re)indexing.

PR #1213, opened 2015-11-11
Use asynchronous indexing with immediate reporting

cbeer:


app/models/spotlight/resource.rb
```diff
@@ -40,7 +41,13 @@ def needs_provider?
     #
     # @param [Hash] All arguments will be passed through to ActiveRecord's #save method
     def save_and_index(*args)
-      save(*args) && reindex
+      save(*args) && reindex_later
```
ndushay: as in reindex at an undetermined later time?

cbeer: Yep. This uses the same `ActiveJob` step that we're using when you click the "Reindex" button.

PR #1216, opened 2015-11-12
spotlight:import task should trigger a reindex

cbeer:


ndushay:
this is the one that you were asking me about in person, yes?   If so, let's discuss the philosophy w Gary.

cbeer:
:ok_hand: @ggeisler @peetucket the question I have is:

We certainly want to trigger a reindex after we initially import an exhibit. Do we also want to automatically reindex when we deploy new application code?

There are pretty good arguments either way:

 - the application code may add new indexing, and add desired functionality to the application, but
 - we may be better separating deploying the application from indexing to reduce unexpected changes (and stop us from wondering "was it application code or new indexing that broke X")

If there's not a clear winner, we should just merge this and revisit as needed.

ggeisler:
Doesn't seem like a clear winner to me, although I would think that adding new indexing-related code would not happen all that often in the future, which would make the argument for automatically reindexing after deploying new code less compelling (it just wouldn't be necessary or relevant to reindex the vast majority of the time new code is deployed).

cbeer:
That's a good point. We should ship this, and revisit reindexing on deploy if/when we need to.

peetucket:
I concur with @ggeisler FWIW.

PR #1223, opened 2015-11-18
Provide an index of all published exhibits

cbeer:
Instead of automatically creating a default exhibit, create an index of exhibits (similar to the exhibits.stanford.edu landing page) with instructions on how to create a new exhibit.

By making the default exhibit unnecessary, this implied some other changes, including:

 - hiding the search box when outside an exhibit context (whereas before we'd fallback on searching the default exhibit)
 - removing the `spotlight_root` routing, instead routing to the `spotlight/exhibits#index`
 - no longer generating the `default` attribute (but not bothering to create a migration to remove it, in case it still has significance downstream)
 - defaulting to private exhibits, so they don't automatically appear on the new index page

If there is only one exhibit, we redirect from the landing page directly to the only exhibit. This should allow us to maintain a backward compatible behavior for single-exhibit users (like exhibits.stanford.edu, for now).

- [x] migrate index view and styles from sul-dlss/exhibits_request and refactor to better match spotlight conventions
- [x] update styling to better match bootstrap defaults, while still allowing us to style it downstream.
- [x] write a feature test that visits the exhibits index

Connects to sul-dlss/sul_exhibits_template#100

screen shot 2015-11-18 at 7 43 07 am


cbeer:
The loss in coverage here seems to be from removing well-covered lines of code.

cbeer:
@jkeck any idea how to deal with the hound violation about vendor prefixes, or should we agree to ignore it?

jkeck:
I think we'll have to ignore it.  I have found a few other cases where autoprefixer unfortunately does not support things I want to use and haven't found a better way around it. :/

PR #1233, opened 2015-11-19
Make all spotlight variables '\!default', so they can be overridden downstream easily

cbeer:


ndushay:
go with hound?  or merge as is?

PR #1243, opened 2015-11-20
Trigger a reindex job after importing an exhibit; fixes #1238

### LINKED ISSUE #1238: Importing new Spotlight::Resources into an exhibit should trigger a reindex

cbeer:


jkeck:
:+1:  Looks like this needs a rebase though.

PR #1245, opened 2015-11-20
Add Uploaded Item Row Block.

### LINKED ISSUE #1206: Browse category widget doesn't respect curator-selected order

jkeck:
Closes #1206 

uploaded-items-block


jkeck:
I think we should ignore the Hound comments on this PR.  They are:

1. Complaints about globals being undefined (they are defined).
2. Complaints about areas where I'm being consistent with the surrounding code. (trying not to introduce inconsistency)
3. Complaints about sir-trevor selectors not using hyphens.

cbeer:
:+1: 

I've created #1246 and #1247 to suggest changes to the linter configurations.

At sprint planning, maybe we should float the idea of cleaning up the javascript and css. I think we've learned new approaches to implementation and testing, and we might want to apply them here.

PR #1250, opened 2015-11-20
Add a DefaultThumbnailJob and hook it up to Exhibit and Search

### LINKED ISSUE #1004: Move getting default thumbnails and the like into a job (that can be performed asynchronously..)
Spotlight creates default thumbnails for browse categories (aka saved searches) using a `before_create` hook:

https://github.com/sul-dlss/spotlight/blob/master/app/models/spotlight/search.rb#L21

Now that we're using ActiveJob, this could be done better by extracting it to an asynchronous job and invoked after creating the saved search.

- [ ] Create an `ActiveJob::Base` subclass that receives a search and sets the default thumbnail, if needed
- [ ] Double-check the `browse#index` action to ensure it displays ok without a thumbnail and/or update Spotlight::Search to provide a placeholder thumbnail instead.

### LINKED ISSUE #1225: Provide a default thumbnail for an exhibit
With #1223, we now use the exhibit thumbnail within spotlight. Just as we provide a default thumbnail for `Spotlight::Search`, we should provide a default thumbnail for `Spotlight::Exhibit`.

https://github.com/sul-dlss/spotlight/blob/master/app/models/spotlight/search.rb#L21

(see #1004 first, which suggests a better approach for Spotlight::Search)

jkeck:
Closes #1004 
Closes #1225 

This adds a `DefaultThumbnailable` concern that is mixed into `Exhibit` and `Search`. This concern adds an `after_create` hook that fetches a default thumbnail asynchronously via the `ActiveJob` queue. Each class that mixes in the `DefaultThumbnailable` concern has to implement how it fetches the default thumbnail (by implementing a `set_default_thumbnail` method).

browse-categories-w-no-thumb

### DIFF
--- a/app/assets/stylesheets/spotlight/_browse.scss
+++ b/app/assets/stylesheets/spotlight/_browse.scss
@@ -25,13 +25,17 @@ $image-overlay-bottom-margin: $padding-large-vertical * 3;
   }
 
   .image-overlay {
-    img {width: 100%;}
-    position: relative;
+    border: 2px solid $gray-light;
+    border-radius: $border-radius-large;
     margin-bottom: $image-overlay-bottom-margin;
     max-height: 240px;
-    border-radius: $border-radius-large;
-    border: 2px solid $gray-light;
+    min-height: 140px;
     overflow: hidden;
+    position: relative;
+
+    img {
+      width: 100%;
+    }
   }
 
   .text-overlay {
--- a/app/jobs/spotlight/default_thumbnail_job.rb
+++ b/app/jobs/spotlight/default_thumbnail_job.rb
@@ -1,0 +1,14 @@
+module Spotlight
+  ###
+  # Calls the #set_default_thumbnail method
+  # on the object passed in and calls save
+  ###
+  class DefaultThumbnailJob < ActiveJob::Base
+    queue_as :default
+
+    def perform(thumbnailable)
+      thumbnailable.set_default_thumbnail
+      thumbnailable.save
+    end
+  end
+end
--- a/app/models/spotlight/exhibit.rb
+++ b/app/models/spotlight/exhibit.rb
@@ -49,6 +49,7 @@ module Spotlight
     after_create :initialize_config
     after_create :initialize_browse
     after_create :initialize_main_navigation
+    include Spotlight::DefaultThumbnailable
 
     scope :published, -> { where(published: true) }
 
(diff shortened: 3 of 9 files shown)


app/jobs/spotlight/default_thumbnail_job.rb
```diff
@@ -0,0 +1,14 @@
...
+  ###
+  # Calls the #set_default_thumbnail method
+  # on the object passed in and calls save
+  ###
+  class DefaultThumbnailJob < ActiveJob::Base
+    queue_as :default
+
+    def perform(thumbnailable)
+      thumbnailable.set_default_thumbnail
+      thumbnailable.save
```
jkeck: Should I require the class that implements `set_default_thumbnail` decide if it should call `save` or not?  This is currently restricted to only being re-used by objects that implement `save`.

app/models/spotlight/exhibit.rb
```diff
@@ -85,6 +86,10 @@ def searchable?
       blacklight_config.search_fields.any? { |_k, v| v.enabled && v.include_in_simple_select != false }
     end
 
+    def set_default_thumbnail
+      self.thumbnail ||= searches.first.try(:thumbnail)
```
jkeck: Is this legit to just set the thumbnail as the search's thumbnail?  Should we dup/clone it instead?

app/assets/stylesheets/spotlight/_browse.scss
```diff
@@ -25,13 +25,17 @@ $image-overlay-bottom-margin: $padding-large-vertical * 3;
...
+    border-radius: $border-radius-large;
     margin-bottom: $image-overlay-bottom-margin;
     max-height: 240px;
-    border-radius: $border-radius-large;
-    border: 2px solid $gray-light;
+    min-height: 140px;
     overflow: hidden;
+    position: relative;
+
+    img {
```
jkeck: I believe there is/will be a proposal to modify our Hound configuration to explicitly point to a scss-lint config which defaults to depth of 3.

app/jobs/spotlight/default_thumbnail_job.rb
cbeer: Explicitly calling `#save` makes sense to me. If it wasn't called, there's a good chance we'd forget somewhere along the line.

app/models/spotlight/exhibit.rb
cbeer: Let's try it, and see what happens. It seems reasonable enough, I think.

cbeer:
Rebase?

PR #1255, opened 2015-11-23
Auto-populate the exhibit slug from the title, unless a slug is explicitly given

cbeer:


app/assets/javascripts/spotlight/exhibits.js
```diff
@@ -1,4 +1,18 @@
 Spotlight.onLoad(function() {
+
+  // auto-fill the exhibit slug on the new exhibit form
+  $('#new_exhibit').each(function() {
+    $('#exhibit_title').on('change keyup', function() {
+      $('#exhibit_slug').attr('placeholder', URLify($(this).val(), $(this).val().length));
```
cbeer: I've addressed the "is not defined" message. I think the "new prefix" is bogus -- `URLify` is (a bad name for) a method from a 3rd party library.

PR #1259, opened 2015-11-23
update to fix potentially flaky css selector; allow local .ruby-version file

peetucket:


cbeer:
Should these changes be in two separate commits? 

Could you also expand the commit message to include the rationale behind `:first` -> `:first-child`? It seems like valuable context for the future (I know I have moments of remembering fixing something in a project, and the commit message is both a good way to discover it and remind me why we made those changes)

jkeck:
Sorry @peetucket, going to need a rebase on this.  :+1: though.

cbeer:
:+1: Thanks, much appreciated.

ndushay:
as the suggestions were implemented, and this was very tiny so there didn't seem anything else possibly to worry about, I merged this one.

PR #1265, opened 2015-11-24
Add migration to remove unused default column from spotlight_exhibits

cbeer:


cbeer:
In 0edc776015231a59bdb2faba1d7659116cad9055, we stopped using the `default` attribute and removed it from new migrations. This removes it from an existing database as well (which was thought unnecessary, but turns out to be harmful for import/export for existing exhibits).

PR #1267, opened 2015-11-24
Update custom field names to not contain the field name

cbeer:
Instead, construct the solr field name with the exhibit prefix on-the-fly for solr.


cbeer:
Fixes #1237

PR #1280, opened 2015-11-25
Advertise the spotlight factories to downstream applications

### LINKED ISSUE #1270: Consider advertising factories in a railtie
See the [PR in FactoryGirlRails](https://github.com/thoughtbot/factory_girl_rails/pull/42)

I *think* we just need to add something along the lines of

```ruby
initializer 'spotlight.factories', after: 'factory_girl.set_factory_paths' do
  FactoryGirl.definition_file_paths << File.expand_path('../../../spec/factories', __FILE__) if defined?(FactoryGirl)
end
```

in `config/engine.rb` and then apps (and gems I believe) that require Spotlight will have access to the Spotlight defined factories in their tests.

cbeer:
Fixes #1270

### DIFF
--- a/lib/spotlight/engine.rb
+++ b/lib/spotlight/engine.rb
@@ -39,6 +39,10 @@ module Spotlight
       OEmbed::Providers.register_all
     end
 
+    initializer 'spotlight.factories', after: 'factory_girl.set_factory_paths' do
+      FactoryGirl.definition_file_paths << File.expand_path('../../../spec/factories', __FILE__) if defined?(FactoryGirl)
+    end
+
     def self.catalog_controller
       Spotlight::Engine.config.catalog_controller_class.constantize
     end


cbeer:
In local testing, this worked like a charm. Once we've merged and released this, I'll update sul-exhibits-template, at least, to use it.

ndushay:
i wanna use it in my gem.  how do i do that?

jkeck:
@ndushay, my hope is that you can just use the exhibits by name (e.g. `FactoryGirl.create(:exhibit)`, but make sure to remove the one you defined yourself)

Let me know if it doesn't work (which would most likely be because you're developing a gem) and I'll try to think of ways to possibly make them available to gems as well.

ndushay:
I will eagerly await the new gem release of spotlight with this PR included!

PR #1290, opened 2015-11-26
Allow admins to invite users who don't have user accounts in Spotlight

jkeck:
Not sure if you all want to merge this in as is or let somebody build on-top of it.  At the very least I think we need to make modifications to the invite email because it's basically useless in its current form since it does not include any exhibit context.

mixed-users

inviting-user


app/controllers/spotlight/roles_controller.rb
```diff
@@ -30,12 +30,36 @@ def update_all
...
+      end
+    end
+
+    def invite
+      params.require(:user)
+      params.require(:role)
+      user = ::User.invite!(email: params[:user])
+      role = Spotlight::Role.create(exhibit: current_exhibit, user: user, role: params[:role])
+      if role.save
+        redirect_to :back, notice: t(:'helpers.submit.role.updated')
```
jkeck: We may want to make this a different piece of text.

jkeck:
Okay, I think this is ready for :eyes:.

config/routes.rb
```diff
@@ -88,6 +88,8 @@
 
     resources :roles, path: 'users', only: [:index, :create, :destroy] do
       collection do
+        get :invite
```
cbeer: Should `invite` really use `GET`? It strikes me more as a `POST` or `PUT`.

config/locales/spotlight.en.yml
```diff
@@ -466,10 +472,15 @@ en:
         bookmarklet: "%{application_name} widget"
       reindexing_in_progress: "Reindexing all resources"
     roles:
+      edit_fields:
+        invite_html: "This user does not yet exist. Would you lke to send them an %{link}?"
```
cbeer: `s/lke/like/`?

config/locales/spotlight.en.yml
```diff
@@ -261,6 +261,12 @@ en:
           non_repo_item: "Add non-repository item"
       new:
         header: Import items
+    invitation_mailer:
+      invitation_instructions:
+        hello: "Hello %{email}!"
+        someone_invited_you: "The Exhibits Administrator has invited to be a member of '%{exhibit_name}' at %{url}. You can accept it through the link below.</p>"
```
cbeer: Since these contain HTML, should the keys be suffixed with `_html`?

app/controllers/spotlight/roles_controller.rb
```diff
@@ -30,12 +30,39 @@ def update_all
...
+      #  app uses html status code to act on response
+      if ::User.where(email: exists_params).present?
+        render json: { message: 'User exists' }
+      else
+        render json: { message: 'User does not exist' }, status: :not_found
+      end
+    end
+
+    def invite
+      params.require(:user)
```
cbeer: Should these requires be moved into e.g. `#invite_params`, and have it return e.g. `params.permit(:user, :role)` to match the other methods?

app/controllers/spotlight/roles_controller.rb
```diff
@@ -30,12 +30,39 @@ def update_all
       end
     end
 
+    def exists
+      # note: the messages returned are not shown to users and really only useful for debug, hence no translation necessary
+      #  app uses html status code to act on response
+      if ::User.where(email: exists_params).present?
+        render json: { message: 'User exists' }
```
cbeer: These strings are just for convenient debugging, I guess?

config/routes.rb
jkeck: This interaction is triggered from an `Invite` link in a form (which is why it made sense at the time to make it a `GET` request).  Otherwise we'll need to have a button that is only triggerable via ajax to submit a `POST`/`PUT`.

I agree that it doesn't quite feel like a `GET` but the interaction is designed that way (and seems like a lot of trouble to do it otherwise, unless I'm missing something obvious).

app/controllers/spotlight/roles_controller.rb
```diff
@@ -30,12 +30,39 @@ def update_all
       end
     end
 
+    def exists
+      # note: the messages returned are not shown to users and really only useful for debug, hence no translation necessary
+      #  app uses html status code to act on response
+      if ::User.where(email: exists_params).present?
```
cbeer: I wonder if we ought to make `::User` configurable at the engine level. Downstream applications might choose a different model name.

app/assets/javascripts/spotlight/check_user_existence.js
```diff
@@ -0,0 +1,74 @@
...
+      roleSelect().on('change', updateNoUserNoteLink);
+      submitButton().prop('disabled', true);
+    }
+
+    function updateNoUserNoteLink() {
+      var link = noUserNote().find('a');
+      var originalHref = link.data('inviteUrl');
+      var userName = target.val();
+      var role = roleSelect().val();
+      link.attr('href', originalHref + '?user=' + userName + '&role=' + role);
```
cbeer: Is there any sort of CGI escaping we need to do here?

config/locales/spotlight.en.yml
jkeck: Actually, I don't think this particular key should have HTML.  The `<p>` is in the mailer body (and this is actually causing a duplication `</p>`.  I'll remove the html and add a `_html` to the keys that do include markup.

app/assets/javascripts/spotlight/users.js
```diff
@@ -27,13 +27,15 @@
     }
 
     function clear_errors(element) {
-      element.find('.has-error').removeClass('has-error');
-      element.find('.help-block').remove(); // Remove the error messages
+      element.find('.has-error')
```
cbeer: Is this really equivalent to the old behavior? I'm not too clear on how jquery chaining works, I guess.

config/locales/spotlight.en.yml
jkeck: :+1:

app/views/devise/invitations/edit.html.erb
```diff
@@ -0,0 +1,14 @@
+<h2><%= t 'devise.invitations.edit.header' %></h2>
+
+<%= form_for resource, :as => resource_name, :url => invitation_path(resource_name), :html => { :method => :put } do |f| %>
```
cbeer: Use Ruby 1.9-style hashes?

app/views/devise/invitations/edit.html.erb
```diff
@@ -0,0 +1,14 @@
+<h2><%= t 'devise.invitations.edit.header' %></h2>
```
cbeer: What's different in this file? It looks the same as the upstream one to me?

app/views/devise/invitations/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h2><%= t "devise.invitations.new.header" %></h2>
```
cbeer: Same question here. What's customized about this?

app/views/spotlight/roles/_edit_fields.html.erb
```diff
@@ -1,5 +1,11 @@
         <tr data-edit-for="<%= f.object.new_record? ? 'new' : f.object.id %>">
-          <td><%= f.text_field :user_key, hide_label: true %></td>
+          <td>
+            <%= f.email_field :user_key, hide_label: true, disabled: f.object.persisted?, data: { behavior: 'check-user-existence' } %>
+            <span style='display:none' class='help-block' data-behavior='no-user-note'>
```
cbeer: Is the inline `display: none` to stop a FOUC, or??

app/views/spotlight/roles/_edit_fields.html.erb
```diff
@@ -1,5 +1,11 @@
...
+              <%= t('.invite_html', link: link_to(t('.invite_link'), spotlight.invite_exhibit_roles_path(current_exhibit), data: { :'invite-url' => spotlight.invite_exhibit_roles_path(current_exhibit) })) %>
```
cbeer: We're requiring Ruby 2.2, so I think:

```
'invite-url': spotlight.inv...
```

is valid here.

app/assets/javascripts/spotlight/users.js
jkeck: It is because `removeClass` is going to return the element that the class was removed on (which is the `.form-group`) and the `.help-block` is a child of that `.form-group`.

config/locales/spotlight.en.yml
```diff
@@ -72,7 +72,7 @@ en:
       role:
         updated: "User has been updated."
         destroyed: "User has been removed."
-        batch_error: "There was a problem saving the users."
+        batch_error: "There was a problem saving the user(s)."
```
cbeer: Can we use i18n's pluralization support to handle both these cases a little cleaner?

lib/generators/spotlight/install_generator.rb
```diff
@@ -1,3 +1,4 @@
+# rubocop:disable ClassLength
```
cbeer: Do you want to just disable `Metrics/ClassLength` for all generators? Or, maybe there's some sub-generators we ought to break out of here?

app/views/devise/invitations/new.html.erb
jkeck: @peetucket care to comment?

app/views/devise/invitations/edit.html.erb
jkeck: @peetucket care to comment?

config/routes.rb
cbeer: Isn't there magic `jquery-ujs` stuff that'll turn a simple link into a form?

config/locales/spotlight.en.yml
cbeer: :+1:

app/assets/javascripts/spotlight/users.js
cbeer: :+1:

cbeer:
And, it looks like a rebase is in order.

app/controllers/spotlight/roles_controller.rb
jkeck: :+1:

app/assets/javascripts/spotlight/check_user_existence.js
jkeck: :+1:

app/controllers/spotlight/roles_controller.rb
jkeck: I'm going to do this in a separate PR since it touches a lot of other code paths.

config/locales/spotlight.en.yml
jkeck: I'm going to punt on this and add a ticket for it for later.

config/routes.rb
jkeck: :+1:

lib/generators/spotlight/install_generator.rb
jkeck: Perhaps I'm being dense, but how would I disable a particular rule for just generators?  I'm guessing we don't want to disable all rules for generators, but I'm unfamiliar with how we would do that with particular rules.

Also, I suppose we could break this out into separate generators.  I'm trying to think of where the logical divide between generators would be.  So do you think something like a dependencies generator (add dependencies and run their generators) and a generator that moves/copies/and injects into app-level files?

cbeer: e.g.

```
Metrics/ClassLength:
  Exclude:
    - 'lib/generators/spotlight/*'
```

Yeah, I don't know what the right separation is. Maybe revisit it later?

config/locales/spotlight.en.yml
cbeer: :+1:

app/controllers/spotlight/roles_controller.rb
cbeer: :+1:

lib/generators/spotlight/install_generator.rb
jkeck: Ah! I like that, I actually didn't know that syntax was possible, but makes sense!

jkeck:
I think I have addressed everything and rebased.  Please let me know if I'm missing anything else, and thanks for all the advice!

cbeer:
Do you want to squash any of those commits?

jkeck:
Bah, yeah, sorry, I was struggling to figure out how to cohesively squash those down into something that made sense.  Hopefully what I did is workable.

cbeer:
:+1:

PR #1294, opened 2015-11-30
Make exhibit cards more responsive at small and medium breakpoints

### LINKED ISSUE #1289: Make exhibit cards adjust better to different browser breakpoints
The exhibit cards don't work too well at small and medium breakpoints (example below). We can improve by:

- Adjust the image to fit to the card width at small and medium breakpoints
- Adjust the card height at small and medium breakpoints so that the image remains square
- ~~Add a site thumbnail-specific jcrop_options method to `jcrop_help.rb` so that site thumbnails are cropped at a square aspect ratio (they are currently cropped using the same method used for other thumbnails, like the browse categories, and we don't want to change their aspect ratio)~~ (moved to #1295)

These changes would ensure that the exhibit cards show pretty much exactly what the curator cropped using the site thumbnail cropping feature, which would be nice since this is the image that visually represents the exhibit.

blacklight

cbeer:
Fixes #1289 

exhibit-card

### DIFF
--- a/app/assets/stylesheets/spotlight/_exhibits_index.scss
+++ b/app/assets/stylesheets/spotlight/_exhibits_index.scss
@@ -50,7 +50,7 @@
   img {
     height: $exhibit-card-image-size;
     object-fit: cover;
-    width: $exhibit-card-image-size;
+    width: 100%;
   }
 
   .card-title {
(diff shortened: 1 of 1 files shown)


ndushay:
in the "please educate me" category:  I assume we have no way to write tests for this sort of code?

PR #1297, opened 2015-12-01
Use the exhibit search builder to retrieve an exhibit's solr documents

cbeer:
Instead of using custom logic for picking out all the exhibit's solr documents, use the application's `SearchBuilder` class, which already has that logic.


ndushay:
This PR looks good.  I'm not sure I fully understand all of it, but it's cool that there's a SearchBuilder class, I think.

PR #1299, opened 2015-12-02
Add spotlight:export rake task for exporting an exhibit as JSON

cbeer:
Support a better exhibit import/export experience from the command line by adding the `spotlight:export` rake task, and updating `spotlight:import` to accept a slug as an argument to the task.


ndushay:
what does exporting an exhibit as JSON mean?

cbeer:
Same thing as you get when you export the exhibit from the UI. It's all the stuff needed to recreate the exhibit elsewhere.

PR #1301, opened 2015-12-02
Change the search field selector to use a smaller font size, so it aligns better with the query input

cbeer:
Before:

screen shot 2015-12-02 at 2 31 04 pm

After:

screen shot 2015-12-02 at 2 30 59 pm


ggeisler:
:+1:

PR #1304, opened 2015-12-03
Copy bootstrap's approach to navbar styling to style the transparent navbar

cbeer:
... that appears over a masthead image. This should make it easier for downstream applications to retheme the navbar, and ensure we're consistently applying styles.(with the ad-hoc approach, we're running into trouble with the `:focus` state not being transparent..)

Before:

screen shot 2015-12-03 at 10 48 26 am

After:

screen shot 2015-12-03 at 10 49 55 am


cbeer:
I suggest we ignore Hound's comments about `navbar.scss`. The file was copied verbatim from upstream, and simply adapted to handle our transparent theme.

ggeisler:
:+1:

PR #1310, opened 2015-12-04
Allow the curator to select the default sort order for a facet

### LINKED ISSUE #1283: Set the default sort order per facet
Admins should be able to set the default sort order of a facet (e.g. "Year" should be sortable by A-Z/number instead of frequency of value occurrence).

cbeer:
Fixes #1283

screen shot 2015-12-04 at 8 54 24 am


ggeisler:
:

PR #1321, opened 2015-12-09
postgres does not support blob field type

ronan-mch:
Postgres doesn't have a blob type, thus this migration breaks when install Spotlight in an existing Rails app using Postgres. This fixes that issue and adds the previously missing rollback task.

### DIFF
--- a/db/migrate/20151110082345_add_metadata_to_spotlight_resource.rb
+++ b/db/migrate/20151110082345_add_metadata_to_spotlight_resource.rb
@@ -1,5 +1,12 @@
 class AddMetadataToSpotlightResource < ActiveRecord::Migration
   def up
-    add_column :spotlight_resources, :metadata, :blob
+    # postgresql does not have a blob type
+    conn_type = connection.adapter_name.downcase.to_sym
+    field_type = (conn_type == :postgresql) ? :bytea : :blob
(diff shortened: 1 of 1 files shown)


db/migrate/20151110082345_add_metadata_to_spotlight_resource.rb
```diff
@@ -1,5 +1,12 @@
 class AddMetadataToSpotlightResource < ActiveRecord::Migration
   def up
-    add_column :spotlight_resources, :metadata, :blob
+    # postgresql does not have a blob type
+    conn_type = connection.adapter_name.downcase.to_sym
+    field_type = (conn_type == :postgresql) ? :bytea : :blob
+    add_column :spotlight_resources, :metadata, field_type
```
jkeck: Instead of having adapter specific type definitions, I wonder if we could simply use `:binary` in all cases. I believe this is the ActiveRecord native type that is intended to map to `:bytea` for postgres and `:blob` for MySQL/sqlite.

ronan-mch: That [seems right](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/TableDefinition.html#method-i-column). I can give it another go tomorrow and see if it works.

PR #1328, opened 2015-12-10
Remove database-level locking when reindexing resources

cbeer:
This doesn't seem to scale well for very large collections, and with all
indexing happening through ActiveJob queuing also seems unnecessary.


cbeer:
Lost coverage is just from removing covered lines of code.

PR #1329, opened 2015-12-10
Page exports should use the original attribute value, before SirTrevo…

cbeer:
…rRails gets a chance to munge it


ndushay:
:+1:  once this is rebased.

PR #1331, opened 2015-12-11
Update invitation instructions email

jkeck:


ggeisler:
Sorry, @jkeck you are working faster than I can provide comments. Not a huge deal, but I think we could incorporate some of my suggestions from https://github.com/sul-dlss/sul_exhibits_template/pull/144 here, beyond the typo:

>
Subject: Exhibit invitation instructions
>
The Exhibits Administrator has invited you to help work on the "Invite Test" exhibit at http://exhibits-stage.stanford.edu/default-exhibit/home. You can accept this invitation by clicking the link below.

The rest is fine for the generic version, although I think you can remove the break tag and make the last two sentences a single paragraph.

Sorry I missed commenting on this before, it must've been merged before I had a chance to look at it.

cbeer:
@ggeisler I think you also suggested using double quotes?

ggeisler:
@cbeer Thanks, yes, I'd also suggest the double-quotes here.

jkeck:
Okay, I think we're ready to :shipit:. Let me know if there are any other changes we need to make.

ggeisler:
:+1:

PR #1341, opened 2015-12-12
Update gradient direction

jcoyne:
Follow the 2012 (latest) spec: http://www.w3.org/TR/css3-images/#linear-gradients
This prevents autoprefixer from issuing warnings.
Fixes sul-dlss/spotlight#1340


cbeer:
:+1:

PR #1343, opened 2015-12-14
Add documentation

jcoyne:


lib/spotlight/catalog/access_controls_enforcement.rb
```diff
@@ -17,7 +17,8 @@ module SearchBuilder
         included do
           self.default_processor_chain += [:apply_permissive_visibility_filter, :apply_exhibit_resources_filter]
         end
-
+
```
cbeer: Could you address this?

PR #1345, opened 2015-12-14
extract exhibit functionality into module fixes #1344

### LINKED ISSUE #1344: What is the best way to override Exhibit#solr_data?
we're currently defining our own `app/models/spotlight/exhibit.rb` and requiring `Spotlight::Engine.root + 'app/models/spotlight/exhibit'` then overriding. This is not autoload safe (we have to restart our server after each change). Is there a better way?

ronan-mch:


cbeer:
How does this fix #1345?

ronan-mch:
Hi, see our downstream app for an [implementation of an overridehttps://github.com/dtulibrary/ddf/blob/spotlight/app/models/spotlight/exhibit.rb). Otherwise we would have to monkey patch the class as described by @jcoyne in #1344

cbeer:
If that's all you're trying to do to, you can also set:

`Spotlight::Engine.config.filter_resources_by_exhibit = false`

in an initializer.

https://github.com/sul-dlss/spotlight/blob/master/lib/spotlight/engine.rb#L75

jcoyne:
We're just doing no filtering right now. Eventually we want to filter the exhibit by a custom ORCID field

jcoyne:
@cbeer if you don't like this approach, we could make `solr_data` run a Proc that is set in the Spotlight configuration.  Would you like that better?

cbeer:
I think that sounds better, yeah. It makes me nervous to move some of the things into that module, as some of it isn't portable to any arbitrary model, and hopefully a `Proc` makes our intent more obvious too.

PR #1346, opened 2015-12-14
use apps local SolrDocument to enable overriding

ronan-mch:


app/models/spotlight/blacklight_configuration.rb
```diff
@@ -252,11 +252,11 @@ def add_exhibit_specific_fields(config)
     def add_exhibit_tags_fields(config)
       # rubocop:disable Style/GuardClause
       unless config.show_fields.include? :exhibit_tags
-        config.add_show_field :exhibit_tags, field: Spotlight::SolrDocument.solr_field_for_tagger(exhibit), link_to_search: true
+        config.add_show_field :exhibit_tags, field: ::SolrDocument.solr_field_for_tagger(exhibit), link_to_search: true
```
cbeer: Instead of using `::SolrDocument` directly, we'll need to use `blacklight_config.document_model`, as that's a configurable class too.

jcoyne:
@cbeer better?

app/models/spotlight/custom_field.rb
```diff
@@ -50,12 +50,12 @@ def configured_to_display?
         end
     end
 
-    def solr_field
-      if field && field.starts_with?(solr_field_prefix)
+    def solr_field(document_model)
```
cbeer: Was there a problem pulling this out of the exhibit's `blacklight_configuration`?

cbeer:
Looks great, just one last comment.

app/models/spotlight/custom_field.rb
jcoyne: Yes, that causes an infinite loop.

cbeer: :ok_hand: I think we've run into that before too. I'll check if there's anything else we can do.

cbeer:
Rebased in #1356.

PR #1347, opened 2015-12-14
move before filter into method to enable overriding

ronan-mch:


lib/spotlight/catalog.rb
```diff
@@ -11,17 +11,18 @@ module Catalog
...
-                                              private: {
-                                                label: I18n.t(:'spotlight.catalog.facets.exhibit_visibility.private'),
-                                                fq: "#{Spotlight::SolrDocument.visibility_field(current_exhibit)}:false" }
-                                            }
-        end
-      end
+      before_action :facet_visibility
+    end
+
+    def facet_visibility
```
cbeer: Could we use a more descriptive name for this filter, maybe: `add_facet_visibility_field`, and add a little method documentation?

PR #1352, opened 2015-12-15
use exhibition config rather than default config in repository - fixes #1351

### LINKED ISSUE #1351: Search model should use a configured repository
https://github.com/sul-dlss/spotlight/blob/761c0b123cee580f3414afa5d8847dc6a54df36b/app/models/spotlight/search.rb#L85

Presently it's ignoring our configured `solr_path`

Because it creates a new configuration instance.
https://github.com/projectblacklight/blacklight/blob/release-5.x/lib/blacklight.rb#L66

ronan-mch:


app/models/spotlight/search.rb
```diff
@@ -82,7 +82,7 @@ def search_builder
     end
 
     def repository
-      Blacklight.default_index
+      @repository ||= Blacklight::Solr::Repository.new(blacklight_config)
```
cbeer: I don't think we necessarily want to hard-code `Blacklight::Solr` here, although `Blacklight.repository_class` seems to be Blacklight 6-only. @jcoyne, does it make sense to add that method to Blacklight 5.x with a hard-coded default, or just remember to revisit this when go to Blacklight 6?

jcoyne: Yes, we discussed this. My thought was to refactor when BL 6 is out.

cbeer: :+1:

PR #1354, opened 2015-12-15
Make the exhibit filter configurable

### LINKED ISSUE #1344: What is the best way to override Exhibit#solr_data?

jcoyne:
This enables us to use a different field/value to select which documents belong to our exhibit.

This is the alternate PR for #1345 Fixes #1344

### DIFF
--- a/app/models/spotlight/exhibit.rb
+++ b/app/models/spotlight/exhibit.rb
@@ -74,7 +74,7 @@ module Spotlight
     end
 
     def solr_data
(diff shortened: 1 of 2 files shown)


cbeer:
Rebased in #1355.

PR #1359, opened 2015-12-15
Allow site administrators to reorder exhibits

### LINKED ISSUE #1326: Enable super-admin to reorder exhibits and view status details
One of the new sidebar links to be added in #1323 is "Manage exhibits." This is only available to users with the super-admin role. The link should go to a new page that enables the super-admin to reorder exhibits and view some additional details about them.

Page header: "Manage exhibits"
Secondary header: "Existing exhibits"

The idea is to use a table where the rows can be reordered, similar to our current Metadata management page.

- Instructions text: "Drag and drop the exhibits below to specify the order in which they are displayed on the exhibits homepage."
- Exhibit titles link to root of the exhibit
- Unlike the metadata page, title is not an in-place editable field
- The other table columns should be right-aligned for easy scanning

landing-page-manage-exhibits png

cbeer:
Fixes #1326 

screen shot 2015-12-15 at 1 50 40 pm


cbeer:
Hopefully the differences between this and @ggeisler are sufficiently minor, or are the kind of thing we should address locally.

app/assets/stylesheets/spotlight/_blacklight_configuration.scss
```diff
@@ -49,15 +46,17 @@
...
   th {
     &:first-child {
       padding-left: 40px;
-      text-align: left;
     }
-    text-align:center;
   }
+}
+
+#nested-fields {
```
cbeer: Not a new selector.

ggeisler:
This is great. Below are a couple of very minor styling issues that I think would improve the appearance a bit. You could argue these should be addressed locally, though I don't think they would be bad for anyone out of the box.

If you think we should deal with them locally, then this looks good to merge to me.

- Make the table headings `font-weight: 500` so they aren't so dominant, since they're bold by default (or reduce them to 14px from 16px, but I don't think we have a built-in variable for 14px font size)

- Add 20px top margin to the table to add some whitespace between it and the instructions text.

Example:

### Before
manage_exhibits_-_existing_exhibits___exhibits_home

--
### After
after

cbeer:
Sure thing.

PR #1360, opened 2015-12-16
Add panel to monitor reindex status.

### LINKED ISSUE #1220: Improved background job processing notifications
Now that reindexing is able to use background jobs for all indexing, we need better in-application reporting for status, progress, and errors.

jkeck:
Also add an index_status enum column to the Spotlight::Resource to track the index state.
Closes #1220 

reindex-status-panel

### DIFF
--- a/app/assets/javascripts/spotlight/reindex_monitor.js
+++ b/app/assets/javascripts/spotlight/reindex_monitor.js
@@ -1,0 +1,109 @@
+Spotlight.onLoad(function() {
+  $('[data-behavior="reindex-monitor"]').reindexMonitor();
+});
+
+(function($) {
+  $.fn.reindexMonitor = function() {
+    var monitorElements = this;
+    var defaultRefreshRate = 3000;
+    var panelContainer;
+
+    $(monitorElements).each(function() {
+      panelContainer = $(this);
+      var monitorUrl = panelContainer.data('monitorUrl');
+      var refreshRate = panelContainer.data('refreshRate') || defaultRefreshRate;
+      setInterval(function() {
+        checkMonitorUrl(monitorUrl);
+      }, refreshRate);
+    });
+
+    function checkMonitorUrl(url) {
+      $.ajax(url).success(success).fail(fail);
+    }
+
+    function success(data) {
+      if (data.in_progress) {
+        monitorPanel().show();
+        updateMonitorPanel(data);
+      } else {
+        monitorPanel().hide();
+      }
+    }
+
+    function fail() { monitorPanel().hide(); }
+
+    function updateMonitorPanel(data) {
+      panelStartDate().text(data.started);
+      panelCurrentDate().text(data.updated_at);
+      panelCompleted().text(data.completed);
+      updatePanelTotals(data);
+      updatePanelErrorMessage(data);
+      updateProgressBar(data);
+    }
+
+    function updateProgressBar(data) {
+      var percentage = calculatePercentage(data);
+      progressBar()
+        .attr('aria-valuemax', data.total)
+        .attr('aria-valuenow', percentage)
+        .css('width', percentage + '%')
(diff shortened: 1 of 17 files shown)


app/models/spotlight/exhibit.rb
```diff
@@ -95,6 +95,11 @@ def set_default_thumbnail
       self.thumbnail ||= searches.first.try(:thumbnail)
     end
 
+    def reindex_progress
+      return unless resources
+      @reindex_progress ||= ReindexProgress.new(resources.order('updated_at'))
```
jkeck: Not sure if there is a more rails-y wait to do this or if we should add this as an explicit scope.

app/models/spotlight/reindex_progress.rb
```diff
@@ -0,0 +1,101 @@
...
+      @resources = if resource_list.present?
+                     resource_list
+                   else
+                     null_resources
+                   end
+    end
+
+    def in_progress?
+      return unless finished
+      finished > MINUTES_IN_PROGRESS.minutes.ago
```
jkeck: This basically will give us a window of 10 minutes after the last item was finished indexing where the panel will be present.  Not sure if we should use a different amount of time or a different algorithm all together.

cbeer: Maybe make this configurable at the engine level?

ggeisler:
The panel looks good. Maybe we can look at the interaction together tomorrow. I'm sure it's fine, but what I'm seeing (I think) is that when I reindex, the page doesn't refresh until all the indexing is done, so the status panel is always showing 100% (even with 60 new items added). But this could be a local enivornment issue or something I'm not understanding so if it works for @cbeer then don't wait on me to merge.

jkeck:
@ggeisler since you presumably do not have delayed job setup on your machine all the indexing is done in one request (and that all happens before your browser is redirected to the items page).

We'll have to run this on a server w/ DelayedJob setup as well as have enough documents that the entire reindexing job isn't done in a few seconds.

app/models/spotlight/reindex_progress.rb
jkeck: Yeah, I can do that.  Let me make that change. 

Also, I wonder since we have the `index_status` now if we can be more clever here.  Like possibly `resources.any?(&:waiting?) && finished > MINUTES_IN_PROGRESS.minutes.ago`.

Otherwise I think this will require one resource to have completed before the reindexing is reported as in progress.

ggeisler:
@jkeck That explains it, thanks!

app/assets/javascripts/spotlight/reindex_monitor.js
```diff
@@ -0,0 +1,109 @@
...
+      } else {
+        monitorPanel().hide();
+      }
+    }
+
+    function fail() { monitorPanel().hide(); }
+
+    function updateMonitorPanel(data) {
+      panelStartDate().text(data.started);
+      panelCurrentDate().text(data.updated_at);
```
jkeck: This is coming from a ruby method and I tend to prefer to keep to ruby's style, but could be convinced to change the keys in the JSON we return (or perhaps relax the jshint rule).

app/assets/javascripts/spotlight/reindex_monitor.js
```diff
@@ -0,0 +1,109 @@
...
+        checkMonitorUrl(monitorUrl);
+      }, refreshRate);
+    });
+
+    function checkMonitorUrl(url) {
+      $.ajax(url).success(success).fail(fail);
+    }
+
+    function success(data) {
+      if (data.in_progress) {
```
jkeck: This is coming from a ruby method and I tend to prefer to keep to ruby's style, but could be convinced to change the keys in the JSON we return (or perhaps relax the jshint rule).

PR #1362, opened 2015-12-16
Extract ExhibitsController behaviors into concerns

jcoyne:
This enables our local application to customize the behavior of
Spotlight::ExhibitsController.

Our plan is to save an associated user editable object (ExhibitFilter)
in the create and update actions.


jcoyne:
I'd be happy to close this is #1385 and #1387 are merged.

Issue #849, opened 2015-01-12
Allow additional super-admins to be created via the GUI


hortongn:
The current GUI has no way for the administrator to create additional admins.  Only admins of exhibits can currently be created.

cbeer:
With the new exhibits index page, it seems particularly important to have a way to manage superadmin privileges.

ggeisler:
One of the new sidebar links to be added in #1323 is "Manage users." This link should go to a new page that enables the super-admin to:

- Remove existing super-admins from the super-admin role
- Add a new user with the super-admin role

Page header: "Manage users"
Secondary header: "Exhibits administrators"
Table caption: "Existing exhibits administrators"
Button text: "Remove from role" (subject to change if we think of something better...)

I'm not including all details here as I expect it will work very similarly to the newly updated feature for managing users within an exhibit. We can discuss details as they come up, but this would be the basic page details:

![landing-page-manage-users](https://cloud.githubusercontent.com/assets/101482/11703793/d50f5090-9e96-11e5-82a4-35e2118fc062.png)

cbeer:
Might require #713

ggeisler:
@jkeck Below is some suggested text for the mailer. I put the SUL version in https://github.com/sul-dlss/exhibits/issues/165

```` yml
exhibits_admin_invitation_mailer:
  invitation_instructions:
    hello: "Hello %{email}!"
    someone_invited_you: "The Exhibits Administrator has invited you to help manage exhibits. You can accept this invitation by clicking the link below."
    accept: "Accept invitation"
    ignore_html: "If you don't want to accept the invitation, please ignore this email. Your exhibits administrator account won't be created until you access the link above."

PR #1365, opened 2015-12-16
Update roles to use polymorphic associations with resources

cbeer:
This is a prerequisite for #849


jkeck:
:+1: LGTM, needs a rebase though.

PR #1367, opened 2015-12-16
Use bootstrap nav styling for the sidenav

cbeer:
screen shot 2015-12-16 at 8 49 58 am
screen shot 2015-12-16 at 8 50 01 am
screen shot 2015-12-16 at 8 54 32 am

### DIFF
--- a/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
@@ -21,18 +21,26 @@
 }
 
 .sidenav {
-
   margin-bottom: 24px;
 
-  li > a {
-    padding: 0 15px 0 0;
-  }
+  .active {
+    a {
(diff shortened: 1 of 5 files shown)


cbeer:
The default padding seems a little extreme, but there's a variable we can tweak locally if we want to.

ggeisler:
Well, this is definitely a case where I don't see how any implementor would want to use that styling, but if you think it's better adjusted in the implementing app, that's fine.

cbeer:
Apparently someone with bootstrap thinks the style is great, but they've helpfully left a variable we can change downstream:

https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L427

ggeisler:
Yep. And maybe this is just another instance of a style that works better on other types of sites, and it's just in our context where it seems like an unusual amount of vertical padding...

PR #1368, opened 2015-12-16
Use naive solr facet query to get facet values and term counts

### LINKED ISSUE #1333: Facet field information on the Configuration > Search page doesn't respect filter_resources_by_exhibit

cbeer:
Fixes #1333

With a multitenant index, the luke request handler data cannot be scoped to the
current exhibit. Instead, use the search builder and ordinary facet queries to
get the appropriate information (with a slight loss of information -- we can't
get the number of terms in an exhibit, so just give '20+' for large counts)

screen shot 2015-12-16 at 2 08 45 pm


ndushay:
should there be a test for the +20  case?

cbeer:
Should there be? Yeah, probably.

cbeer:
I broke out the facet metadata into its own partial and added a view spec.

PR #1370, opened 2015-12-16
Remove indexes before removing columns, or we end up with the wrong unique index

cbeer:


cbeer:
Interesting story.. when you remove columns that are in a unique index, mysql(?) helpfully removes them from the unique index too. so, say you have an index on `[exhibit_id, user_id]`, if you remove the `exhibit_id` column, you're left with a unique index on `user_id`.

PR #1381, opened 2015-12-17
Add validation to ensure slugs are unique

cbeer:


app/models/spotlight/exhibit.rb
```diff
@@ -56,8 +57,6 @@ class Exhibit < ActiveRecord::Base
     after_create :initialize_main_navigation
     include Spotlight::DefaultThumbnailable
 
-    scope :published, -> { where(published: true) }
```
jcoyne: Why is this change in this PR?

cbeer: This line is duplicated on line 9. It probably should be in a separate commit, though.

PR #1384, opened 2015-12-17
Update browse block to use Search#documents enumerable to get the num…

cbeer:
…ber of documents


ndushay:
looks like we need a test for this?

cbeer:
Pretty sure we have a test, but it wasn't in a multitenant environment. `Search#documents` is appropriately tested for such an environment.

PR #1386, opened 2015-12-17
Pin sprockets-rails to 2.x versions

cbeer:


cbeer:
Or this can wait for an upstream fix in https://github.com/projectblacklight/blacklight/pull/1350.

PR #1387, opened 2015-12-17
Allow solr filters to be configurable on a per-exhibit basis

jcoyne:
This still needs a controller test, but I wanted your feedback before you head out.

### DIFF
--- a/app/controllers/spotlight/exhibit_filters_controller.rb
+++ b/app/controllers/spotlight/exhibit_filters_controller.rb
@@ -1,0 +1,28 @@
+module Spotlight
+  # Create and update filters for an exhibit
+  class ExhibitFiltersController < ApplicationController
+    before_action :authenticate_user!
+    load_and_authorize_resource :exhibit, class: 'Spotlight::Exhibit'
+    load_and_authorize_resource through: :exhibit
+
+    def create
+      if @exhibit_filter.save
+        flash[:notice] = t('helpers.submit.exhibit_filter.updated', model: @exhibit_filter.model_name.human)
+      else
+        flash[:alert] = @exhibit_filter.errors.full_messages.join('<br/>'.html_safe)
+      end
+      redirect_to spotlight.edit_exhibit_path @exhibit, anchor: 'filter'
+    end
+
+    def update
+      unless @exhibit_filter.update(exhibit_filter_params)
+        flash[:alert] = @exhibit_filter.errors.full_messages.join('<br/>'.html_safe)
+      end
+      redirect_to spotlight.edit_exhibit_path @exhibit, anchor: 'filter'
+    end
+
+    def exhibit_filter_params
+      params.require(:exhibit_filter).permit(:exhibit_id, :field, :value)
+    end
+  end
+end
--- a/app/views/exhibit_filters/_edit.html.erb
+++ b/app/views/exhibit_filters/_edit.html.erb
@@ -1,0 +1,9 @@
+<div role="tabpanel" class="tab-pane" id="filter">
+
+<%= bootstrap_form_for [@exhibit, @exhibit.exhibit_filters.first] do |f| %>
+  <%= f.text_field :field %>
+  <%= f.text_field :value %>
+  <%= f.submit %>
+<% end %>
+</div>
+
--- a/app/views/spotlight/exhibits/edit.html.erb
+++ b/app/views/spotlight/exhibits/edit.html.erb
@@ -6,6 +6,11 @@
       <li role="presentation" class="active">
         <a href="#basic" aria-controls="basic" role="tab" data-toggle="tab"><%= t(:'.basic_settings.heading') %></a>
       </li>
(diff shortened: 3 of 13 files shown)


lib/spotlight/engine.rb
```diff
@@ -87,7 +87,12 @@ def self.blacklight_config
 
     # A lambda expression that filters the solr index per exhibit
     config.exhibit_filter = lambda do |exhibit|
-      { :"#{config.solr_fields.prefix}spotlight_exhibit_slug_#{exhibit.slug}#{config.solr_fields.boolean_suffix}" => true }
+      filter = exhibit.exhibit_filters.first
+      if filter
+        { filter.field => RSolr.solr_escape(filter.value) }
```
cbeer: Can we push this logic up into `ExhibitFilter`?

db/migrate/20151217211019_create_spotlight_exhibit_filters.rb
```diff
@@ -0,0 +1,11 @@
+class CreateSpotlightExhibitFilters < ActiveRecord::Migration
+  def change
+    create_table :spotlight_exhibit_filters do |t|
+      t.string :field
+      t.string :value
+      t.references :spotlight_exhibit, index: true, foreign_key: true
```
cbeer: I think if this was just `references :exhibit`, everything would work out automagically?

app/views/spotlight/exhibits/edit.html.erb
```diff
@@ -6,6 +6,9 @@
       <li role="presentation" class="active">
         <a href="#basic" aria-controls="basic" role="tab" data-toggle="tab"><%= t(:'.basic_settings.heading') %></a>
       </li>
+      <li role="presentation">
```
cbeer: Could we add + check an ability? `can :manage, current_exhibit.filters.build` maybe?

app/controllers/spotlight/exhibit_filters_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Spotlight
+  # Create and update filters for an exhibit
+  class ExhibitFiltersController < ApplicationController
+    before_action :authenticate_user!
+    load_and_authorize_resource :exhibit, class: 'Spotlight::Exhibit'
+    load_and_authorize_resource through: :exhibit
+
+    def create
+      if @exhibit_filter.save
+        flash[:notice] = 'Filter added'
```
cbeer: I think we have some i18n magic that maybe do this for us?

.rubocop.yml
```diff
@@ -41,5 +41,6 @@ Metrics/ModuleLength:
 
 Metrics/ClassLength:
   Exclude:
+    - 'app/models/spotlight/exhibit.rb'
```
cbeer: I'll try to figure out another logical concern to pull out of `Exhibit`..

app/views/exhibit_filters/_edit.html.erb
```diff
@@ -0,0 +1,9 @@
+<div role="tabpanel" class="tab-pane" id="filter">
+
+<%= bootstrap_form_for [@exhibit, @exhibit.exhibit_filters.first_or_initialize] do |f| %>
```
cbeer: I've asked @ggeisler to give some thought to the UI for this behavior. I don't think it should necessarily hold up the PR, though.

app/views/exhibit_filters/_edit.html.erb
```diff
@@ -0,0 +1,9 @@
...
+  <%= f.text_field :field %>
+  <%= f.text_field :value %>
+  <%= f.submit %>
```
cbeer: I think we have some wrapping `div`s for this for other forms?

cbeer:
:+1: Makes sense to me, couple of minor comments, but the idea looks promising.

db/migrate/20151217211019_create_spotlight_exhibit_filters.rb
jcoyne: Yeah, I was chasing a bug where Postgres wansn't generating the correct foreign_key constraint, but that may only exist when both models are not in the same namespace.  I'll try it the simplest way again.

app/views/exhibit_filters/_edit.html.erb
jcoyne: It seems like bootstrap_form_for writes out all the correct divs.

spec/models/spotlight/exhibit_spec.rb
```diff
@@ -179,6 +191,9 @@
     before do
       allow(subject).to receive(:blacklight_config).and_return(blacklight_config)
       allow(subject).to receive(:slug).and_return(slug)
+      allow(subject).to receive(:exhibit_filters)
+        .and_return([Spotlight::ExhibitFilter.new(field: subject.send(:default_filter_field),
```
jcoyne: Any idea why rubocop isn't catching this style violation?

PR #1397, opened 2015-12-18
Use a fallback image on the exhibits index when no exhibit thumbnail exists

### LINKED ISSUE #1389: Add fallback image for case where exhibit doesn't have thumbnail image defined
On the exhibits home page, the exhibit title and the "Unpublished" label elements look out of place when an exhibit has no thumbnail image. To ensure the image element is there (which ensures the title and label elements will be positioned correctly) use a fallback image when no exhibit thumbnail exists.

### No image exists
blacklight 2

--
### Fallback image used
blacklight

### Fallback image file

exhibit-thumbnail-fallback

cbeer:
Fixes #1389

screen shot 2015-12-18 at 8 44 02 am


app/views/spotlight/exhibits/_exhibit_card.html.erb
```diff
@@ -1,7 +1,11 @@
 <div class="col-xs-12 col-sm-4 exhibit-card">
   <div class="flipper">
     <div class="card-front card-face">
-      <%= image_tag(exhibit.thumbnail.image.square.url) if exhibit.thumbnail.present? %>
+      <% if exhibit.thumbnail.present? %>
+        <%= image_tag(exhibit.thumbnail.image.square.url) %>
+      <% else %>
+        <%= image_tag 'spotlight/default_thumbnail.jpg' %>
```
jkeck: Would it be helpful to make this configurable so that people can use their own image, or is it possible to just override the image itself? (I'm not 100% sure how that all works w/ engines and the asset pipeline)

jkeck: I wonder if it's worth having a test for the default thumbnail rendering.

cbeer: Overriding the image downstream should be fine, but I haven't tried it. I'll try to pull the front and back of the card out into separate partials and write a simple view spec for it.

cbeer: #1399

PR #1398, opened 2015-12-18
Allow solr filters to be configurable on a per-exhibit basis

cbeer:
Fixes #1387


ndushay:
this looks fine to me, but I don't understand it's relationship to #1387 -- is it instead of?  in addition to?   So :+1:  for this but not entirely clear if i should merge or wait for #1387 or ...

PR #1404, opened 2016-01-05
Allow site admins to manage other site admins

### LINKED ISSUE #849: Allow additional super-admins to be created via the GUI

jkeck:
Closes #849 

admin-user-admin

### DIFF
--- a/app/assets/javascripts/spotlight/check_user_existence.js
+++ b/app/assets/javascripts/spotlight/check_user_existence.js
@@ -10,8 +10,7 @@
 
     function checkIfUserExists() {
       target = $(this);
-      var form = target.closest('form');
-      if (target.val() !== '' && form[0].checkValidity()) {
+      if (target.val() !== '' && form()[0].checkValidity()) {
         $.ajax(userExistsUrl())
          .success(userExists)
          .fail(userDoesNotExist);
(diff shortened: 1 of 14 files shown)


jkeck:
There is a bit of a spacing issue in the form where the label has `.sr-only` as well as a `.col-md-2` but the `.col-*` declaration is keeping the allotted space available.  I don't see why the `.col-*` class isn't being added other places where we're using `hide_label: true`.

If anybody has any ideas as to why this is happening or proposed solutions, I'm all ears.

jkeck:
Odd about the reduced coverage.  I think it's caused by relevant lines of code being moved, but that code is tested elsewhere (and showing 100% coverage). It seems odd to me that this would cause a drop of 0.2%.

cbeer:
Agreed about the coverage; the coveralls report is less than helpful.

Should we wait to merge this until we can look at the label spacing problem?

jkeck:
Yeah, maybe I can show you what is going on.  It's pretty minor so we can fix it up and :shipit: later today or tomorrow.

cbeer:
:ok_hand:

cbeer:
I've fixed the label padding problem  -- it looks like the horizontal form layout (which we didn't actually seem to need) forced columns on us. 

screen shot 2016-01-07 at 8 09 24 am

ggeisler:
:+1:

jkeck:
Ah!  Makes perfect sense.  Thanks @cbeer!  I'll rebase this and re-push as well.

cbeer:
I'm not sure about perfect sense, but it probably helps keeping things aligned (and if we weren't using tables, we might need it..)

PR #1411, opened 2016-01-07
avoid error when exhibit is nil

ndushay:
sul-dlss/spotlight-dor-resources#53 is blocked (see https://travis-ci.org/sul-dlss/spotlight-dor-resources/builds/100713436)

because spotlight-dor-resources/spec/models/spotlight/resources/purl_spec.rb#L129  had this error:

"NoMethodError:
       undefined method `resource_type_field' for nil:NilClass
     # /Users/ndushay/.rvm/gems/ruby-2.2.3/gems/blacklight-spotlight-0.15.0/app/models/spotlight/resource.rb:109:in `spotlight_resource_metadata_for_solr'
..."

I *think* spotlight-dor-resources is trying to allow reindexing to occur outside of the Rails app (i.e. from the command line) and this error stems from that use case.

### DIFF
--- a/app/models/spotlight/resource.rb
+++ b/app/models/spotlight/resource.rb
@@ -104,10 +104,9 @@ module Spotlight
       end
 
       def spotlight_resource_metadata_for_solr
-        {
-          Spotlight::Engine.config.resource_global_id_field => (to_global_id.to_s if persisted?),
-          document_model.resource_type_field => self.class.to_s.tableize
-        }
+        hash = { Spotlight::Engine.config.resource_global_id_field => (to_global_id.to_s if persisted?) }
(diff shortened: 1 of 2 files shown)


cbeer:
Can we confirm why `spotlight-dor-resources` needs this behavior? I wonder, now that we have shareable factories, if we can easily require an exhibit?

ndushay:
@peetucket  are you okay with removing the ability to index with a nil exhibit from spotlight-dor-resources?

ndushay:
Spoke to @peetucket.  He is okay with removing the failing test from spotlight-dor-resources, and in a separate PR, updating the documentation on how to index from the command line with the latest spotlight gem.

So closing this PR without merging.

PR #1412, opened 2016-01-08
filtering exhibits by tag

### LINKED ISSUE #1400: Add UI elements for filtering exhibits by tag
@jcoyne @cbeer @jkeck This is to take advantage of the exhibit tags added in #1385. This is a pretty quick idea so we might need some additional discussion and design iteration but I wanted to get something out there before the winter break.

### Filter by exhibit tag

The idea is that the "All topics" link is the default, and any other existing exhibit tags are then shown as additional links to filter by that tag. Examples:

--
all

--
arts

--
science

cbeer:
Fixes #1400 

screen shot 2016-01-08 at 8 10 27 am

### DIFF
--- a/app/controllers/spotlight/exhibits_controller.rb
+++ b/app/controllers/spotlight/exhibits_controller.rb
@@ -8,6 +8,9 @@ module Spotlight
     load_and_authorize_resource
 
     def index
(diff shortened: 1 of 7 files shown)


ggeisler:
:+1: Works well. I'll let @jkeck merge in case he wants to review the code.

PR #1414, opened 2016-01-08
Use i18n for the 'all tags' link, and sort tags by name

cbeer:


ggeisler:
I like the sorting by name (I thought about that after the earlier pull request and was going double-check we were doing that).

I'm not sure about the "All tags" label, though. I think we should stick to "All." I'm not sure we need or want to expose the word "tags" to end-users in this context, and although "All" is kind of short, I think the options read fine when it is "All" "Science & Technology" "Art" etc.

Basically, tagging is the mechanism we use to categorize exhibits, but the end-user doesn't need to know that. They just need to see the categories they can choose from.

cbeer:
:ok_hand:

PR #1421, opened 2016-01-14
Move Date sort to the bottom of the list in the test-app

### LINKED ISSUE #1406: Update sort order for fields on Curation > Items page
As some of us discussed in person, it probably would be more useful if we changed the sort order for a couple of the fields available for sorting on the Curation > Items page. I think it's just the two date-related fields that need updating:

- Date (old to new) 
- Timestamp

For each of these, let's go with newest first. This also means changing the label for the Date field to `Date (new to old)`. I don't think we need to worry about the label for Timestamp.

jkeck:
Closes #1406 

This is purely cosmetic since it is only in the `CatalogController` generated into the internal test app.

### Items Admin Page
admin

### Search Results Admin Page
search-results

### DIFF
--- a/spec/features/edit_search_fields_spec.rb
+++ b/spec/features/edit_search_fields_spec.rb
@@ -54,7 +54,7 @@ describe 'Search Administration', type: :feature do
 
         # #field_labeled doesn't appear to work for disabled inputs
         expect(page).to have_css("input[name='blacklight_configuration[sort_fields][relevance][enable]'][disabled='disabled']")
-        expect(page).to have_css('#nested-sort-fields .dd-item:nth-child(5) h3', text: 'Identifier')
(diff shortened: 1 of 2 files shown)


ggeisler:
Wait, I think the main goal of this issue might have been lost in our discussion of "it's only in the test app" etc.

The main goal was to switch `Date` and `Timestamp` so they sort from new to old instead of old to new. It looks like you moved `Date` down but it is still sorting the same way?

ggeisler:
Looking back at the ticket, I think you were reading it as "field order" when I wrote "change the sort order" in the ticket... sorry for not realizing you were interpreting it that way when we were discussing the issue.

jkeck:
:+1: my bad, miss-read that part.  Just pushed up the change to reverse the date sort so it is new to old.

Issue #1401, opened 2015-12-18
Use bootstrap tags control with controlled vocabularies for the exhibit tags


cbeer:

cbeer:
@ggeisler also indicated some preference, maybe, for something like: https://github.com/davidstutz/bootstrap-multiselect

ggeisler:
That looks promising. I'll try to come up with a design proposal soon.

ggeisler:
Below are mockups for creating and assigning exhibit tags using a controlled vocabulary approach. This approach is more involved but would be preferable to #1413.

Idea is to create a "Tag exhibits" page, and on that page use a tabbed UI to separate managing tags and assigning tags.

### New menu item to link to new "Tag exhibits" page

Add "Tag exhibits" link to exhibits sidebar (if role is super-admin)

![add menu item](https://cloud.githubusercontent.com/assets/101482/12273504/fa34b75e-b91a-11e5-8c26-26ff30e4fefb.png)
--

### Manage tags

On the new page, we have two tabbed content areas. For Manage tags, we show existing tags in a table and enable adding, deleting, and renaming tags. If feasible, we could also add a checkbox ("Also allow free text tagging") that would control whether we show the "Tags" input (that currently allows tags separated by a comma) on the new exhibit and exhibit Config > General > Basic settings forms. The default for the checkbox would be off (because the preferable use case is for exhibit tags to be assigned using a controlled vocabulary).

Mockup doesn't show it correctly, but sort tags alphabetically in the table.

![manage-tags](https://cloud.githubusercontent.com/assets/101482/12273753/701dd77e-b91c-11e5-9cc5-c6f664b69f68.png)
--

When the user selects the Add new tag button, we'd show an input for the tag in a new row at the top or bottom of the table (mockup shows bottom, but I now think top might be better to keep the new element close to the action that triggered its appearance).

![manage-tags-new](https://cloud.githubusercontent.com/assets/101482/12273845/ecf4bdd0-b91c-11e5-957a-eff3187407ed.png)
--

To enable renaming, we could make the tag label edit-in-place, similar to how we do it elsewhere in Spotlight. 

Deleting would show a confirmation dialog: "Are you sure you want to delete this tag? If deleted, the tag will be removed from all exhibits to which it has been assigned."

![manage-tags-edit](https://cloud.githubusercontent.com/assets/101482/12273935/50d89ee8-b91d-11e5-8365-68d156c8e416.png)
--

[...]

PR #1422, opened 2016-01-15
Create and assign exhibit tags from a controlled vocabulary

cbeer:


jcoyne:
Related to #1401?

cbeer:
Before merging, this could use some more testing. I think @jkeck discovered some problem with it (which is why it's languished..), but I don't remember what it was.

hackartisan:
We looked at this in ensemble PR review. Our memory is that this work was done via another PR, and released as part of spotlight 5.0.

PR #1425, opened 2016-01-19
Update Spotlight to support and require Blacklight 6.0

cbeer:


cbeer:
I believe the loss of coverage is simply from removing covered lines.

cbeer:
Replaced by #1449

PR #1426, opened 2016-01-19
Exhibits admin pages updates

ggeisler:
Includes a handful of updates, all related to tidying up the exhibit admin pages:

* Several adjustments for better alignment of page elements
* I updated the page title and subtitles scheme, so that all page titles are "Manage exhibits" and the subtitles are the same as the sidebar links. This is more consistent with how we do things in the exhibit admin pages, and generally feels more logical than the previous scheme.


cbeer:
:+1:

PR #1428, opened 2016-01-26
Fix foreign key reference

ronan-mch:


ronan-mch:
Background - I made this PR yesterday when I wasn't able to run the relevant migration since Postgres couldn't identify the table being referenced. Continuing with it today, I have found that exhibit creation doesn't work since the exhibit filter won't respond to an `exhibit_id=` assignment. So, I apologise, the PR shouldn't be merged. I'm leaving it open though in the hope that someone can shed some light on whether the error is something wrong with my environment or whether the ExhibitFilter feature is broken. Pinging @jcoyne.

cbeer:
Fixed by #1431.

jcoyne:
Has anyone tried this?

```
add_foreign_key :spotlight_filters, :exhibit, column: :spotlight_exhibits_id
``

PR #1448, opened 2016-04-01
ReadOnly has a different field name.

tpendragon:
Allows imported fields to have the same label as a user-defined field,
but have its own content.

### DIFF
--- a/app/models/spotlight/custom_field.rb
+++ b/app/models/spotlight/custom_field.rb
@@ -17,7 +17,7 @@ module Spotlight
     end
 
     before_save do
-      update_field_name(field_name) if persisted? && field_type_changed?
+      update_field_name(field_name) if update_field_name?
     end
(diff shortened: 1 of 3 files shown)


app/models/spotlight/custom_field.rb
```diff
@@ -78,7 +78,15 @@ def field_name
     end
 
     def field_slug
-      configuration['label'].parameterize
+      "#{readonly_marker}#{configuration['label'].parameterize}"
+    end
+
+    def readonly_marker
```
cbeer: `field_prefix`, maybe? I can imagine a similar pattern for e.g. multilingual values.

tpendragon:
@cbeer Are the failures here because of Blacklight 6?

tpendragon:
@cbeer Take a look?

PR #1450, opened 2016-04-26
Fix tests.

tpendragon:
Looks like Capybara's default functionality is to hide invisible
elements. Proposing we go with it rather than turning off Capybara's
 #ignore_hidden_elements.


tpendragon:
@cbeer Can you take a look?

PR #1460, opened 2016-06-30
Reword default invitation emails to remove all links besides the accept invitation action

cbeer:
Part of sul-dlss/exhibits#201


ggeisler:
👍 

I would merge but it appears that I don't have write access for projectblacklight. If you can give me that it would be handy...

cbeer:
Fixed.

PR #1480, opened 2016-07-08
Replace mail_form with ordinary ActionMailer class

### LINKED ISSUE #1478: Remove mail_form.

cbeer:
Fixes #1478

### DIFF
--- a/app/controllers/spotlight/contact_forms_controller.rb
+++ b/app/controllers/spotlight/contact_forms_controller.rb
@@ -11,11 +11,7 @@ module Spotlight
 
     def create
       if @contact_form.valid?
-        if @contact_form.respond_to? :deliver_now
-          @contact_form.deliver_now
-        else
-          @contact_form.deliver
-        end
+        ContactMailer.report_problem(@contact_form).deliver_now
 
         redirect_to :back, notice: t(:'helpers.submit.contact_form.created')
       else
--- a/app/views/spotlight/contact_mailer/report_problem.html.erb
+++ b/app/views/spotlight/contact_mailer/report_problem.html.erb
@@ -1,0 +1,14 @@
+<p><%= @contact_form.name %></p>
+<p>From: <%= %("#{@contact_form.name}" <#{@contact_form.email}>) %></p>
+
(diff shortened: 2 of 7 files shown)


app/controllers/spotlight/contact_forms_controller.rb
```diff
@@ -11,11 +11,7 @@ def new
 
     def create
       if @contact_form.valid?
-        if @contact_form.respond_to? :deliver_now
-          @contact_form.deliver_now
-        else
-          @contact_form.deliver
-        end
+        ContactMailer.report_problem(@contact_form).deliver_now
```
jcoyne: Why not deliver_later?

app/views/spotlight/contact_mailer/report_problem.html.erb
```diff
@@ -0,0 +1,14 @@
+<p><%= @contact_form.name %></p>
+<p>From: <%= %("#{@contact_form.name}" <#{@contact_form.email}>) %></p>
+
+<p><%= @contact_form.message %></p>
+
+----
+
+<dl>
+  <dt>REMOTE_IP</dt><dd><%= @contact_form.request.remote_ip %></dd>
```
jcoyne: How does `request` get set on ContactForm?

jcoyne: Nevermind: https://github.com/projectblacklight/spotlight/blob/9017a229113f8ffb216b86c0c4356cc10be7b83c/app/controllers/spotlight/contact_forms_controller.rb#L27

app/controllers/spotlight/contact_forms_controller.rb
cbeer: We weren't doing it before (and things end up getting ugly; ActiveJob can't serialize `Spotlight::ContactForm`, so you end up tacking on a bunch of primitive arguments instead)

jcoyne: We weren't doing it before because MailForm didn't support delivering later. But not being serializable is a good reason.

PR #1481, opened 2016-07-08
Unpin devise dependency

cbeer:


spec/models/spotlight/contact_email_spec.rb
```diff
@@ -16,12 +16,6 @@
     before { subject.email = 'foo@example.com' }
     it { is_expected.to be_valid }
 
-    describe 'when saved' do
-      it 'sends a confirmation' do
-        expect(subject).to receive(:send_devise_notification)
```
cbeer: devise moved the notifications into an `after_commit` hook, which is tricky to test. We could pull in https://github.com/grosser/test_after_commit and try to do it, but it seems like we're testing devise here anyway..

PR #1487, opened 2016-07-11
Install Riiif as a way to return carrierwave downloads

### LINKED ISSUE #1182: Add IIIF support for uploaded images using rack-iiif
Spotlight accepts image uploads using Carrierwave. By default, I believe these end up somewhere in `./public`. We should put a IIIF shim in front of these uploads, perhaps using the `rack-iiif` or `riiif` gems.

https://github.com/IIIF/rack-iiif
https://github.com/curationexperts/riiif

jcoyne:
Fixes #1182

### DIFF
--- a/app/services/spotlight/carrierwave_file_resolver.rb
+++ b/app/services/spotlight/carrierwave_file_resolver.rb
@@ -1,0 +1,8 @@
+module Spotlight
+  # Used by RIIIF to find files uploaded by carrierwave
+  class CarrierwaveFileResolver < Riiif::AbstractFileSystemResolver
+    def pattern(id)
+      Spotlight::FeaturedImage.find(id).image.file.file
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/spotlight/carrierwave_file_resolver.rb
```diff
@@ -0,0 +1,8 @@
+module Spotlight
+  # Used by RIIIF to find files uploaded by carrierwave
+  class CarrierwaveFileResolver < Riiif::AbstractFileSystemResolver
+    def pattern(id)
+      Spotlight::Masthead.find(id).image.file.file
```
cbeer: Just mastheads for now? Can we use `Spotlight::FeaturedImage` and get it for thumbnails too?

jcoyne: Sure, this is just the thin end of the wedge.

Issue #1503, opened 2016-07-19
Opengraph helper should return meta tags containing the opengraph metadata from the Spotlight resource


mjgiarlo:
The [opengraph helper](https://github.com/projectblacklight/spotlight/blob/master/app/views/layouts/spotlight/spotlight.html.erb#L24) should return meta tags containing the opengraph metadata from the Spotlight resource (to enable, e.g., Facebook rich snippets). Originally surfaced as part of sul-dlss/exhibits#218.

The helper currently returns an empty string. FWIW, I don't consider this bug high priority, because Facebook is somewhat smart about inferring snippet properties from the page. *I.e.*, snippets ​already​ work in Facebook without fixing this helper. (The advantage of fixing this is we'd have more predictability and control over how snippets appear.) Note too that other consumers of opengraph metadata may not do as good a job at inferring properties.

mjgiarlo:
Note: #1506 may be removing indexing of open graph properties. See that PR for more info.

jcoyne:
@mjgiarlo the OpenGraph stuff is for harvesting open graph resources into your exhibit. It's not for producting OpenGraph fields on your exhibit documents

PR #1506, opened 2016-07-20
Extract Solr document building from the resource

jcoyne:
This positions us to run indexing for a document independent of a single
exhibit.

This would require changes to `spotlight-dor-resources`.

https://github.com/sul-dlss/spotlight-dor-resources/blob/master/app/models/spotlight/resources/dor_harvester.rb#L18

### DIFF
--- a/app/models/concerns/spotlight/resources/open_graph.rb
+++ b/app/models/concerns/spotlight/resources/open_graph.rb
@@ -23,10 +23,6 @@ module Spotlight
           ["#{k.parameterize('_')}_tesim", v]
         end]
       end
-
-      def to_solr
-        super.merge(opengraph_properties)
-      end
     end
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/concerns/spotlight/resources/open_graph.rb
```diff
@@ -23,10 +23,6 @@ def opengraph_properties
           ["#{k.parameterize('_')}_tesim", v]
         end]
       end
-
-      def to_solr
```
mjgiarlo: I see these properties being used in our tests but otherwise we don't seem to be indexing `opengraph_properties`.

mjgiarlo:
This refactoring looks good to me except for the bit about `opengraph_properties` no longer getting indexed. Can you comment on that, @jcoyne?

jcoyne:
@mjgiarlo we aren't using OpenGraph anywhere that I can see, and @cbeer wrote it, so it didn't seem like a priority.  I don't think we need to rush this in. We can talk it over with @cbeer when he's back next week.

mjgiarlo:
This is related to #1503. I'm OK with nuking this in this PR as long as we make a connection to that ticket (which I think I just did ;) ).

PR #1509, opened 2016-07-20
This illustrates the fundamental problem with #205

jcoyne:
This behavior is triggered by `CatalogController#update` here:

https://github.com/projectblacklight/spotlight/blob/c0b782dda044faa6a724db18934f1063b3209cb7/app/controllers/spotlight/catalog_controller.rb#L78

### DIFF
--- a/spec/models/solr_document_bug_spec.rb
+++ b/spec/models/solr_document_bug_spec.rb
@@ -5,6 +5,10 @@ describe SolrDocument, type: :model do
 
   describe '#save' do
     context 'when filter_resources_by_exhibit is true' do
+      let(:expectation) do
+        hash_including(:"exhibit_#{exhibit.slug}_public_bsi",
+                       :"exhibit_#{exhibit_alt.slug}_public_bsi")
+      end
       before do
         Spotlight::Engine.config.filter_resources_by_exhibit = true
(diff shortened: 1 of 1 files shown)


mjgiarlo:
@jcoyne I'm not sure I see the fundamental problem. This commit fixes the broken test:

https://github.com/projectblacklight/spotlight/pull/1509/commits/24be8c24e27d5da4a74d65079b8c677aab949d24

Is the fundamental problem the *names* of the hash keys? that's all I materiall changed in the spec.

jcoyne:
@mjgiarlo no your fix is not valid.  It  should have a key like `spotlight_exhibit_slug_#{exhibit.slug}_bsi` and a key like `exhibit_#{exhibit.slug}_public_bsi`

mjgiarlo:
@jcoyne OK, so the expectation should be that both the visibility fields and the filter fields *both* appear for each exhibit?

jcoyne:
Yes. Both fields must be present.

cbeer:
Yes - it's two different things:

- the filter field makes it available to the curator for the exhibit
- the visibility field makes it visible to the end users

mjgiarlo:
OK, thanks, @jcoyne @cbeer. Closing this PR now.

PR #1511, opened 2016-07-25
Add DOCTYPE markup to spotlight layout

cbeer:


cbeer:
Fixes part of https://github.com/sul-dlss/exhibits/issues/256.

PR #1513, opened 2016-07-25
Remove default_scope from Spotlight::Exhibits, and add an explicitly ordered scope instead

cbeer:


mjgiarlo:
@cbeer One :red_circle:

PR #1524, opened 2016-07-28
Cast #last_indexed_finished to a Time object

cbeer:


app/models/spotlight/resource.rb
```diff
@@ -120,6 +123,13 @@ def write_to_index(batch)
                                headers: { 'Content-Type' => 'application/json' }
       end
 
+      # Wrap update call in a transaction
+      def update_last_indexed_count!(count)
+        transaction do
+          update(last_indexed_count: count)
```
jcoyne: What is the transaction helping here?

PR #1528, opened 2016-07-30
Explicitly require bundler dependency in the application template

cbeer:


jcoyne:
What does this fix?

cbeer:
Something in the Rails 4 stack gives us bundler for free, not so with Rails 5.

PR #1530, opened 2016-08-01
Remove unused method: generated_feature_methods

jcoyne:


cbeer:
I'm pretty sure this is an override to support some feature Rails expects to have; I'll look into it, and make sure we're not missing a test to exercise it.

cbeer:
Looks like it was a

PR #1547, opened 2016-08-03
Generate list of email addresses of current site administrators, exhibit administrators, and exhibit curators

### LINKED ISSUE #1472: Generate list of email addresses of current exhibit administrators
For service purposes, it would be helpful to be able to generate a list email addresses of all users (administrators and curators) associated with all exhibits. The email list could then be used by the institution service team to communicate with exhibit admins/curators.

Some design details TBD.

mjgiarlo:
Fixes #1472

Implementation:

screenshot from 2016-08-03 10-06-44

Copy button color changes on hover:

screenshot from 2016-08-03 10-06-50

Addresses are highlighted when button is pressed (and they are copied to the clipboard):

screenshot from 2016-08-03 10-06-55

### DIFF
--- a/spec/features/site_admin_management_spec.rb
+++ b/spec/features/site_admin_management_spec.rb
@@ -1,6 +1,8 @@
 describe 'Site admin management', js: true do
   let(:user) { FactoryGirl.create(:site_admin) }
-  let(:existing_user) { FactoryGirl.create(:exhibit_visitor) }
+  let!(:existing_user) { FactoryGirl.create(:exhibit_visitor) }
+  let!(:exhibit_admin) { FactoryGirl.create(:exhibit_admin) }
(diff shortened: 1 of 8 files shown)


spec/features/site_admin_management_spec.rb
```diff
@@ -11,6 +13,28 @@
     expect(page).to have_css('td', text: user.email)
   end
 
+  describe 'copy email addresses' do
+    it 'displays the email addresses of site admins' do
+      expect(page).to have_css('div#all_users', text: user.email)
+    end
```
jcoyne: Please combine these assertions into a single it block.  This test is very slow because it does several http requests for each `it` block.  If you need more fine grained tests, consider using a view spec.

mjgiarlo: Done.

PR #1555, opened 2016-08-03
Hide facet config rows for facets containing zero items. Fixes #1438.

### LINKED ISSUE #1438: Hide facet from the Search > Facets tab if that facet has no items
Showing facets on the Configuration > Search > Facets page when they have 0 items is potentially confusing for the exhibit curator (the concern has been expressed that curators might think there is some problem with the exhibit item records if they see a facet with 0 items).

Could we filter out facets with 0 items from the Configuration > Search > Facets page to eliminate this potential confusion? An additional benefit of doing this is it would keep the Facets page shorter (e.g., for the SUL case, we now have a good handful of facets that won't be applicable to most exhibits).

(It's probably obvious, but the arrows in the screenshot below is an example of facets that would not be shown with this approach.)

configuration_-_search___the_bob_fitch_photography_archive_-_online_exhibits

mjgiarlo:


spec/views/spotlight/search_configurations/_facets.html.erb_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  before do
+    assign(:exhibit, exhibit)
+    assign(:blacklight_configuration, exhibit.blacklight_configuration)
+    allow(view).to receive_messages(current_exhibit: exhibit,
+                                    translate_sort_fields: '',
+                                    blacklight_config: exhibit.blacklight_configuration)
+    allow(field_metadata).to receive(:field).with('genre_ssim').and_return(empty_metadata)
+    allow(field_metadata).to receive(:field).with('personal_name_ssm').and_return(nonempty_metadata)
```
atz: Maybe a single `with(any_args)` is sufficient for all the nonempty cases?

PR #1560, opened 2016-08-04
Generate an initializer that shows all the possible config values; fixes #1353

### LINKED ISSUE #1353: Generate an initializer that shows all the possible config values
This could be completely commented out. It would make it easier to discover which config options are available.

cbeer:


mjgiarlo:
Needs rebase.

PR #1575, opened 2016-08-11
Refactor Spotlight::ResourcesController to use cancan's resource loading

cbeer:


jcoyne:
Should this have a test?

cbeer:
I'm not sure it needs one (here, at least); as far as Spotlight is concerned, everything is the same.

PR #1587, opened 2016-08-17
Move SolrDocumentSidecar#index_status to a new field

cbeer:


atz:
Addresses https://github.com/sul-dlss/exhibits/issues/330

jcoyne:
Do we need a data migration?

PR #1591, opened 2016-08-18
Preserve user pagination parameters in browse category searches

cbeer:
Fixes https://github.com/sul-dlss/exhibits/issues/331


app/helpers/spotlight/pages_helper.rb
```diff
@@ -21,8 +21,8 @@ def disable_save_pages_button?
     end
 
     def get_search_widget_search_results(block)
-      if block.searches?
-        search_results(search_widget_search_query(block))
+      if block.searche.present?
```
jcoyne: "serche"?

PR #1609, opened 2016-09-08
A fix for batch CSV imports containing UTF-8 text.

bgoodmon:


bgoodmon:
The Travis CI build is failing because it's waiting for some input when initializing paper_trail.rb.  How should I proceed? The initializer is asking if it should overwrite the file <something>/initializers/paper_trail.rb.  I saw the same issue when I was running rake ci in my local repo - there I was able to respond "Y" and proceed.   The error message is shown below:

 initializer  paper_trail.rb
ig/initializers/paper_trail.rb? (enter "h" for help) [Ynaqdh] 
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated

app/controllers/spotlight/resources/csv_upload_controller.rb
```diff
@@ -14,7 +14,7 @@ class CsvUploadController < ApplicationController
 
       def create
         file = csv_params[:url]
-        csv = CSV.parse(file.read, headers: true, return_headers: false, encoding: 'utf-8').map(&:to_hash)
+        csv = CSV.parse(File.read(file.tempfile.to_path.to_s, encoding: 'UTF-8'), headers: true, return_headers: false, encoding: 'utf-8').map(&:to_hash)
```
cbeer: Are we sure all uploads have a backing `tempfile`? The Rails documentation suggests in some cases it'll be a simple `StringIO` object: http://guides.rubyonrails.org/form_helpers.html#uploading-files

bgoodmon: I am sure that there will be a tempfile.  My reasoning is that the variable "file" refers to an instance of ActionDispatch::Http::UploadedFile, not an instance of IO.  I can see this in the debugger.  UploadedFile by definition always has a tempfile according to http://api.rubyonrails.org/classes/ActionDispatch/Http/UploadedFile.html since its purpose is to automatically close and dispose of the file when the read method is called.

If you would prefer, here's another way we could fix the bug:

        csv = CSV.parse(file.read.force_encoding('UTF-8'), headers: true, return_headers: false, encoding: 'utf-8').map(&:to_hash)

Is there any reason to prefer one solution over the other?

jkeck:
@bgoodmon FWIW your issue with the build hanging on PaperTrail should be solved now if you rebase on top of master.

bgoodmon:
Thanks for the fix!

bgoodmon:
I can confirm that this fixes the problem we were experiencing.  Thanks again for your work.

cbeer:
👏 thanks for checking. This fix was released in blacklight-spotlight 0.30.

PR #1610, opened 2016-09-15
Parameterize configuration keys in Nestable's data-id attribute.

### LINKED ISSUE #1608: Search Configuration Sort Fields drag and drop/reordering works -- but isn't saving
Reported by curator. @ggeisler verified: I experience the same problem as Daniel reported. I tried moving title from the last to the first field but after saving it is still listed last. The reorder *appears* to work, but saving the new order is not successful. I tried using Chrome, Safari, and FF, just to be sure it wasn’t a browser bug.

jkeck:
Fixes #1608 

## Before
sort-fields-before

## After
sort-fields-after

@ggeisler it would be helpful if you could validate that the behavior in the **Before** gif was what you were seeing on the exhibits site (particularly the empty dotted border).


ggeisler:
@jkeck Yep, your Before gif is the behavior we currently see, with the empty dotted slot.

PR #1611, opened 2016-09-15
Remove explicit disabling of papertrail associations in test app…

jkeck:
…generator.

This file is already generated as-is by the papertrail installer.

See https://github.com/airblade/paper_trail/blob/c37e48973d8006edf635c72d5c742ce2d76c2f25/lib/generators/paper_trail/install_generator.rb#L38 which appears to set it to false by default.

### DIFF
--- a/blacklight-spotlight.gemspec
+++ b/blacklight-spotlight.gemspec
@@ -29,7 +29,7 @@ these collections.)
   s.add_dependency 'mini_magick'
   s.add_dependency 'bootstrap_form', '~> 2.2'
   s.add_dependency 'acts-as-taggable-on', '>= 4.0.0.pre'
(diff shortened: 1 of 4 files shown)


jkeck:
@cbeer do you have any insight into why this was added?  As best I can tell, the same file is generated  (sans some leading whitespace) by `papertrail:install`, so we're just overwriting it with the same content.

cbeer:
Nope; I'd guess papertrail used to not do it?

jkeck:
Yeah, looks like it was added 21 days ago: https://github.com/airblade/paper_trail/commit/7d60a41e131983af5e232b4c0e1118b6513cbc18

Now onto the next thing broken. 😿

cbeer:
Can you pin the papertrail version?

PR #1613, opened 2016-09-17
Fix autocomplete bug in browse category and feature page blocks

### LINKED ISSUE #1606: Autocomplete not working for Browse and Feature Page widgets
When using the Browse Categories widget and the Feature Page widget, typing the name of a browse category or feature page does not change the items shown in the selection overlay. Autocomplete does work in widgets where you are typing the name of an item, rather than a browse category or feature page title.

In this example below, "Portraits" is a valid browse category title.

curation_-_edit_page___new_exhibit_for_yale_-_online_exhibits

This is a blocker for at least one current exhibit, where the curator is not able to select desired feature page titles from a long list (the desired title isn't visible in the default selection overlay due to #1603).

jkeck:
Closes #1606 

Allow blocks to extend Bloodhound options.

The default behavior is what the existing behavior was.

The browse and pages block extend with their own options to switch from full remote to prefetch.

## Before
feature-pages-before

## After
feature-pages-after


jkeck:
I'm pushing this up just to share progress.

This is failing the tests right now because the way the autocomplete works for prefetched results seems to not be compatible w/ how we fill-in the autocomplete in our tests.

One issue with the `prefetch` approach is that the autocomplete results are not opened until you start typing (as opposed to the `remote` approach which opens them up as soon as you focus on the input).  This might be why the tests are failing, but not entirely sure.

I'm not sure if that difference will be a problem or not.  If it is a problem, then we may need to figure out a way to have our JSON response for non-solr backed objects respond to a query parameter (but that sounds like more trouble that it's worth).

PR #1614, opened 2016-09-20
Make autocomplete results with many items more usable

### LINKED ISSUE #1603: Thumbnail selection options are limited & need scrolling through available options
See screenshot. This is an exhibit with hundreds of photos. When entering a keyword to select a thumbnail to display for a feature or browse page, the first ~5 thumbnails in the search return are displaying & that is all. Curators need to be able to scroll through the thumbnails for all available results to select the best thumbnail.
screen shot 2016-08-28 at 3 22 06 pm

jkeck:
Closes #1603 

- [x] Allow more that 10 items to be returned in autocomplete results.
- [x] Allow autocomplete pop-up to scroll after ~5 items.
- [x] Override sir-trevor block overflow styling to prevent the autocomplete pop-up from being obscured.

## Before
dropdown-before

## After
dropdown-after

### DIFF
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -53,7 +53,7 @@ module Spotlight
     # results when a partial match is passed in the "q" parameter.
     def autocomplete
       search_params = params.merge(search_field: Spotlight::Engine.config.autocomplete_search_field)
-      (_, @document_list) = search_results(search_params.merge(public: true))
+      (_, @document_list) = search_results(search_params.merge(public: true, rows: 100))
 
       respond_to do |format|
         format.json do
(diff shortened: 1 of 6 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -53,7 +53,7 @@ def show
     # results when a partial match is passed in the "q" parameter.
     def autocomplete
       search_params = params.merge(search_field: Spotlight::Engine.config.autocomplete_search_field)
-      (_, @document_list) = search_results(search_params.merge(public: true))
+      (_, @document_list) = search_results(search_params.merge(public: true, rows: 100))
```
jkeck: Not sure if this is the best place to merge the row param in.

Was also thinking about making this configurable, but it's also used in javascript in a few places so `¯\_(ツ)_/¯`.

PR #1617, opened 2016-09-30
Add index for solr document sidecar document lookup

cbeer:


jcoyne:
Is this the 3 second thing?

cbeer:
No, that was unrelated: https://github.com/sul-dlss/stacks/pull/109.

Issue #851, opened 2015-01-12
Multi-image object support in adding non-repository content


snydman:
As as user I want to upload a single item that consists of multiple images.  Like a book.

elrayle:
We have had this request as well.

PR #1631, opened 2016-11-09
Compound object

sephirothkod:
This implements compound objects for pictures only. It can be expended for audio/video when that feature is added to spotlight.


cbeer:
Thanks @sephirothkod. Can you add some screenshots to this ticket of this feature in action?  I'm not sure I understand what compound objects are, and how this feature is intended to work.

Also, in this project, we use a style guide (enforced by rubocop); you'll note some of the notices when running the project's tests (and in travis, e.g. https://travis-ci.org/projectblacklight/spotlight/jobs/174554036#L349). Some of them seem easy (like using 2 spaces instead of tabs). I'll try to add some additional comments inline about some non-style changes that would help this fit in better.

app/assets/javascripts/spotlight/application.js
```diff
@@ -24,5 +24,6 @@
 //= require bootstrap-tagsinput
 //= require jquery.serializejson
 //= require clipboard
+//= require dragula.min
```
cbeer: The `dragula.min.js` should probably be in `vendor` (as an external component) or (even better) fully external (as we do with e.g. openseadragon-rails).

Elsewhere in the project, we've been using `Nestable.js`. I'm a little wary of using two different drag + drop libraries. Dragula looks pretty nice, though, so do you think we should investigate swapping out Nestable for it?

app/controllers/spotlight/catalog_controller.rb
```diff
@@ -39,7 +39,15 @@ class CatalogController < ::CatalogController
 
     def show
       super
-
+      @docs = []
+      if  @document.sidecars.length>0 && @document.sidecars.first.compound?
```
cbeer: It seems like you're repeating this logic both here and in `edit`.  Maybe an accessor on the `SolrDocumentSidecar` model would help?

app/controllers/spotlight/catalog_controller.rb
```diff
@@ -39,7 +39,15 @@ class CatalogController < ::CatalogController
...
+		  @document.sidecars.first.data["configured_fields"]["items"].each do |cur_id|
```
cbeer: I'm not sure `sidecars.first` is what we really want, as `SolrDocument#sidecars` returns all the sidecars for an object across all exhibits. Maybe `@document.sidecar(current_exhibit)` would be better?

app/controllers/spotlight/compounds_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+	class CompoundsController < Spotlight::ApplicationController
+		include Blacklight::SearchHelper
+		load_and_authorize_resource :exhibit, class: Spotlight::Exhibit, prepend: true
+		before_action :authenticate_user!, only: [:new, :create, :update, :edit]
+		before_action :check_authorization, only: [:new, :create, :update, :edit]
+		require 'fileutils'
+		
+		def index
+			@exhibit = Spotlight::Exhibit.find params[:exhibit_id]
+			@resources = Spotlight::Resource.where("data like '%items%' and exhibit_id = '#{@exhibit.id}'")
```
cbeer: This raw query is surprising. If we need to handle these differently, maybe a subclass of `Spotlight::Resource` (like we're doing for the uploaded resources) would be more appropriate?

app/controllers/spotlight/compounds_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+		def update
+			begin
+				@exhibit = Spotlight::Exhibit.find params[:exhibit_id]
+				id = params[:id].split('-')[1].to_i
+				@resource = Spotlight::Resource.find(id)
+				@resource.data = params[:data].to_hash
+				FileUtils.rm_rf(Dir.glob(@resource.url.file.file.dup.sub(@resource[:url], "*")))
+				@thumb_resource = Spotlight::Resource.find(params[:data][:items].first.split("-").last)
+				@resource[:url] = @thumb_resource[:url]
+				copy_files
```
cbeer: Is it possible to use carrierwave (as we're doing elsewhere) here? Offhand, it seems pretty similar.

app/controllers/spotlight/compounds_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+				@resource.sidecar.data["configured_fields"] = params[:data].to_hash
+				@resource.sidecar.save
+				@resource.save_and_index
+				redirect_to "/spotlight/#{params[:exhibit_id]}/dashboard", notice: 'compound object updated successfully'
```
cbeer: Please use Rails' `_url` or `_path` helpers to build these routes.

app/views/spotlight/catalog/_curation_mode_toggle_default.html.erb
```diff
@@ -1 +1,5 @@
-<%= exhibit_edit_link document, [:edit, current_exhibit, document], class: 'pull-right btn btn-primary' if can?(:curate, current_exhibit) and !current_page? [:edit, current_exhibit, document]  %>
\ No newline at end of file
+<% if(document.sidecars.length>0 && document.sidecars.first.compound?) %>
+	<%= link_to "Edit Compound Object", edit_exhibit_compound_path({exhibit_id: current_exhibit, id: document.id}) , class: 'pull-right btn btn-primary' if can?(:curate, current_exhibit) and !current_page? [:edit, current_exhibit, document]  %>
```
cbeer: Echoing above, this seems like even more motivation to create a new `Spotlight::Resource` subclass. Rails' polymorphic routing may do the right thing for you for free.

app/views/spotlight/catalog/_openseadragon_default.html.erb
```diff
@@ -0,0 +1,57 @@
+<% image = document.to_openseadragon(blacklight_config.view_config(:show)) %>
+<%
+	image = []
+	@docs.each do |doc|
+		image << doc.to_openseadragon(blacklight_config.view_config(:show)).first.first.first
```
cbeer: If we changed `document.to_openseadragon` to implement this behavior, would that allow us to avoid overriding this template?

app/views/spotlight/compounds/_document_admin_table.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <tr>
+      <th class="col-sm-1"></th>
+      <th class="col-sm-5"><%= t(:'spotlight.catalog.fields.title') %></th>
+      <th class="col-sm-2"><%= t(:'spotlight.catalog.fields.date_added') %></th>
+      <th class="col-sm-1 checkbox-toggle"><%= t(:'spotlight.catalog.fields.visibility') %></th>
+    </tr>
+  </thead>
+  <% @documents.each do |document| %>
+  <tr>	
+  <td><a href="<%=exhibit_solr_document_path({exhibit_id: @exhibit.slug, id: document._source[:id]})%>">
```
cbeer: In general, we prefer using methods on `SolrDocument`, instead of digging into `_source`. For example, `document._source[:id]` is better accessed as `document.id` or `document._source[]` as `document[]`

cbeer (changes requested):

sephirothkod:
Hi Chris,

Thanks for letting me know about all the stylistic stuff that rubocop finds. I was just running "rake ci" and it wasn't running rubocop but it looks like if I just run "rake" it runs rubo cop? so I'll make the changes and keep that in mind for the future.

app/controllers/spotlight/compounds_controller.rb
sephirothkod: I'm not super familiar with carrier wave I'll look into though.

app/controllers/spotlight/compounds_controller.rb
sephirothkod: Hmm, yes this might work better!

app/assets/javascripts/spotlight/application.js
sephirothkod: dragula is really nice and simple! however it doesn't allow for nesting objects so if we were to implement it into spotlight to replace nestable.js we would have to come up with a new way to do that for feature pages. might be best to stick with nestable and ill have to change my code at some point so that it uses that in stead.

sephirothkod:
new
view

sephirothkod:
The idea of compound objects is to allow exhibit makers to keep books all grouped together in one object rather than having to add each individual page to the exhibit the can add the entire book all in one go.

The next iteration will have each page's metadata displayed along with the object metadata, and change when you "flip" the page.

Here is one of our working examples here at UVic: http://exhibits.library.uvic.ca/spotlight/wwi/catalog/1-1314

sephirothkod:
Hi @cbeer I believe I have made all the necessary changes for this branch there are only two errors from ci. the fact that routes has an extra line (required for the compound objects) and one of my functions has a large Assignment Branch Condition size. If needed I can break up the function so it meets the ABC requirement but I can't do anything about the routes as that extra line needs to be there.

Cheers

cbeer:
Now that we support IIIF manifests, I wonder if support for compound objects can be reframed as either:

- IIIF manifest building (https://github.com/bodleian/iiif-manifest-editor? Something from PUL?)
- Support for hierarchical resources (perhaps in scope for Arclight?)

sephirothkod:
It probably could be re-framed but I will not be able to do it. My time for working on spotlight was 3 months ago when I submitted this code; I have other projects that need my attention right now. I think we should discuss how better to collaborate on this project during LDCX at the end of March. I will have some more time to work on Spotlight during the spring/summer so we should be able to work things out in March/April, in advanced of that time period.

hackartisan:
This pull request is currently inactive and is being closed. Please use #851 for further discussion. Thank you!

PR #1640, opened 2016-12-06
Update to sir-trevor 0.6.4

cbeer:


cbeer:
The build failure ought to be fixed by #1639.

PR #1647, opened 2016-12-09
remove remaining jcrop references

### LINKED ISSUE #1623: Jcrop.gif is referenced, but not provided
https://github.com/projectblacklight/spotlight/blob/37f6a4c266db9aa9d2a59529340a634d1796fefc/app/assets/stylesheets/spotlight/_croppable.scss#L2

jmartin-sul:
also, get rid of no-longer-applicable `rubocop:disable` statement

note that there are still a couple references to jcrop in add_contacts_spec.rb, but those should be taken care of by PR #1643.

closes #1623

### DIFF
--- a/app/assets/stylesheets/spotlight/typeahead.css
+++ b/app/assets/stylesheets/spotlight/typeahead.css
@@ -37,7 +37,6 @@
 }
 
 .twitter-typeahead {
-  z-index: 1000 !important; /* get typeahead to float on top of e.g. jcrop */
   .tt-suggestion {
     font-size: 18px;
     line-height: 24px;
--- a/lib/spotlight/engine.rb
+++ b/lib/spotlight/engine.rb
@@ -15,7 +15,6 @@ require 'clipboard/rails'
 module Spotlight
   ##
   # Spotlight::Engine
-  # rubocop:disable Metrics/ClassLength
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/spotlight/typeahead.css
```diff
@@ -37,7 +37,7 @@
 }
 
 .twitter-typeahead {
-  z-index: 1000 !important; /* get typeahead to float on top of e.g. jcrop */
```
cbeer: If we've done away with jcrop, I wonder if we can remove this override? I think it happens on e.g. the appearance page when you're cropping a masthead and try to select a different item from the repository.

jmartin-sul: sure, i'd be happy to just remove the line... i wasn't sure whether it was useful for other things, since it said "e.g." jcrop.

jmartin-sul: removed the override and things still appeared to work fine when testing on my laptop:

before:
screen shot 2016-12-11 at 5 18 28 pm

after removing (typeahead still falls over crop tool):
screen shot 2016-12-11 at 5 19 21 pm

pushed with the line omitted entirely (and rebased on latest iiif_crop).

PR #1649, opened 2016-12-09
Update Sir Trevor customized blocks

### LINKED ISSUE #1464: List widget display bug
There's an issue with the Sir Trevor List widget where each time the page in which it is embedded is saved, the list is indented, because the initial `ul > li` structure is wrapped in an additional `ul` on every page save.

First time the list is saved:
page_one___test_2_-_blacklight

Third time the list is saved:
page_one___test_2_-_blacklight 3

See the extra two `ul` tags that the initial list is wrapped in after the third save:

page_one___test_2_-_blacklight 4

### LINKED ISSUE #1604: Add formatting for quotes to quote widget
The Quote widget automatically puts the Credit in italics, but there is no other formatting on the quote or credit text to set it off from the rest of the content on the page. 
screen shot 2016-08-28 at 3 38 42 pm

In the attached image, the text "I think every scientist should consider...." was entered using the Quote widget.

### LINKED ISSUE #1618: List widget doesn't display list when entered & then saved
As a curator, I navigate to the list widget by using the admin console, navigating to Curation > Feature Pages, and clicking edit on a feature page I have created. In edit mode, I click on the "+" to add a widget element to my page, and then click again on the left-hand "+". This pulls up both standard & exhibit item widgets. The list widget is a standard widget choice.

When I use the list widget to add a list to a feature page or about page, I am able to create a list, and when I click on the save button at the bottom of the page, I get no indication my action has not saved. But -- when I view the "saved" feature or about page, my list does not display. 

I need the list widget to both save and subsequently display, so that I can use bulleted lists in the design/display of my exhibit.

tingulfsen:
Fixes #1464  and fixes #1618 

Some of our view customizations for Sir Trevor are out of date. This PR brings them up to date with what Sir Trevor Rails uses.

(now also fixes #1604)

### DIFF
--- a/app/views/spotlight/sir_trevor/blocks/_list_block.html.erb
+++ b/app/views/spotlight/sir_trevor/blocks/_list_block.html.erb
@@ -1,3 +1,11 @@
 <div class="st__content-block st__content-block--list">
-	<%= sir_trevor_markdown list_block.text %>
+  <% if list_block.format == :html %>
+    <ul>
+      <% list_block.listItems.each do |item| %>
+        <li><%= sir_trevor_format item[:content], format: list_block.format %></li>
+      <% end %>
+    </ul>
+  <% else %>
+    <%= sir_trevor_markdown list_block.text %>
+  <% end %>
 </div>
--- a/app/views/spotlight/sir_trevor/blocks/_quote_block.html.erb
+++ b/app/views/spotlight/sir_trevor/blocks/_quote_block.html.erb
@@ -1,13 +1,19 @@
 <div class="st__content-block st__content-block--quote">
   <div class="quote">
-    <div class="quote__content">
-    	<%= sir_trevor_markdown quote_block.text %>
-    </div>
-
-    <% if quote_block.cite.present? %>
-      <div class="quote__cite">
-        <cite class="quote__cite-text">&ndash; <%= quote_block.cite %></cite>
+    <blockquote>
+      <div class="quote__content">
+        <% if quote_block.format == :html %>
+            <%= sir_trevor_format quote_block.text, format: quote_block.format %>
+        <% else %>
+          <%= sir_trevor_markdown quote_block.text %>
+        <% end %>
       </div>
-    <% end %>
+
+      <% if quote_block.cite.present? %>
+        <div class="quote__cite">
+          <cite class="quote__cite-text">&ndash; <%= quote_block.cite %></cite>
+        </div>
+      <% end %>
+    </blockquote>
   </div>
 </div>
--- a/app/views/spotlight/sir_trevor/blocks/_text_block.html.erb
+++ b/app/views/spotlight/sir_trevor/blocks/_text_block.html.erb
@@ -1,3 +1,4 @@
 <div class="st__content-block st__content-block--text">
-	<%= sir_trevor_markdown text_block.text %>
+  <%# We prefer the GitHub markup here. %>
+  <%= sir_trevor_markdown text_block.text %>
 </div>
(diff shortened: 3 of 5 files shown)


tingulfsen:
Is the tweet_block type enabled at all in Spotlight? I don't see how to add content of that type in the UI.

app/views/spotlight/sir_trevor/blocks/_text_block.html.erb
```diff
@@ -1,3 +1,3 @@
 <div class="st__content-block st__content-block--text">
-	<%= sir_trevor_markdown text_block.text %>
+  <%= sir_trevor_format text_block.text, format: text_block.format%>
```
tingulfsen: Did we intentionally decide to use sir_trevor_markdown only, and if so, should the change here on line 2 be reverted?

app/views/spotlight/sir_trevor/blocks/_list_block.html.erb
```diff
@@ -1,3 +1,11 @@
 <div class="st__content-block st__content-block--list">
-	<%= sir_trevor_markdown list_block.text %>
+	<% if list_block.format == :html %>
```
jmartin-sul: looks like a tab instead of spaces here?

app/views/spotlight/sir_trevor/blocks/_quote_block.html.erb
```diff
@@ -1,7 +1,13 @@
...
   <div class="quote">
     <div class="quote__content">
-    	<%= sir_trevor_markdown quote_block.text %>
+      <% if quote_block.format == :html %>
+        <blockquote>
+          <%= sir_trevor_format quote_block.text, format: quote_block.format %>
+        </blockquote>
+      <% else %>
+        <%= sir_trevor_markdown quote_block.text %>
+      <% end %>
```
jmartin-sul: one option to offset both the quotation and the attribution more clearly would be to wrap the entire `<div class="quote">` in `<blockquote>`.  but @ggeisler says he's also fine w/ the sir trevor default, so if that's lower maintenance burden (i.e. if we're not doing any other customization of our own here), then we could just leave this as-is.  i have no strong opinion on it myself.

see also: https://github.com/projectblacklight/spotlight/pull/1651#issuecomment-267193593

tingulfsen:
I've updated this PR per the comments from @jmartin-sul - it would be great to get more feedback on this one.

app/views/spotlight/sir_trevor/blocks/_quote_block.html.erb
```diff
@@ -1,9 +1,13 @@
 <div class="st__content-block st__content-block--quote">
-  <div class="quote">
-    <div class="quote__content">
-    	<%= sir_trevor_markdown quote_block.text %>
-    </div>
-
+  <blockquote>
```
jkeck: I think there is something odd going on w/ the closing tags here.

It appears that perhaps the `blockquote` is never closed.

tingulfsen: 😳

cbeer: Is the `div.quote` closed too?

jkeck:
I just ran into this build failure locally myself.  See https://github.com/rubygems/rubygems/issues/1448.  Updating rubygems seemed to address this issue fore me (which we can do in a `before_install` as suggested by [TravisCI](https://docs.travis-ci.com/user/languages/ruby/#Upgrading-RubyGems)).

According to the issue in the rubygems repo this could also be addressed by using a diff. version of ruby (but didn't track that down because upgrading rubygems was easy enough).

cbeer:
Thanks, good sleuthing. I figure we might as well bump the ruby versions anyway, yeah? https://github.com/projectblacklight/spotlight/pull/1661

jkeck:
Yeah, absolutely.  Some people in the thread were indicating that they were having success with using earlier versions of ruby but I'm going to guess that's because it was so long ago.  Looks like rubygems gets bumped to a later version in [ruby 2.3](https://bugs.ruby-lang.org/issues/12326), so that should do it for us as well.

PR #1650, opened 2016-12-12
Update IIIF Cropper.

jkeck:


cbeer:
This updates it to https://github.com/sul-dlss/iiif-cropper/commit/46e91b4947250a08c75b63c1d9d880e821e2d259? Is it worth trying to cut a release from master?

jkeck:
@cbeer, yes it does.

I'm not 100% sure how to cut a release for that project and there is no mention of the release process in their README.  I would ask the maintainer, but he's on vacation.  If you know how to do that and can point me to anything so I can learn, I would be more than happy to cut a release of that.

cbeer:
Ok. Something to work out when he's back.

Issue #777, opened 2014-12-03
Should we touch about pages when saving contacts?


jkeck:
Currently the about page would not know that you had updated a contact and would continue to show you the cached version (@cbeer agrees this is probably the correct solution for this).

Are there other places where this may happen?  First thing that comes to my mind is when publishing feature pages it may update the sidebar of other feature pages (adding the link).

PR #1651, opened 2016-12-13
style sir trevor quote widget to make quotation more visually distinct

### LINKED ISSUE #1604: Add formatting for quotes to quote widget

jmartin-sul:
~~leaving as WIP because i wasn't sure if this needed a test, and if so, where that test should go.~~ (ran it by @jkeck, we agreed that a new test was unnecessary since it's just a small CSS addition)

i basically ripped off and adapted the styling that github uses for quotes.

after the styling change:
screen shot 2016-12-12 at 4 06 50 pm

closes #1604

### DIFF
--- a/app/assets/stylesheets/spotlight/_sir-trevor_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_sir-trevor_overrides.scss
@@ -17,6 +17,12 @@
   margin-bottom: 3*$padding-large-vertical;
 }
 
+.quote__content {
+  border-left: 0.4em solid $gray-lighter;
+  color: $gray;
+  padding: 0 1em;
+}
+
 // Sir-trevor sets the overflow to hidden which
 // obscures the bottom of things like the typeahead
 .st-blocks {


jmartin-sul:
requested review from @ggeisler

app/assets/stylesheets/spotlight/_sir-trevor_overrides.scss
```diff
@@ -17,6 +17,12 @@
   margin-bottom: 3*$padding-large-vertical;
 }
 
+.quote__content {
+  border-left: 0.4em solid $gray-lighter;
```
cbeer: I wonder if there's a more semantic variable than `$gray-lighter`.

app/assets/stylesheets/spotlight/_sir-trevor_overrides.scss
```diff
@@ -17,6 +17,12 @@
...
+  color: $gray;
```
cbeer: I assume `$text-muted` is probably too light?

cbeer:
Now that I've said all that, I wonder if we ought to just copy all of https://github.com/twbs/bootstrap-sass/blob/5d6b2ebba0c2a5885ce2f0e01e9218db3d3b5e47/assets/stylesheets/bootstrap/_type.scss#L240 for our use? Or update the sir-trevor template to use `<blockquote> [1]`?

[1] http://getbootstrap.com/css/#naming-a-source

app/assets/stylesheets/spotlight/_sir-trevor_overrides.scss
ggeisler: I think `$text-muted` is actually darker than `$gray-lighter` (#eee versus #777). So `$text-muted` might be fine (I was just going by the screenshot since I don't have an active Spotlight dev environment going at the moment).

ggeisler:
I suspect the `blockquote` styling would work fine. It looks like the font size might get a bit bigger but that was something I debated about suggesting anyway, so it would likely look fine.

jmartin-sul:
it turns out, the work that @tingulfsen's doing on pulling in upstream sir trevor changes might obviate the need for this.

with the updated sir trevor template, the quote looks like this (screenshot from tommy):
sirtrevor_quoteblock

one other option (which i tried yesterday after @cbeer suggested it) is to put the `blockquote` tag around the quote content and the attribution, which looks like this:
screen shot 2016-12-13 at 1 45 17 pm


the former seems like it'd be slightly easier from a maintenance burden perspective, since we get it for free with work that's already happening.  the latter seems like it offsets the whole quote/attribution combination a little more clearly, in the case where there's maybe no horizontal rule or similar separating it from surrounding content.

ggeisler:
@jmartin-sul  I agree with your assessment about the two options. Since curators have used quotes in at least a few different contexts that I remember (i.e., different types of blocks before and after the quote), and because we’d just be using the ST styling and not doing any overriding, I’d probably lean slightly towards using the ST option. I wouldn’t argue with either option though.

jmartin-sul:
thanks, @ggeisler, will go with the ST option.

closing this in favor of going with what we get from PR #1649.

PR #1653, opened 2016-12-13
Fix migration task to not error out when there is no image for a…

jkeck:
…contact.


jkeck:
This is going to be taken care of in a more comprehensive PR soon.

PR #1654, opened 2016-12-14
add non-visible field for catching spammy feedback submissions, add tests for field validation

jmartin-sul:
also, turn the email fields from text fields to proper HTML email fields.

here's a screenshot of the feedback form, with the new DOM element for the (visually) hidden form field in view in the debugger:
screen shot 2016-12-13 at 6 17 24 pm

paired w/ @jkeck.

closes sul-dlss/exhibits#345

### DIFF
--- a/app/models/spotlight/contact_form.rb
+++ b/app/models/spotlight/contact_form.rb
@@ -4,9 +4,16 @@ module Spotlight
   class ContactForm
     include ActiveModel::Model
 
-    attr_accessor :current_exhibit, :name, :email, :message, :current_url, :request
+    attr_accessor :current_exhibit, :name, :email, Spotlight::Engine.config.spambot_honeypot_email_field, :message, :current_url, :request
 
     validates :email, format: { with: /\A([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})\z/i }
+
+    # the spambot_honeypot_email_field field is intended to be hidden visually from the user,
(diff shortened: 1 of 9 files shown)


app/models/spotlight/contact_form.rb
```diff
@@ -4,10 +4,15 @@ module Spotlight
...
 
-    attr_accessor :current_exhibit, :name, :email, :message, :current_url, :request
+    attr_accessor :current_exhibit, :name, :email, :email_address, :message, :current_url, :request
 
     validates :email, format: { with: /\A([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})\z/i }
 
+    # the email_address field is intended to be hidden visually from the user, in hope that
+    # a spam bot filling out the form will enter a value, whereas a human with a browser wouldn't,
+    # allowing us to differentiate and reject likely spam messages.
+    validates :email_address, length: { is: 0 }
```
cbeer: Do we know that we need to call it something that sounds vaguely useful, instead of something that developers can read as bogus data?

cbeer: Or, if we have to call it something plausible, what do you think about pulling this name into, say, a `Spotlight::Engine.config` setting, so at least in the code it reads as obviously bogus?

jmartin-sul:
ok, i believe @cbeer's feedback has been incorporated as per discussion here and on PR #1655.

ready for review again...

cbeer:
👏 Thanks @jmartin-sul

PR #1655, opened 2016-12-14
[WIP] put the spambot honeypot field name in the config to make usage clearer in the code

jmartin-sul:
as per @cbeer's suggestion.

if we want to go this route, i think i'd prefer to not merge this PR into that branch, but would rather just squash both commits into the the base exhibits-345 branch.  this is more of a convenient place to see the differences in the two approaches.

### DIFF
--- a/app/controllers/spotlight/contact_forms_controller.rb
+++ b/app/controllers/spotlight/contact_forms_controller.rb
@@ -29,7 +29,7 @@ module Spotlight
     end
 
     def contact_form_params
-      params.require(:contact_form).permit(:name, :email, :message, :current_url)
+      params.require(:contact_form).permit(:name, :email, :email_address, :message, :current_url)
     end
   end
 end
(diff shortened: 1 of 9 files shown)


jmartin-sul:
hmm, looks like the template part of this version needs some tweaking to keep from 500ing.  can't seem to read `Spotlight::Engine.config.spambot_honeypot_email_field`?

need to head out, will revisit tonight or tomorrow morning.

jmartin-sul:
actually, this works, it turns out i just needed to restart my local instance for the config addition to get picked up.

jmartin-sul:
here's what this branch looks like vs master (instead of vs its predecessor version like in this PR):
https://github.com/projectblacklight/spotlight/compare/exhibits-345-can-spam-feedback_config-field-name

opinion on this approach vs the one in PR #1654?  @cbeer, @jkeck?  i'm happy to go either way.

cbeer:
Cool, thanks. I think this makes the intent clearer (even if it's because I wouldn't read the text around it.). Last question -- should those explanations be i18n strings?

PR #1662, opened 2017-01-11
reindexing jobs summary

### LINKED ISSUE #298: A visitor should be able to see a list of exhibits
Ref #296

Up in the dropdown maybe?  This would only work if there's a handful of things.

jmartin-sul:
screen shot 2017-01-18 at 12 17 31 am

screen shot 2017-01-18 at 12 21 31 am

* Sorts by job start time, in descending order, nulls sorted first (to more easily surface unstarted jobs).  This wasn't really covered in the mockup/description (didn't occur to me till after i'd started implementing), so nulls first was a behavioral assumption on my part.  Happy to change if @ggeisler or others would prefer to do otherwise.
* A log entry is created synchronously when the user hits the "Reindex items" command, and is left in the "Not Yet Started" state till a worker picks it up and starts reindexing the exhibit's resources.  My screenshot uses faked test data for this, since all of my laptop instance's reindexing is done synchronously and the jobs are short.  Guessing we can do more realistic testing in stage?

addresses sul-dlss/exhibits#298


spec/models/spotlight/reindexing_log_entry_spec.rb
```diff
@@ -0,0 +1,49 @@
+describe Spotlight::ReindexingLogEntry, type: :model do
+  subject { FactoryGirl.build(:reindexing_log_entry) }
+
+  describe 'scope' do
+    before(:all) do
+      # we only want to persist these rows for the duration a given test run...
+      DatabaseCleaner.start
```
jkeck: Hmm, was this added because of some odd behavior you were seeing?  I would think this would be not necessary due to https://github.com/projectblacklight/spotlight/blob/master/spec/spec_helper.rb#L56-L70.

jmartin-sul: yeah, i had a note to myself to ask you about that... i'd have expected the spec helper invocation to do the trick, but i noticed that i was running into a problem where the test user was hanging around between runs, and subsequent test runs would fail unless i manually deleted the test user from the db using rails console, or added the `DatabaseCleaner` invocations above.  but i was also surprised by the need for that.  would be happy to get rid of it if possible.  this was only on my laptop, FWIW (though i assume it's not an issue in travis, since it seems like the whole test env is recreated for each run there).

app/jobs/spotlight/reindex_job.rb
```diff
@@ -8,7 +8,20 @@ class ReindexJob < ActiveJob::Base
       resource_list(job.arguments.first).each(&:waiting!)
     end
 
-    def perform(exhibit_or_resources)
+    around_perform do |job, block|
+      log_entry(job).reindexing_started! if log_entry(job)
```
jkeck: I can't remember if I had asked this already, but can `log_entry(job)` be stored in a variable in this block?

app/models/spotlight/exhibit.rb
```diff
@@ -79,8 +80,8 @@ def solr_data
       end
     end
 
-    def reindex_later
-      Spotlight::ReindexJob.perform_later(self)
+    def reindex_later(user = nil)
```
jkeck: It looks like we attempt to send messages to the user object in the view unchecked, so it seems like the user is required in this case.  Should we explicitly require that a user object is passed?

spec/models/spotlight/reindexing_log_entry_spec.rb
jkeck: Meh, doesn't bother me too much.  If you found it was necessary then it's 🆗 by me.

app/models/spotlight/exhibit.rb
jmartin-sul: would you be ok with checking the usage in the view and having a view test that exercises a `user`-less log entry?  i think that'd be my slight preference, since a rake task (`import`) and and `ExhibitsController#process_import` both call `Exhibit#reindex_later` without a user.

app/jobs/spotlight/reindex_job.rb
jmartin-sul: i believe so, fixed it to do that and tests still passed

app/models/spotlight/reindexing_log_entry.rb
```diff
@@ -0,0 +1,42 @@
...
+  ##
+  # a log entry representing an attempt to reindex some number of records in an exhibit
+  class ReindexingLogEntry < ActiveRecord::Base
+    enum job_status: { unstarted: 0, in_progress: 1, succeeded: 2, failed: 3 }
+
+    belongs_to :exhibit, class_name: 'Spotlight::Exhibit'
+    belongs_to :user, class_name: '::User'
+
+    # null start times sort to the top, to more easily surface pending reindexing
+    default_scope { order('start_time IS NOT NULL, start_time DESC') }
```
cbeer: Do you know if this is this generic SQL?

app/models/spotlight/reindexing_log_entry.rb
```diff
@@ -0,0 +1,42 @@
...
+    def duration
+      end_time - start_time if end_time
+    end
+
+    def reindexing_started!
+      self.start_time = Time.zone.now
+      self.job_status = :in_progress
+      save
+    rescue
+      Rails.logger.error "unexpected error updating log entry to :in_progress from #{caller}"
```
cbeer: What sort of unexpected errors are we... expecting?

app/models/spotlight/reindexing_log_entry.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def reindexing_succeeded!
+      self.end_time = Time.zone.now
+      self.job_status = :succeeded
```
cbeer: I think Rails has some clever helpers for this, like `succeeded!`, maybe?

config/locales/spotlight.en.yml
```diff
@@ -1,4 +1,9 @@
 en:
+  time:
+    formats:
+      date:
+        formats:
+          default: '%Y-%m-%d %l:%M%P'
```
cbeer: I'm not sure it's our place to force a time format on downstream applications. At the very least, we should pick a different format name for our custom format.. but I wonder if we can just live with some of the Rails default ones?

app/models/spotlight/reindexing_log_entry.rb
jkeck: Yeah, I was wondering if @jmartin-sul was trying to be more explicit about the fact that `end_time` and `job_status` would be changing when `reindexing_succeeded!` was called.  Otherwise the save is a side affect of the `succeeded!` call and not immediately obvious when looking at this code (unless you know the particulars of `ActiveRecord#enum`.

I'm on the fence either way.

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: that's what the stack exchange post i found indicated.  i'll try and find a more solid reference.  my initial attempt was `start time DESC NULLS FIRST`, which it turns out is an oracle-ism and not generic SQL.

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: i think this was more of a vague feeling that reindexing a bunch of things asynchronously has the potential for errors... but thinking about it more, i guess something like a network error connecting to solr seems like the most likely possible problem to me?

also, it seems like it'd be a good idea to log the actual error.  not sure if that was part of what you were getting at, but seems like an oversight, now that i'm thinking about that line again.

would you also prefer a more realistic example error in the test?  or a comment in the test as to what sorts of things we might expect?

config/locales/spotlight.en.yml
jmartin-sul: i was just trying to format as per the mockup, happy to move/rename/remove if others would prefer.  i also sure that was the right place to put that, that's just what i'd gathered from reading i18n docs and discussion.

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: ah, i was unaware of that clever rails helper.  i'm having a little trouble finding documentation on a `succeeded!` method in active record or rails in general... could you provide a link or elaborate on usage, @cbeer?

i guess if we were worried about the `succeeded!` method auto-saving and callers being unaware, we could at least comment the method.  it seems like if a developer saw it, and was already familiar with the rails method, they'd know that it saved, and if they were unfamiliar with the method, they'd hopefully look at the code and comments to see how it works.  plus, the `!` convention signals some state change in the object itself.

config/locales/spotlight.en.yml
ggeisler: I'm not too set on the time format, but as @jmartin-sul and I discussed a bit previously, the goal was to present the elapsed time in a more user-friendly format than just hh:mm:ss. I assume that the person using this page, even in downstream applications, is most likely to be a librarian or other non-technical curator, and thus it would be nice to present the elapsed time in a friendly, easily-scannable format (which doesn't have to be exact).

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: regardless, after reading up a bit more on basic activerecord usage the other day, it seems to me that this might be more appropriately written in one line as
```ruby
update(start_time: Time.zone.now, job_status: :in_progress)
```
instead of the current 3 lines (and likewise for the similar sibling methods).

jkeck: @jmartin-sul just for edification re: enum: http://api.rubyonrails.org/classes/ActiveRecord/Enum.html

And the underlying rails code: https://github.com/rails/rails/blob/919a876a04d991bb5a984314ca985168cfbbbef5/activerecord/lib/active_record/enum.rb#L190

config/locales/spotlight.en.yml
jmartin-sul: actually, this part is only applied to the "Date" column (the job's start time).  the elapsed time is handled separately via a rails helper method that isn't influenced by this (AFAIK).

cbeer: It's totally the right way to do it, it's just not appropriate for Spotlight to decide the default time format for someone else's application. Maybe `l time, format: :short` (out of the box: 18 Jan 06:10) is roughly equivalent? Or, again, we can always invent a new format key and use that.

ggeisler: Ah, sorry, I didn't look closely enough at the code @cbeer was commenting on.

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: ok, got it, i see what you're saying.

yeah, so i did come across that functionality when reading up on enum, but didn't think to leverage it, maybe because it wouldn't set `end_time` by default, or maybe because i just hadn't used it previously.

would it be more conventionally rails-y to leave it as-is, or to override the `succeed!` method and have it do the time setting thing?  it does seem like both the status setting and start/end time setting should be encapsulated in a single method call.  is overriding something like `succeed!` and adding functionality encouraged or discouraged?

config/locales/spotlight.en.yml
jmartin-sul: if @ggeisler is ok with the `:short` format, i'll pull this section and switch over to that and adjust tests accordingly.  otherwise, i can just add a new key and call that.  i might be slightly in favor of the latter, but i don't feel strongly about it at all.

app/models/spotlight/reindexing_log_entry.rb
cbeer: Hm. Overriding `succeed!` , doing your thing, and calling `super` makes sense to me.

config/locales/spotlight.en.yml
cbeer: The big advantage of a built-in format is you'll get existing localizations for free. It's just a question whether that format is so important or semantically different that it deserves special handling.

ggeisler: @cbeer's point about localizations is good one and I'm not sure I'd want to argue that doing something special outweighs that. I really don't feel that strongly about the time formats, so leave it up to you guys to decide.

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: TL;DR:  i think it's generic SQL, and have some citations that imply that, but i couldn't find a single solid reference confirming that it is.

references i do have:
* the stack exchange post i initially relied on:  https://stackoverflow.com/a/7055259
* sqlite: ordering term can be an expr:  https://www.sqlite.org/syntax/ordering-term.html
* mysql: similar advice in old comments on docs, but can't find syntax diagram:  https://dev.mysql.com/doc/refman/5.7/en/working-with-null.html
* postgres: "The sort expression(s) can be any expression that would be valid in the query's select list":  https://www.postgresql.org/docs/8.3/static/queries-order.html
* and seems worth making explicit, because of differing default behavior in at least mysql and oracle:
  * mysql:  "When doing an ORDER BY, NULL values are presented first if you do ORDER BY ... ASC and last if you do ORDER BY ... DESC."    https://dev.mysql.com/doc/refman/5.7/en/working-with-null.html
  * oracle:  "If the null ordering is not specified then the handling of the null values is:  ... NULLS FIRST if the sort is DESC"    https://docs.oracle.com/javadb/10.8.3.0/ref/rrefsqlj13658.html

app/models/spotlight/exhibit.rb
jmartin-sul: added checking in the view to properly handle nil user, and a view test for nil user:
https://github.com/projectblacklight/spotlight/pull/1662/files#diff-b2c42be6424785aafeaae4f6bfe7a094R18
https://github.com/projectblacklight/spotlight/pull/1662/files#diff-c46dbe4a4b76306e1fd38e90047cd3b5R76

can do otherwise if that's preferred, though.

app/models/spotlight/reindexing_log_entry.rb
jmartin-sul: d'oh, sorry, i replied to this as if you were talking about the error handling in `reindex_job.rb`.

in this case, i think my desire was to be extra-cautious so that if logging itself went wrong, it wouldn't mess up things for the caller.  logging to the DB felt slightly more prone to flakiness than traditional logging to a log file.  the test is also slightly more contrived than it was before:  the mock error now gets raised on `start_time=`/`end_time=` in the tests, because i had trouble with placing expectations on `super`.

if this error handling feels like overkill, or prone to its own fragility, i can get rid of it.  i wasn't totally sure about it when i put it in originally.

jmartin-sul:
i believe i've addressed all the feedback so far.  an exhibits branch that uses this spotlight branch has been deployed to exhibits-stage, and should be available for testing.

app/models/spotlight/reindexing_log_entry.rb
cbeer: 👌 and fine by me. it's just new to me. if it wasn't generic, I'd have suggested alternate ways of sorting, or suggest we do it client side.

app/models/spotlight/exhibit.rb
```diff
@@ -108,5 +109,9 @@ def reindex_progress
     def sanitize_description
       self.description = ::Rails::Html::FullSanitizer.new.sanitize(description)
     end
+
+    def new_reindexing_log_entry(user = nil)
+      Spotlight::ReindexingLogEntry.create(exhibit: self, user: user, items_reindexed_count: resources.size, job_status: 'unstarted')
```
cbeer: Yeah, `resources.size` is the count of the explicitly provided resources, not the count including items in the collection. Note too that you won't want to calculate it synchronously either, since it can be expensive to calculate.

jmartin-sul:
paired with @cbeer on the refactoring he suggested, so i think this is ready for review again if/when it passes travis.

jmartin-sul:
also, the refactoring is all contained in the last commit.  i left the commits that were already there untouched.  those are pretty clean and mostly additive, most of the churn is in the last one.  that seemed logical since the refactoring leveraged the functionality that was added in the prior commits, but i'm happy to rejigger if people would strongly prefer the refactoring to be folded into the other commits (though it's not immediately apparent to me how that'd be structured).

PR #1667, opened 2017-01-21
Hide newly imported custom fields

tpendragon:


tpendragon:
@cbeer @jkeck Can you look at this?

jkeck:
Sorry for the delay @tpendragon.  Can you describe more about what the use case for this is?  I think we intentionally have the field enabled on the record view by default (and render an alert to the user when editing the metadata for an object).

screen shot 2017-01-31 at 3 18 48 pm

I would also want to pass this by @ggeisler, as he has done some thinking around the UX of this form.

cbeer:
I wonder if we need to better distinguish user-created custom fields from import-created ones?

tpendragon:
@jkeck When imported custom fields are created as vocab fields, they show up as facets by default because they don't have the properties in the blacklight_configuration to make them hidden. When you import a IIIF resource as a bunch of facetable fields, it's silly for the title facet to show up by default.

jkeck:
Ah, okay, I think I was worried about an issue that doesn't actually exist with this implementation.

After taking this for a spin, I'm 👍.  Would you mind giving it a quick rebase @tpendragon?

PR #1677, opened 2017-01-30
IIIF manifests for documents

### LINKED ISSUE #1641: Generate IIIF Manifests for Documents
Generate IIIF manifests to make content accessible to external IIIF users.

Note that this does not do anything to address our image performance problem.

jmartin-sul:
TODO:
* [x] failing test (was passing locally for me) needs fixing
* [x] commit messages need cleanup

closes #1641

### DIFF
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -109,6 +109,16 @@ module Spotlight
       end
     end
 
+    def manifest
+      _, document = fetch params[:id]
+
+      if document.uploaded_resource?
+        render json: Spotlight::IiifManifestPresenter.new(document, self).iiif_manifest_json
+      else
+        head :not_found
+      end
+    end
+
     protected
 
     # TODO: move this out of app/helpers/blacklight/catalog_helper_behavior.rb and into blacklight/catalog.rb
--- a/spec/controllers/spotlight/catalog_controller_spec.rb
+++ b/spec/controllers/spotlight/catalog_controller_spec.rb
@@ -111,6 +111,46 @@ describe Spotlight::CatalogController, type: :controller do
         expect(json['docs'].first['title']).to eq 'KAART der REYZE van drie Schepen naar het ZUYDLAND in de Jaaren 1721 en 1722'
       end
     end
+
+    describe 'GET manifest' do
+      context 'document is an uploaded resource' do
+        it 'returns the json manifest produced by Spotlight::IiifManifestPresenter, based on the retrieved document and the controller' do
+          uploaded_resource = FactoryGirl.create(:uploaded_resource)
+          compound_id = uploaded_resource.compound_id
+          slug = uploaded_resource.exhibit.slug
+
+          uploaded_resource.save_and_index
+
+          get :manifest, params: { exhibit_id: uploaded_resource.exhibit, id: compound_id }
+
+          expect(response).to be_successful
+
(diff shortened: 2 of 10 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -109,6 +109,16 @@ def make_public
       end
     end
 
+    def manifest
+      @response, @document = fetch params[:id]
```
cbeer: Is `@response` used? If now, can we just call that parameter `_`?

spec/controllers/spotlight/catalog_controller_spec.rb
```diff
@@ -111,6 +111,29 @@
...
+
+    describe 'GET manifest' do
+      context 'document is an uploaded resource' do
+        it 'returns the json manifest produced by Spotlight::IiifManifestPresenter, based on the retrieved document and the controller' do
+          allow_any_instance_of(::SolrDocument).to receive(:uploaded_resource?).and_return(true)
+          response_hash = { field: 'value' }
+          iiif_manifest_presenter = double(Spotlight::IiifManifestPresenter, iiif_manifest_json: response_hash.to_json)
+          expect(Spotlight::IiifManifestPresenter).to receive(:new).with(an_instance_of(::SolrDocument), controller).and_return(iiif_manifest_presenter)
+
+          get :manifest, params: { exhibit_id: exhibit, id: '1-1' }
```
jkeck: I was able to reproduce this locally.  I believe that in the context of this test (and on a fresh index) there is no resource w/ the id `1-1`.  You probably have one locally, which is why this passes for you.

What we probably want to do here is instead of using a hard-coded ID, we should use the `uploaded_resource` FactoryGirl factory to build a resource, and then pass that ID here to the `manifest` action.

spec/controllers/spotlight/catalog_controller_spec.rb
```diff
@@ -111,6 +111,29 @@
...
+          expect(response).to be_successful
+          expect(response.body).to eq(response_hash.to_json)
+        end
+      end
+      context 'document is not an uploaded resource' do
+        it 'returns a 404 when called on something other than an uploaded resource' do
+          allow_any_instance_of(::SolrDocument).to receive(:uploaded_resource?).and_return(false)
```
jkeck: Did you find that this was necessary?  I wonder if just passing an id to a solr document fixture that is not an uploaded resource should just do the right thing.

jmartin-sul:
@jkeck & @cbeer, i think this is ready for another look, assuming travis is happy

jkeck:
@jmartin-sul I pushed up a commit that makes sure to index the uploaded resource (and delete after itself).

Feel free to squash that into your commits.

app/controllers/spotlight/catalog_controller.rb
```diff
@@ -109,6 +109,16 @@ def make_public
       end
     end
 
+    def manifest
+      _, @document = fetch params[:id]
```
jkeck: This is a super minor nitpick, but since we're simply rending json inline below the `@document` ivar is pointless.  A local variable would suffice since we're not rending views.

If you're in there and feel like changing it, by all means, but I wouldn't hold this PR up for it.

jmartin-sul:
@jkeck:  made your suggested change in catalog_controller.rb and squashed in your indexing fix, repushed, tests passed.

PR #1683, opened 2017-02-01
Improve performance for catalog#edit page with lots of custom fields

### LINKED ISSUE #1682: Edit page takes a VERY long time to load
Not sure what causes it, but we're seeing > 45 seconds to load the edit page. Looking at the log it's probably something to do with generating a `blacklight_configuration` from the database. I'm running the profile now.

cbeer:
Fixes #1682

### DIFF
--- a/app/views/spotlight/catalog/_edit_default.html.erb
+++ b/app/views/spotlight/catalog/_edit_default.html.erb
@@ -19,7 +19,8 @@
         <% end %>
       <% end %>
 
-      <% current_exhibit.custom_fields.each do |field| %>
+      <%# Using `includes(:exhibit)` to ensure all fields are using the same exhibit object to take advantage of memoization %>
+      <% current_exhibit.custom_fields.includes(:exhibit).each do |field| %>
         <div class="form-group">
           <%= d.label field.field, field.label %>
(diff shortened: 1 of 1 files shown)


cbeer:
I'm not sure it's a great fix, but I think it'll work. Other things I thought of trying (like passing the configuration in as a parameter) seemed worse.

@tpendragon, can you give it a try?

```
irb(main):014:0> z = Spotlight::Exhibits.first.custom_fields.includes(:exhibit)
=> #<ActiveRecord::AssociationRelation [#<Spotlight::CustomField id: 1, exhibit_id: 1, slug: "x1", field: "x1", configuration: {}, created_at: "2017-02-01 18:30:54", updated_at: "2017-02-01 18:30:54", field_type: "text", readonly_field: false>, #<Spotlight::CustomField id: 2, exhibit_id: 1, slug: "x2", field: "x2", configuration: {}, created_at: "2017-02-01 18:31:04", updated_at: "2017-02-01 18:31:04", field_type: "text", readonly_field: false>]>
irb(main):015:0> z.first.exhibit.object_id
=> 70098016340840
irb(main):016:0> z.last.exhibit.object_id
=> 70098016340840
```

tpendragon:
Yeah, this works. It's about a second slower than not doing it at all, but significantly better than before.

PR #1686, opened 2017-02-02
Store IIIF manifest url, canvas id, and image id.

### LINKED ISSUE #1669: Allow for a IIIF Image URL to be reconstructed based on stored identifiers
#1498 adds the necessary columns to store these identifiers (but does not store them).

We should store the identifiers from the IIIF manifest and then add the tooling to update the image URL from the given manifest via the identifiers.  This may not necessarily be trigger-able (at this point), but we should be storing all these levels of identifiers before we ship #1498.

The use-case for this is in case a resource is renamed or remediated.  Storing the identifiers ensures that we can reconstruct the image url to the correct resource in case something goes wrong.

jkeck:
Add IiiFResourceResolver service to update the tilesource based on the stored identifiers.

Closes #1669


app/assets/javascripts/spotlight/multi_image_selector.js
```diff
@@ -103,7 +103,7 @@
     }
     function addThumbsToList(){
       $.each(image_versions, function(i){
-        var listItem = $('<li><a href="javascript:;"><img data-iiif-tilesource="' + image_versions[i]['tilesource'] + '" data-full-image="' + image_versions[i]['full'] +'" data-src="' + image_versions[i]['thumb'] +'" /></a></li>');
```
jkeck: It turns out that we didn't actually need this attribute.  We pass the tilesource directly when we parse the manifest.

app/services/spotlight/iiif_resource_resolver.rb
```diff
@@ -0,0 +1,73 @@
...
+    def manifest
+      @manifest ||= begin
+        JSON.parse(response)
+      rescue JSON::ParserError => e
+        Rails.logger.warn("#{self.class.name} failed to parse #{iiif_manifest_url} with: #{e}")
+        {}
+      end
+    end
+
+    class ManifestError < StandardError; end
```
cbeer: I wonder if this is actually a `RuntimeError` (subclass of `StandardError`)?

app/services/spotlight/iiif_resource_resolver.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    # Currently only supporting a single sequence
+    def sequence
+      Array.wrap(manifest['sequences']).first || {}
+    end
+
+    def response
+      @response ||= begin
+        Faraday.get(iiif_manifest_url).body
+      rescue Faraday::Error::ConnectionFailed => e
```
cbeer: Why just `ConnectionFailed`, instead of all `Faraday::Error`?

app/services/spotlight/iiif_resource_resolver.rb
jkeck: Totally fine with doing that.  I have been in the habit of subclassing `StandardError`, but it looks like `RuntimeError` makes more sense.

app/services/spotlight/iiif_resource_resolver.rb
jkeck: Copy-pasta.  Good call.

PR #1688, opened 2017-02-02
Reindex progress improvements

cbeer:


app/assets/javascripts/spotlight/reindex_monitor.js
```diff
@@ -48,6 +48,10 @@ Spotlight.onLoad(function() {
         .attr('aria-valuenow', percentage)
         .css('width', percentage + '%')
         .text(percentage + '%');
+
+      if (data.completed) {
```
jkeck: I _think_ `completed` is the number of items that have completed indexing.  At least that's what I gather from [this](https://github.com/projectblacklight/spotlight/pull/1688/files#diff-f2d64cc45e88a929fa17f3fb1f8de802R74).

Would `data.completed === data.total` work here instead? (or maybe I'm totally missing what this is intending to do)

cbeer: Right you are (maybe suggesting that field is poorly named..). I'll just push our `finished` attribute through the JSON and use that.

Issue #1685, opened 2017-02-02
Add default thumbnail for browse category


jkeck:
With the move to IIIF, it's a bit unsatisfying to have to write IIIF manifest parsing code just to fetch the first image of the first document in the exhibit (which may not exist).

If we could instead have a placeholder image, that could indidcate to the user that they should choose a good image for the browse category, exhibit etc.

ggeisler:
@jkeck Here's the placeholder image we discussed:

![placeholder_thumbnail](https://cloud.githubusercontent.com/assets/101482/22570991/9eb8679a-e952-11e6-9397-b0a36dad4daa.jpg)

PR #1692, opened 2017-02-03
Update sir-trevor widgets to use IIIF-backed images

### LINKED ISSUE #1673: Update image based SirTrevor widgets to use IIIF
Currently the sir-trevor widgets that use images from documents in the exhibit choose from 3 image derivatives stored as URLs in solr document fields (square, thumb, and full).

We should instead use the manifest stored in the document and IIIF identifiers (as well as cached image API references) similar to how Mastheads and Thumbnails are handled.

This will allow us to change image sizes for widgets to be more performant (as well as adds the potential for more responsiveness).

cbeer:
Fixes #1673 
Fixes https://github.com/sul-dlss/exhibits/issues/315

- [x] consider whether we should support non-IIIF backed image resources, especially for thumbnail-only applications
- [x] add non-IIIF alert (unnecessary. we're falling back to single-image objects with indexed data for non-IIIF stuff)
- [x] use shared IIIF manifest parsing
- [x] refactor away classes supporting non-IIIF-backed image versions
- [x] update uploaded resources to render using IIIF manifests (#1696)
- [x] #1685


app/assets/javascripts/spotlight/blocks/solr_documents_block.js
```diff
@@ -49,20 +49,55 @@ SirTrevor.Blocks.SolrDocuments = (function(){
...
+      $(this.inner).find('[name$="[iiif_canvas_id]"]').val(data.canvasId);
+    },
     afterPanelRender: function(data, panel) {
       var context = this;
+      var manifestUrl = data.iiif_manifest || data.iiif_manifest_url;
 
-      if (_.isUndefined(data['image_versions'])) {
-        $.getJSON(this.autocomplete_url().replace("%QUERY", "id:" + data.id), function(data) {
-          var doc = context.transform_autocomplete_results(data)[0];
+      $.ajax(manifestUrl).success(
```
cbeer: This manifest parsing is almost entirely copied from https://github.com/projectblacklight/spotlight/blob/master/app/assets/javascripts/spotlight/search_typeahead.js#L68

jkeck:
@cbeer: I'm getting a bunch of JS errors from the blocks on the edit page (and I believe that it's preventing the blocks icons from rendering)

sir-trevor-blocks

cbeer:
Huh, that's strange. Apparently `default` is a reserved word in js?

jkeck:
I think it is w/ es6 w/ the `export default class Foo`.

cbeer:
I'll take a look next week -- in checking this out, I think I found another bug too.

jkeck:
👍 totally.  All-in-all this is looking great so far though.  Thanks.

jkeck:
@cbeer I'm getting some of the hidden fields set to `undefined` when rendering the edit screen for a block that has already been created.  I believe `undefined` is being persisted as a value in the db and then it is trying to be used to generate image URLs.

undefined

cbeer:
Thanks, I think I've addressed that in 26b3e35.

PR #1697, opened 2017-02-03
Update IIIF migration script to reflect the current FeaturedImage…

### LINKED ISSUE #1672: Ensure a migration path exists for existing sites to update post IIIF
An initial pass at a migration path exists (as part of #1498, see `lib/migration/iiif.rb`)

This migration script will need some updating as what we store in the `FeaturedImage` class is different than when this was first written.  We'll want to ensure that this is a robust script that we can ship the IIIF work with (and reliably allow sites to update themselves).

jkeck:
…data model.

Closes #1672 (still will need SirTrevor migration path work, but that could be added to this class separately)

### DIFF
--- a/lib/migration/iiif.rb
+++ b/lib/migration/iiif.rb
@@ -17,13 +17,6 @@ module Migration
 
     attr_reader :hostname
 
-    def iiif_url(image)
-      riiif.image_url(image.id,
-                      region: coordinates(image),
-                      size: size(image),
-                      host: hostname)
-    end
-
     private
 
     def riiif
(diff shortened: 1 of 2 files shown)


lib/migration/iiif.rb
```diff
@@ -33,16 +26,32 @@ def riiif
...
+      image.save
+      # AR + STI seems to require that we re-query for this
+      # otherwise we get an association miss-match
+      reloaded_image = Spotlight::ExhibitThumbnail.find(image.id)
+      reloaded_image.image.store!(File.new(old_file))
+    end
+
     # Looks for a file at the old uploader location and copies it to a FeaturedImage
     def copy_contact_image_to_avatar(contact)
       filename = contact.read_attribute_before_type_cast('avatar')
```
jkeck: So, I believe the file path below is hard coded like this because we're trying to derive the CarrierWave storage directory from a class that does not exist anymore (an uploader was attached directly to the contact class, now that's handled through `FeaturedImage`).  Above, we can use the `store_dir` attribute safely because the classes are still present, but I'm not sure how we could derive the old classes w/o having some sort of hard coded path somewhere.

Totally up for suggestions on how to improve this though.

PR #1703, opened 2017-02-04
Add configuration and UI for selected themes for an exhibit

### LINKED ISSUE #389: Strategy for theming
How do we enable curators to select a theme from some set of visual themes, so not every Spotlight exhibit looks the same or requires hand-editing CSS to produce a different look?

cbeer:
Fixes #389

screen shot 2017-02-07 at 7 57 46 am


jkeck:
Looks good! We should document this somewhere, but I think we'll have a few things we need to document at the end of this work cycle, and we can add this there.

PR #1498, opened 2016-07-18
Add iiif-crop to spotlight

jcoyne:
Things preventing this from being merged:
- [x] https://github.com/sul-dlss/exhibits/issues/311
- [x] https://github.com/projectblacklight/spotlight/issues/1570
- [x] https://github.com/sul-dlss/iiif-cropper/issues/40
- [x] https://github.com/projectblacklight/spotlight/issues/1569
- [x] https://github.com/projectblacklight/spotlight/issues/1545
- [x] https://github.com/projectblacklight/spotlight/issues/1648

Nice to have:

- [ ] https://github.com/sul-dlss/exhibits/issues/309
- [x] https://github.com/projectblacklight/spotlight/issues/1535

PR #1706, opened 2017-02-04
Address regression (introduced in #1498) that added thumbnails to home pages

### LINKED ISSUE #1698: Thumbnail cropping for Home Page
There is an option to crop thumbnails for the home page of an exhibit.  This currently does not work in the latest release of exhibits (and we believe that it is not intended to).

Thumbnails are intended to be cropped for Feature Pages (as there is a SirTrevor widget that allows you to add the thumb + link to a Feature Page on another page).

With the IIIF cropping work, we set this up to work (since it was in the UI), but that may have been an oversight.

Should we remove the option to crop a thumbnail for the home page?

cbeer:
Fixes #1698

### DIFF
--- a/app/models/spotlight/home_page.rb
+++ b/app/models/spotlight/home_page.rb
@@ -8,8 +8,6 @@ module Spotlight
     before_save :publish
     before_create :default_content
 
-    accepts_nested_attributes_for :thumbnail, update_only: true, reject_if: proc { |attr| attr['iiif_tilesource'].blank? }
-
     class << self
       def default_title_text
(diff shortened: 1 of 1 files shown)


jkeck:
Should we remove the ability for users to attempt to crop a thumbnail for the home page?

exhibit-home

cbeer:
Ah -- looks like the regression was introduced even further back, huh?

https://github.com/projectblacklight/spotlight/commit/4daaffbaa7c99474d3fec2b5944160f3e41aa4fa

PR #1707, opened 2017-02-05
Allow browse categories to be updated

### LINKED ISSUE #1463: Enable curator to update search parameters of a browse category
Currently, if a curator wants to update the search underlying a browse category, they can't; they'd have to delete the browse category and create it again with the new search parameters. While this isn't a huge problem, if the curator has done the work to associate some text and a thumbnail (and possibly a masthead) image with the browse category, it's understandably annoying to have to redo that work.

Perhaps the modal that is displayed with the Save this search button could show the titles of existing browse categories and enable the curator to select an existing category, which would replace its underlying search parameters with the current ones?

Alternatively, we could reevaluate the whole workflow around saved search/browse categories, which is not particularly intuitive, and handle this issue as part of the redesigned workflow. That would need considerably more design (and likely development) work than the option above, however.

cbeer:
Fixes #1463 

screen shot 2017-02-05 at 7 43 59 am

### DIFF
--- a/app/views/catalog/_save_search.html.erb
+++ b/app/views/catalog/_save_search.html.erb
@@ -8,7 +8,9 @@
         <h4 class="modal-title" id="save-modal-label"><%= t(:'spotlight.saved_search.label') %></h4>
       </div>
       <div class="modal-body">
-        <%=f.text_field :title, label: t(:'spotlight.saved_search.title'), required: true %>
+        <%= f.text_field :title, label: t(:'spotlight.saved_search.title') %>
+        <%= label_tag :id, t(:'spotlight.saved_search.id'), class: 'control-label' %>
+        <%= select_tag :id, options_for_select(current_exhibit.searches.map { |s| [s.title, s.id] }), include_blank: true, class: 'form-control' %>
         <%= render_hash_as_hidden_fields(search_state.params_for_search.except(:qt, :page, :utf8)) %>
       </div>
       <div class="modal-footer">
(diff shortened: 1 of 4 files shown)


ggeisler:
Suggestion for form labels:

date__1981_-_an_exhibit_for_a_demo_-_online_exhibits_search_results

ggeisler:
Or, for the first label:

> Enter a title to create a new saved search

(parallels the wording of the second label a bit more)

cbeer:
👌 updated with your labels.

app/views/catalog/_save_search.html.erb
```diff
@@ -8,7 +8,8 @@
         <h4 class="modal-title" id="save-modal-label"><%= t(:'spotlight.saved_search.label') %></h4>
       </div>
       <div class="modal-body">
-        <%=f.text_field :title, label: t(:'spotlight.saved_search.title'), required: true %>
+        <%= f.text_field :title, label: t(:'spotlight.saved_search.title') %>
+        <%= select_tag :id, options_for_select(current_exhibit.searches.map { |s| [s.title, s.id] }), include_blank: true, label: t(:'spotlight.saved_search.id') %>
```
jkeck: I'm getting a bit of odd behavior with these field locally.

modal

markup

Perhaps `f.select` instead of `select_tag` could clear that up?

cbeer: Yeah, looks like I broke it updating the parameter name (although, apparently, not in a way the tests catch..). We need `select_tag` to ensure we get `params[:id]` instead of `params[:search][:id]`, but I guess that means we need to be explicit about everything.

PR #1710, opened 2017-02-06
RIIIF is now a runtime dependency

### LINKED ISSUE #1681: Could not find gem riiif . . . when installing from template.
If you try to install a new site from the template and have a clean slate of gems, you get the error:

> "Could not find gem 'riiif (~> 0.4.0)' in any of the gem sources listed in your Gemfile."
terminal errors

This causes spotlight:install:migrations, db:migrate and spotlight:initialize to fail during installation.

Steps to recreate:
  1. gem uninstall --all -x --force
  2. gem install rails --no-ri --no-rdoc && gem install bundle
  3. rails new test-spotlight -m https://raw.githubusercontent.com/projectblacklight/spotlight/master/template.rb

You should see the errors above and if you try to start the site, it reports that db:migrate was never run.
browser errors

This is easy enough to work around. Run `rake spotlight:install:migrations`, `rake db:migrate`, and `rake spotlight:initialize` manually and everything works. Just wanted to report it here.

cbeer:
Fixes #1681

### DIFF
--- a/blacklight-spotlight.gemspec
+++ b/blacklight-spotlight.gemspec
@@ -50,6 +50,8 @@ these collections.)
   s.add_dependency 'clipboard-rails', '~> 1.5'
   s.add_dependency 'almond-rails', '~> 0.0.1'
   s.add_dependency 'sprockets-es6'
+  s.add_dependency 'riiif', '~> 1.0'
+  s.add_dependency 'iiif_manifest'
 
   s.add_development_dependency 'sqlite3'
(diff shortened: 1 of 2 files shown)


jmartin-sul:
FWIW, i only added `iiif_manifest` as a development dependency because `riiif` was also a development dependency, and `iiif_manifest` was only getting used when `riiif` was getting used.  should `iiif_manifest` also be a runtime dependency?

PR #1716, opened 2017-02-06
Merge spotlight_resources_iiif gem into this project

cbeer:


jkeck:
Hmm, apparently it says that there is a conflict on rebase.  I kinda feel like we shouldn't squash this all into once commit.  What do you think @cbeer?

cbeer:
Hm. I'll rebase and try again.

cbeer:
Ugh -- git isn't tracking those early commits as against the subtree. I can try to redo the merge right, or we can just accept a squash merge. I'm slightly inclined to the latter, but doesn't really matter.

jkeck:
I'm fine with that.  Probably more trouble that it is worth.

PR #1719, opened 2017-02-07
Instantiate the multi image selector for masthead/thumbnail…

jkeck:
… typeaheads on page load.

### DIFF
--- a/app/assets/javascripts/spotlight/blocks/solr_documents_block.js
+++ b/app/assets/javascripts/spotlight/blocks/solr_documents_block.js
@@ -94,7 +94,11 @@ SirTrevor.Blocks.SolrDocuments = (function(){
           var iiifManifest = new Iiif(manifestUrl, manifest);
 
           var thumbs = iiifManifest.imagesArray();
-          context.setIiifFields(panel, thumbs[0], !!data.iiif_manifest_url);
+
+          if (!data.iiif_image_id) {
+            context.setIiifFields(panel, thumbs[0], !!data.iiif_manifest_url);
+          }
+
 
           if(thumbs.length > 1) {
             panel.multiImageSelector(thumbs, function(selectorImage) {
--- a/app/assets/javascripts/spotlight/blocks/solr_documents_block.js
+++ b/app/assets/javascripts/spotlight/blocks/solr_documents_block.js
@@ -99,7 +99,7 @@ SirTrevor.Blocks.SolrDocuments = (function(){
           if(thumbs.length > 1) {
             panel.multiImageSelector(thumbs, function(selectorImage) {
               context.setIiifFields(panel, selectorImage, false);
-            });
+            }, data.iiif_image_id);
           }
         }
       );
--- a/app/assets/javascripts/spotlight/multi_image_selector.js
+++ b/app/assets/javascripts/spotlight/multi_image_selector.js
@@ -1,7 +1,7 @@
 // Module to add multi-image selector to widget panels
 
 (function(){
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/spotlight/multi_image_selector.js
```diff
@@ -1,7 +1,7 @@
 // Module to add multi-image selector to widget panels
 
 (function(){
-  $.fn.multiImageSelector = function(image_versions, clickCallback) {
+  $.fn.multiImageSelector = function(image_versions, clickCallback, activeTilesource) {
```
jkeck: Not sure if this param should come after the callback.  We don't use this necessarily in the other multi-image contexts.

jkeck:
I believe that the SirTrevor feature works based on a hidden field that we don't use for mastheads/thumbs.  I could try to unify these approaches, but it seems like these could be unified later if/when we try to unify the masthead/thumbnail + SirTrevor auto-complete features.

cbeer:
> I believe that the SirTrevor feature works based on a hidden field

Yeah, it's using https://github.com/projectblacklight/spotlight/blob/master/app/assets/javascripts/spotlight/multi_image_selector.js#L29

I'm more than a little worried about two competing approaches to selecting the active thumbnail floating around, particularly when they're backed by different data (the ST widgets are currently using `image-id`(annotation `@id`), while this uses the IIIF tilesource (`info.json` URL)).

That said, I'm 👍 about being explicit about the current selection, like you're doing here. Maybe this whole class needs a good refactor?

jkeck:
Hadn't noticed those changes from when I initially did this work (pre-SirTrevor changes).   I will try to make those changes this morning.

spec/features/autocomplete_typeahead_spec.rb
```diff
@@ -27,6 +27,30 @@
         expect(featured_image.iiif_image_id).to eq 'https://purl.stanford.edu/gk446cj2442/iiif/annotation/gk446cj2442_1'
         expect(featured_image.iiif_tilesource).to eq 'https://stacks.stanford.edu/image/iiif/gk446cj2442%2Fgk446cj2442_05_0001/info.json'
       end
+
+      it 'instantiates the multi-image selector when an multi-image item is chosen in the typeahead (and again on edit)' do
+        # Poltergeist doesn't play nice with croppers behind tabs. This is not the case in real browsers
+        allow(Spotlight::Engine.config).to receive(:exhibit_themes).and_return(['default'])
```
jkeck: This feels unsatisfactory, but the only other way would be to skip the test entirely.  Otherwise a javascript error is thrown (from somewhere down in the leaflet-iiif code) that is not reproducible in a real browser.

cbeer: Huh, very curious.

app/assets/javascripts/spotlight/blocks/solr_documents_block.js
```diff
@@ -94,12 +94,16 @@ SirTrevor.Blocks.SolrDocuments = (function(){
           var iiifManifest = new Iiif(manifestUrl, manifest);
 
           var thumbs = iiifManifest.imagesArray();
-          context.setIiifFields(panel, thumbs[0], !!data.iiif_manifest_url);
+
+          if (!data.iiif_image_id) {
+            context.setIiifFields(panel, thumbs[0], !!data.iiif_manifest_url);
```
cbeer: I think we need to do this even if there's only 1 thumbnail, no?

PR #1730, opened 2017-02-08
allow deleting contact email addresses

### LINKED ISSUE #669: Delete Contact email on Settings page
We currently allow adding new contact email addresses on the Dashboard > Settings page, but there is no way for the Admin to remove a contact email address.

Need a way to delete an existing contact email address.

jmartin-sul:
successful deletion example (including manual page refresh to show contact email was actually deleted):
669-delete-contact-email-success

error deleting (contact address already deleted, e.g. in another page load in a different tab):
669-delete-contact-email-error

closes #669

### DIFF
--- a/app/controllers/spotlight/contact_email_controller.rb
+++ b/app/controllers/spotlight/contact_email_controller.rb
@@ -1,0 +1,22 @@
+module Spotlight
+  ##
+  # CRUD actions for exhibit contact emails
+  class ContactEmailController < Spotlight::ApplicationController
+    rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
+
+    before_action :authenticate_user!
+    load_and_authorize_resource :exhibit, class: 'Spotlight::Exhibit'
+    load_and_authorize_resource through: :exhibit
+
+    def destroy
+      @contact_email.destroy
+      render json: { success: true, error: nil }
+    end
+
+    private
+
+    def record_not_found(_error)
+      render json: { success: false, error: 'Not Found' }, status: :not_found
+    end
+  end
+end
--- a/app/assets/javascripts/spotlight/exhibits.js
+++ b/app/assets/javascripts/spotlight/exhibits.js
@@ -26,6 +26,7 @@ Spotlight.onLoad(function() {
     });
 
     input_container.find('.first-row-only').remove();
+    input_container.find('.confirmation-status').remove();
 
     // bootstrap does not render input-groups with only one value in them correctly.
     input_container.find('.input-group input:only-child').closest('.input-group').removeClass('input-group');
--- a/app/assets/javascripts/spotlight/exhibits.js
+++ b/app/assets/javascripts/spotlight/exhibits.js
@@ -25,7 +25,6 @@ Spotlight.onLoad(function() {
       $(this).attr('name', $(this).attr('name').replace('0', contacts.length));
     });
 
-    input_container.find('.first-row-only').remove();
     input_container.find('.confirmation-status').remove();
 
     // bootstrap does not render input-groups with only one value in them correctly.
--- a/app/assets/javascripts/spotlight/exhibits.js
+++ b/app/assets/javascripts/spotlight/exhibits.js
@@ -16,21 +16,32 @@ Spotlight.onLoad(function() {
   $("#another-email").on("click", function() {
     var container = $(this).closest('.form-group');
     var contacts = container.find('.contact');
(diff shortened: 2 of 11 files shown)


jmartin-sul:
finally got the tests working and cleaned up the commit history on friday.

just added a couple short screencaps demonstrating the basic behavior.  @ggeisler, would you mind taking a look at the interaction and giving some feedback?  specifically:
* i moved the `+` button for adding addresses so that it's its own row under the address list.  leaving it in place on the first row would've made UI implementation more complex when deleting the first address in the list.
* there's a confirmation, but deletion happens via ajax request and doesn't require hitting the `Save changes` button.
* delete button is a basic red x (`glyphicon-remove` from bootstrap classed with `alert-danger`).
* if there's an error deleting (e.g. the address was already deleted), the error message displays in red to the right of the delete button.

jmartin-sul:
rebased from current master and pushed again

ggeisler:
@jmartin-sul Overall this looks good. I have a couple of suggestions for you to consider, which have to do with visual design, not the interaction, which seems good to me.

In the mockups below I replaced the `glyphicon-remove` icon with a button. The icon you used is nice in that it fits well with the confirmation icon, but I wonder if it might not be better to be more explicit here, rather than relying just on the icon? It does make the UI a little less subtle, but given only exhibit admins see this page that doesn't bother me too much.

configuration_-_general___test_exhibit_-_blacklight

Alternatively, we could just add a label next to the remove icon: "Delete contact." This is closer to what you have, although I kind of like the button better, since this is an action, while the confirmation label below ii is just a message, but they feel the same with the non-button approach.

configuration_-_general___test_exhibit_-_blacklight 2

In the first mockup I also replaced the `+` button for adding new contacts with a button that is more explicit about the action it provides, and is just a standard size button (i.e. doesn't fill the column).

For styling the buttons, I used `btn-xs btn-danger` for the delete contact and `btn-sm btn-info` for the add new contact button.

One issue with switching to the button for delete contact is it might complicate how you present the error message, since there is less space to present the message inline. Maybe it will still look okay presented below the button? Or maybe try placing it below the email address input in the column to the left?

jmartin-sul:
incorporated feedback from @ggeisler, re-did screenshots, should be ready for review again

app/controllers/spotlight/contact_email_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Spotlight
+  ##
+  # CRUD actions for exhibit contact emails
+  class ContactEmailController < Spotlight::ApplicationController
+    rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
```
jcoyne: This is something Rails does by default. In `config.action_dispatch.rescue_responses`. You don't need to provide this here.

app/controllers/spotlight/contact_email_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    before_action :authenticate_user!
+    load_and_authorize_resource :exhibit, class: 'Spotlight::Exhibit'
+    load_and_authorize_resource through: :exhibit
```
jcoyne: Can you explain this to me? We're loading an exhibit through an exhibit?

app/controllers/spotlight/contact_email_controller.rb
jcoyne: Hmm, I see what you're doing here. This is for a custom JSON response. Rails doesn't seem to have a good way to customize that. 👍

app/controllers/spotlight/contact_email_controller.rb
mejackreed: My understanding of this is the following:

`load_and_authorize_resource through: :exhibit` loads and authorizes `@contact_email` appropriately using `exhibit`

`load_and_authorize_resource :exhibit, class: 'Spotlight::Exhibit'` loads and authorizes the `@exhibit` in the case where the `@contact_email` is unavailable.

Is there a better way to do this?

I see a similar pattern here: https://github.com/projectblacklight/spotlight/blob/master/app/controllers/spotlight/custom_fields_controller.rb#L7-L8

jcoyne: Okay, I'm just a little under-caffeinated this morning. This makes perfect sense.

mejackreed: No problemo.. It made me do a double take. And frankly CanCan always feels a bit like magic.

PR #1732, opened 2017-02-08
A few IIIF Importer improvements

### LINKED ISSUE #1731: IIIF harvester explicitly uses the configured title field from the index
We should have some way to index the title into multiple fields to support multiple views.

jkeck:
Closes #1731 

* Fix thumbnail url for uploaded resources (slightly unrelated)
* Allow for string, array, or hash style labels.
* Allow for IIIF imported items to configure multiple title fields

### DIFF
--- a/app/models/spotlight/resources/iiif_manifest.rb
+++ b/app/models/spotlight/resources/iiif_manifest.rb
@@ -64,8 +64,9 @@ module Spotlight
 
       def add_label
         return unless title_fields.present? && manifest.try(:label)
-        title_fields.each do |field|
-          solr_hash[field] = manifest.label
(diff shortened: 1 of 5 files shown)


app/models/spotlight/resources/iiif_manifest.rb
```diff
@@ -3,6 +3,8 @@ module Resources
     ##
     # A PORO to construct a solr hash for a given IiifManifest
     class IiifManifest
+      DEFAULT_JSON_LD_LANGUAGE = 'en'.freeze
```
cbeer: Should this be an engine configuration?

PR #1738, opened 2017-02-09
Guard against attempting to copy files that do not exist.

jkeck:


jkeck:
I wasn't able to find a good way to test a case where we generated a `FeaturedImage` with mounted uploader and the file did not exist on disk.

It seems like any stubs I set end up needing to be on the thing that I need to make assertions against.

Totally willing to look at this tomorrow a bit harder (or any advice on how to test this reliably).

PR #1753, opened 2017-02-14
Exhibit delete

wwelling:
Using @escowles strategy for cleaning Solr. https://github.com/pulibrary/pomegranate/pull/204

### DIFF
--- a/app/models/spotlight/resource.rb
+++ b/app/models/spotlight/resource.rb
@@ -118,10 +118,6 @@ module Spotlight
         solr_document_sidecars.size > 1
       end
 
-      def solr_document_sidecars
-        Spotlight::SolrDocumentSidecar.where(document_id: document_id, document_type: SolrDocument)
-      end
-
       def document_id
         document_builder.documents_to_index.to_a.map { |y| y[:id] }
       end
--- a/app/models/spotlight/resource.rb
+++ b/app/models/spotlight/resource.rb
@@ -20,6 +20,8 @@ module Spotlight
     after_index :commit
     after_index :touch_exhibit!
 
+    after_destroy :cleanup_solr_record, :cleanup_featured_image
+
     ##
     # Persist the record to the database, and trigger a reindex to solr
     #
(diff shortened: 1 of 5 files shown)


wwelling:
If this is an acceptable way to cleanup resources after deletion of an exhibit, I will need assistance in writing the specs to maintain the test coverage.

app/models/spotlight/resource.rb
```diff
@@ -100,6 +102,19 @@ def touch_exhibit!
       def write?
         Spotlight::Engine.config.writable_index
       end
+
+      def cleanup_solr_record
+        blacklight_solr.delete_by_id(document_ids, params: { softCommit: true })
+      end
+
+      def document_ids
+        document_builder.documents_to_index.to_a.map { |y| y[:id] }
```
cbeer: This will not work for a couple institutions:

- Stanford, for one, expects and allows solr documents to appear in multiple exhibits, so blindly deleting them will impact other exhibits.

- The Danes (?) have a read-only index (using ` Spotlight::Engine.config.writable_index = false`), so also wouldn't want the index blindly updated.


I'd suggest a couple alternative approaches, including:

- putting this behind a configuration setting, so institutions aren't obligated to use it
- checking for `Spotlight::SolrDocumentSidecar` instances with that document id to determine whether a document appears in multiple exhibits
- introducing a new field to the index that stores either the list of exhibits an item appears in, or a list of resources that "own" the document (perhaps `Spotlight::Engine.config.resource_global_id_field` is a relevant starting point?)

wwelling: Thanks for the suggestions. I did not know Solr documents could appear in multiple exhibits. But, that makes since to allow them to. I assume removal of the orphaned resource is a must when an exhibit is deleted. The Solr index would then have to be amended appropriately. I will look into making the Solr cleanup configurable and how to cleanup Solr entries without removing documents used within other exhibits. What do you think about the upload image cleanup upon destroy of the resource?

wwelling: Maybe your second suggestion is the better approach as to not making to many changes.

wwelling:
Is this change meeting your second alternative approach? "checking for Spotlight::SolrDocumentSidecar instances with that document id to determine whether a document appears in multiple exhibits"

nicolasfranck:
What happened to this pull request? For now it does not seem possible to remove an exhibition (or even a resource, or item) and its attached uploaded images.

PR #1762, opened 2017-03-02
Add limit to facets, fix custom field facet mapping

tpendragon:
It turns out the key is important for the "more" button to work in Blacklight.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -79,6 +79,10 @@ RSpec/MultipleExpectations:
   Exclude:
     - 'spec/features/**/*'
 
+RSpec/BeforeAfterAll:
+  Exclude:
+    - 'spec/models/spotlight/reindexing_log_entry_spec.rb'
+
 Bundler/DuplicatedGem:
   Enabled: false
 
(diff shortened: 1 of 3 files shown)


cbeer:
> It turns out the key is important for the "more" button to work in Blacklight.

Can you say more? That seems like a blacklight bug, off-hand.

tpendragon:
> Can you say more? That seems like a blacklight bug, off-hand.

When the configuration key in blacklight wasn't the solr key of the item, it would throw an error trying to run a query for the facet action, because the expanded key wouldn't exist. This patch actually fixed my facet preview counts in the "search" configuration panel, too.

cbeer:
Ok, that sounds like a definite blacklight bug (and then a related spotlight one?). We should consistently use the configured key in anything blacklight exposes to end users, and consistently use the field name for talking to solr.

tpendragon:
Does that mean you won't merge this PR?

cbeer:
Give me a chance to look upstream first. Being able to use a different url parameter is a feature we intend to support in Blacklight, so I'd rather not patch around it here if there's an easy enough fix in Blacklight.

PR #1767, opened 2017-03-04
Update Carrierwave's extension whitelist method name.

jkeck:
The method name was changed in 1.0.0


app/assets/javascripts/spotlight/search_typeahead.js
```diff
@@ -1,5 +1,5 @@
 //= require typeahead.bundle.min.js
-//= require handlebars-v1.3.0.js
+// require handlebars-v1.3.0.js
```
cbeer: Does this change belong here?

jkeck: Ah, sorry, you caught me working on two things at once.  Good catch.

PR #1781, opened 2017-03-22
Adds the ability to configure which sirtrevor widgets are shown

### LINKED ISSUE #1804: How to add new Sir Trevor block to widget menu
Hello,

I'm trying to add a "Subheading" block to the Sir Trevor widget options, which is the same as heading, except that it uses h3s instead of h2s. I'm following the instructions on [the wiki](https://github.com/projectblacklight/spotlight/wiki/Page-widgets).

I can see it's being loaded into the SirTrevor.Blocks, it's just not showing up in the widget icons when someone makes a new block.
blocks

I've chased the issue down to line 21 in javascripts/spotlight/sir_trevor/block_controls.js

```javascript
 function generateBlocksHTML(Blocks, availableTypes) {
    var groups = {};
    for(var i in availableTypes) {
      //. . . . etc
```
It seems my 'Subheading' block is not in the availableTypes array, so it's not being loaded. 

So my question is, how can I get it into availableTypes? Have I set up my view/model/js wrong? I'll attach my files at the bottom.

Thanks for any help,

-- Travis

**app/assets/javascripts/spotlight/blocks/subheading_block.js**
```js
SirTrevor.Blocks.Subheading = (function() {

  return SirTrevor.Block.extend({
    type: "subheading",

    title: function() { return "hi!" },

    icon_name: "tweet", //Just for testing, will make an icon later

    formable: true,

    editorHTML: function() {
      return _.template(this.template, this)(this);
    },

    template: '<input type="text" name="title" />'

  });
})();
```

**app/views/spotlight/sir_trevor/blocks/_subheading_block.html.erb**
```ruby
<h3><%= subheading_block.title %></h3>
```

**app/models/sir_trevor_rails/blocks/subheading_block.rb**
```ruby
module SirTrevorRails
  module Blocks
    ##
    # Display a subheading
    class SubheadingBlock < SirTrevorRails::Block
      include Displayable
      include Textable

    end
  end
end
```

mejackreed:
Closes #1804

### DIFF
--- a/app/assets/javascripts/spotlight/pages.js.erb
+++ b/app/assets/javascripts/spotlight/pages.js.erb
@@ -22,8 +22,7 @@ Spotlight.onLoad(function(){
   if (instance.length) {
     var editor = new SirTrevor.Editor({
       el: instance[0],
-      blockTypes:["Heading", "Text", "List","Quote","Iframe", "Video", "Oembed", "Rule", "UploadedItems",
-                 "Browse", "FeaturedPages", "SolrDocuments",  "SolrDocumentsCarousel", "SolrDocumentsEmbed", "SolrDocumentsFeatures", "SolrDocumentsGrid", "SearchResults"],
+      blockTypes: instance.data('blockTypes'),
       defaultType:["Text"],
       onEditorRender: function() {
         serializeObservedForms(observedForms());
--- a/app/views/spotlight/pages/_form.html.erb
+++ b/app/views/spotlight/pages/_form.html.erb
@@ -51,7 +51,7 @@
         </div>
         <div class="form-group">
           <%= f.label :content, class: 'sr-only' %>
-          <%= f.text_area_without_bootstrap :content, value: { data: f.object.content.as_json }.to_json, class: 'js-st-instance' %>
+          <%= f.text_area_without_bootstrap :content, value: { data: f.object.content.as_json }.to_json, class: 'js-st-instance', data: { blockTypes: Spotlight::Engine.config.sir_trevor_widgets } %>
         </div>
       </div>
 
--- a/lib/generators/spotlight/templates/config/initializers/spotlight_initializer.rb
+++ b/lib/generators/spotlight/templates/config/initializers/spotlight_initializer.rb
@@ -55,3 +55,10 @@
 # Spotlight::Engine.config.ga_email = nil
 # Spotlight::Engine.config.ga_analytics_options = {}
 # Spotlight::Engine.config.ga_page_analytics_options = config.ga_analytics_options.merge(limit: 5)
+
+# ==> Sir Trevor Widget Configuration
+# Spotlight::Engine.config.sir_trevor_widgets = %w(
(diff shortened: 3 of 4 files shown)


app/assets/javascripts/spotlight/pages.js.erb
```diff
@@ -22,8 +22,7 @@ Spotlight.onLoad(function(){
   if (instance.length) {
     var editor = new SirTrevor.Editor({
       el: instance[0],
-      blockTypes:["Heading", "Text", "List","Quote","Iframe", "Video", "Oembed", "Rule", "UploadedItems",
-                 "Browse", "FeaturedPages", "SolrDocuments",  "SolrDocumentsCarousel", "SolrDocumentsEmbed", "SolrDocumentsFeatures", "SolrDocumentsGrid", "SearchResults"],
+      blockTypes: <%= Spotlight::Engine.config.sir_trevor_widgets %>,
```
cbeer: I thought of another approach for this too... we could add this list as a `data-` attribute on the `js-st-instance`. This would have the benefit of not depending on ERB parsing, and open it up to things like exhibit-specific customization (or run-time things like... if there are no solr documents, you don't get those widgets).

:man_shrugging:

mejackreed: I'm not sure I fully grock what you are proposing. Would this allow `blockTypes` to be just grabbed from the instantiating block type based on the data-attribute?

cbeer: I'm suggesting that we do something in this partial:

https://github.com/projectblacklight/spotlight/blob/dd9805cb240cbc6425c47ea69619b6fd1ed6aa2f/app/views/spotlight/pages/_form.html.erb#L54

Maybe starting with something like:

```
<%= f.text_area_without_bootstrap :content, value: { data: f.object.content.as_json }.to_json, class: 'js-st-instance', data: { blockTypes: Spotlight::Engine.config.sir_trevor_widgets } %>
```

and in the javascript, pull the data from `instance.data('blockTypes')`.

app/assets/javascripts/spotlight/pages.js.erb
```diff
@@ -22,8 +22,7 @@ Spotlight.onLoad(function(){
   if (instance.length) {
     var editor = new SirTrevor.Editor({
       el: instance[0],
-      blockTypes:["Heading", "Text", "List","Quote","Iframe", "Video", "Oembed", "Rule", "UploadedItems",
-                 "Browse", "FeaturedPages", "SolrDocuments",  "SolrDocumentsCarousel", "SolrDocumentsEmbed", "SolrDocumentsFeatures", "SolrDocumentsGrid", "SearchResults"],
+      blockTypes: instance.data['blockTypes'],
```
cbeer: Is this hash access conventional? I usually see this as`.data('blockTypes')`.

mejackreed: 👍

PR #1789, opened 2017-04-04
update .rubocop_todo.yml

jmartin-sul:
run rubocop --auto-gen-config for rubocop 0.48.1


jkeck:
Thanks @jmartin-sul!

@cbeer do you think we ought to auto-fix any of these?  Just wondering how much of a PITA that would be for other outstanding PRs.  It seems like a lot of this should be fixable (and I'm happy to do that in a commit to this PR)

cbeer:
No opinion. I'm happy to rebase my PRs with or without autofixing these.

PR #1823, opened 2017-08-18
Fix label for rake task

jcoyne:
It wasn't creating an exhibit


cbeer:
I wonder if there's any value in `spotlight:initialize` over `spotlight:admin`?

PR #1839, opened 2017-10-06
Merge upstream

wwelling:


jcoyne:
Was this intentional? I don't think we can accept Texas A&M branding here.

wwelling:
No this was not intentional. Sorry.

PR #1849, opened 2017-10-23
make sure the data attribute is in form of data-block-types for Sir T…

mejackreed:
…revor widgets

This bug was introduced in PR #1781. Essentially, previously, the JavaScript method `instance.data('blockTypes')` was not finding anything because the data-attribute was in the form of `data-blocktypes`. This caused all of the sir trevor registered widgets to be loaded unexpectedly allowing exhibit curators to possibly select unsupported widgets.

### DIFF
--- a/spec/views/spotlight/pages/_form.html.erb_spec.rb
+++ b/spec/views/spotlight/pages/_form.html.erb_spec.rb
@@ -1,0 +1,18 @@
+describe 'spotlight/pages/edit', type: :view do
+  let(:exhibit) { stub_model(Spotlight::Exhibit) }
+  let(:page) { stub_model(Spotlight::FeaturePage, exhibit: exhibit) }
+  let :blacklight_config do
+    Blacklight::Configuration.new
+  end
+  before do
+    assign(:page, page)
+    allow(view).to receive_messages(blacklight_config: blacklight_config,
+                                    available_index_fields: [],
+                                    available_view_fields: [],
+                                    featured_images_path: '/foo')
+  end
+  it 'contains data-block-types attribute needed for SirTrevor instantiation' do
+    render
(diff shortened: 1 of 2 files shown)


spec/views/spotlight/pages/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+  before do
+    assign(:page, page)
+    # allow(view).to receive_messages(blacklight_config: blacklight_config)
+    allow(view).to receive_messages(blacklight_config: blacklight_config,
+                                    available_index_fields: [],
+                                    available_view_fields: [],
+                                    featured_images_path: '/foo')
+  end
+  it '' do
```
camillevilla: hi @mejackreed! Can you add a test description here?

spec/views/spotlight/pages/_form.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
+describe 'spotlight/pages/edit', type: :view do
+  let(:exhibit) { stub_model(Spotlight::Exhibit) }
+  let(:page) { stub_model(Spotlight::FeaturePage, exhibit: exhibit) }
+  let :blacklight_config do
+    Blacklight::Configuration.new
+  end
+  before do
+    assign(:page, page)
+    # allow(view).to receive_messages(blacklight_config: blacklight_config)
```
camillevilla: I think this comment snuck in 👀

spec/views/spotlight/pages/_form.html.erb_spec.rb
mejackreed: Thanks nice catch. I was too eager :) Added.

Issue #1782, opened 2017-03-23
Bug updating browse category that exists in SirTrevor blocks


jkeck:
If you update a browse category in a way that triggers a slug update, any page that includes that browse category in the browse category block will throw an error on the edit page.

This is because the `as_json` method calls a `find` which can trigger an `ActiveRecord::NotFound` error.

cbeer:
I've been thinking about how to solve this without too much trouble. I'm beginning to suspect we may want to introduce a new join table to link pages with embedded resources (and use the primary key of that link in the blocks themselves).

caaster:
I think this problem can also occur if a browse category is renamed (not just deleted).

jkeck:
@caaster yes I agree that this would happen in either of those cases.

jkeck:
I agree that ultimately https://github.com/projectblacklight/spotlight/issues/1782#issuecomment-288609585 is going to be the approach that we should take (or should have taken initially) however it will require a non-trivial amount of work to accomplish and will require data migrations to be written to migrate all existing data to an appropriate join table.

If we want a stop-gap to deal w/ the issue in the interim I can think of two ways that we might approach this which will have varying user experience.  As it stands the rendered view of the page works out fine and simply does not show the browse category that has changed.  On the edit screen we can potentially do one of the following.

1. Show the browse category whose slug has changed with some sort of styling and/or note indicating that the browse category will not render and most likely has had its slug changed.
2. We can simply omit the browse category from displaying on the edit screen  all-together.

If we want to go with the approach in the comment above that'll take a bit more time, and wanted to stop to check-in before making a decision to go one way or the other.

ggeisler:
I also agree keeping track of the embedded categories by PK rather than slug is way we want to go, long-term, since eliminates the problem with no unexpected UX issues.

If we don't think we can work on that ideal approach in the near future, I think @jkeck's second option sounds best to me (if I understand things correctly). The downside of the curator having to re-select a category whose slug changed, while not ideal and possibly slightly confusing (especially if someone else was the one to change the browse category slug), seems tolerable and cleaner than the first option would be.

jkeck:
I'm mentioning something that can be implemented nearly immediately and will solve the issue of certain pages being completely un-editable in the immediate term while working on a more permanent solution in the long term.

[...]

PR #1850, opened 2017-10-24
Cast empty items to a Hash for empty browse categories & feature pages

mejackreed:
We noticed this issue downstream in Exhibits, that when enabling "browse categories" and then creating an empty "BrowseBlock" (no browse category added), an "undefined method each for nil" error could be raised. This PR ensures that `items` retrieved from `data` are cast to a `Hash` with `.each` available.

See: https://github.com/sul-dlss/exhibits/issues/789

### DIFF
--- a/app/models/sir_trevor_rails/blocks/browse_block.rb
+++ b/app/models/sir_trevor_rails/blocks/browse_block.rb
@@ -32,7 +32,9 @@ module SirTrevorRails
       def as_json
         result = super
 
-        result[:data][:item].each do |_k, v|
+        result[:data][:item] ||= {}
+
+        result[:data][:item].each_value do |v|
           v['thumbnail_image_url'] = parent.exhibit.searches.find(v['id']).thumbnail_image_url
         end
 
--- a/app/models/sir_trevor_rails/blocks/featured_pages_block.rb
+++ b/app/models/sir_trevor_rails/blocks/featured_pages_block.rb
@@ -22,7 +22,9 @@ module SirTrevorRails
       def as_json
         result = super
 
-        result[:data][:item].each do |_k, v|
+        result[:data][:item] ||= {}
+
+        result[:data][:item].each_value do |v|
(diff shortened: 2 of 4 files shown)


app/models/sir_trevor_rails/blocks/browse_block.rb
```diff
@@ -32,7 +32,7 @@ def display_item_counts?
       def as_json
         result = super
 
-        result[:data][:item].each do |_k, v|
+        result[:data][:item].to_h.each_value do |_k, v|
```
cbeer: You can lose the `_k` now, right?

app/models/sir_trevor_rails/blocks/browse_block.rb
```diff
@@ -32,7 +32,7 @@ def display_item_counts?
       def as_json
         result = super
 
-        result[:data][:item].each do |_k, v|
```
cbeer: I wonder if we should pull `result[:data][:item]` up to a method on the parent class or mixin that returns a "safe" value.

cbeer: Or, an `items` accessor that returns an empty array?

jkeck: I think this also could tie into #1782 (at least a stop-gap fix for #1782)

mejackreed: Yep likely. But I haven't tested yet on an unavailable browse category, just ones that are missing data.

app/models/sir_trevor_rails/blocks/browse_block.rb
```diff
@@ -32,7 +32,7 @@ def display_item_counts?
...
+        result[:data][:item].to_h.each_value do |v|
```
cbeer: I wonder if we should revisit how this blocks get serialized to json. Maybe we're missing a useful class or other way to inject custom serializations (rather than munging it after it's been serialized)

PR #1854, opened 2017-10-30
Store a copy of the original field configuration before applying spotlight-specific changes

cbeer:
Use this, instead of the upstream_if hack, to evaluate whether to offer fields for admin users or not


mejackreed:
Is there a need to deprecate `upstream_if` or was that just something internal?

cbeer:
I think it was internal-only, so no deprecation needed.

PR #1856, opened 2017-11-01
Only handle browse block items whose block id/slug exists.

jkeck:
This is a temporary fix to stop errors being thrown on edit pages.  A more permanent solution should be implemented in the future.

Connects to #1782

### DIFF
--- a/app/models/sir_trevor_rails/blocks/browse_block.rb
+++ b/app/models/sir_trevor_rails/blocks/browse_block.rb
@@ -34,6 +34,10 @@ module SirTrevorRails
 
         result[:data][:item] ||= {}
 
+        # TODO: This is a temporary fix that simply removes any item if the search identifier does not exist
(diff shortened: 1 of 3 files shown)


app/models/sir_trevor_rails/blocks/browse_block.rb
```diff
@@ -34,6 +34,10 @@ def as_json
 
         result[:data][:item] ||= {}
 
+        # TODO: This is a temporary fix that simply removes any item if the search identifier does not exist
+        #       We should have a more permanent solution that will allow browse blocks to be updated without erroring
+        result[:data][:item].select! { |_, v| parent.exhibit.searches.exists?(v['id']) }
```
jkeck: I can think of a few other ways to accomplish this, but this one seemed succinct (particularly since we'll eventually want to remove this line all together).

Happy to discuss alternate approaches.

jkeck:
Ah, I hand't noticed that other test.  👍

PR #1868, opened 2017-11-30
Deliver user invitation after the role has been created…

jkeck:
…(since role information is used in the email content).

### DIFF
--- a/app/controllers/spotlight/admin_users_controller.rb
+++ b/app/controllers/spotlight/admin_users_controller.rb
@@ -10,6 +10,7 @@ module Spotlight
 
     def create
       if update_roles
+        Spotlight::InviteUsersService.call(resource: @site)
         flash[:notice] = t('spotlight.admin_users.create.success')
       else
         flash[:error] = t('spotlight.admin_users.create.error')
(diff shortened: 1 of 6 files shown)


app/models/spotlight/role.rb
```diff
@@ -8,6 +8,15 @@ class Role < ActiveRecord::Base
     validates :role, inclusion: { in: ROLES }
     validate :user_must_be_unique, if: :user
 
+    after_create do
+      # Deliver an invitation to the user if their account
+      # was created by an invite and have not accepted it yet.
+      # This is being done after create because the user and role
+      # both need to be persisted given that the role information
+      # is used in the content of the invitation email.
+      user.deliver_invitation if user.created_by_invite? && user.invitation_accepted_at.blank?
```
jkeck: This doesn't really feel like this right place for this to be to me, but this appears to be in the spirit of what the refactoring of this code was intended to do.

cbeer:
Could we remove `skip_invitation` from https://github.com/projectblacklight/spotlight/blob/master/app/models/spotlight/role.rb#L23?

app/models/spotlight/role.rb
cbeer: Yeah, doing this in the model isn't awesome. Maybe it's worth looking into the `RolesController` and figuring out if we have enough information there to send out invites after the models are created?

jkeck:
@cbeer how does this look?

PR #1874, opened 2017-12-21
Add helptext to add user form

camillevilla:
TODO:
- [x] Investigate why `master` build is failing on Travis #1875 

Related: sul-dlss/exhibits#1040

This PR adds help text to the new user form in Configuration > Users. The default i18n key can be overridden in `spotlight.roles.edit_fields.help`.

@ggeisler What do you think about this default helper text?
@jkeck opinions? 

## ~~Option 1: Bootstrappy~~
bootstrappy

```html
<div class="form-group">
  <label class="sr-only control-label" for="exhibit_roles_attributes_2_user_key">User key</label>
  <input aria-describedby="newUserHelpBlock" class="form-control" type="email" name="exhibit[roles_attributes][2][user_key]" id="exhibit_roles_attributes_2_user_key">
</div>
<small id="newUserHelpBlock" class="form-text text-muted">Please use a valid email address</small>
```

- [x] ~~throw in some CSS to fix the margin?~~

## Option 2: Form helper inline magic
form helper

```html
<div class="form-group">
  <label class="sr-only control-label" for="exhibit_roles_attributes_2_user_key">User key</label>
  <input class="form-control" type="email" name="exhibit[roles_attributes][2][user_key]" id="exhibit_roles_attributes_2_user_key">
  <span class="help-block">Please use a valid email address</span>
</div>
```

### DIFF
--- a/app/views/spotlight/roles/_edit_fields.html.erb
+++ b/app/views/spotlight/roles/_edit_fields.html.erb
@@ -1,6 +1,6 @@
         <tr data-edit-for="<%= f.object.new_record? ? 'new' : f.object.id %>">
           <td>
-            <%= f.email_field :user_key, hide_label: true, disabled: f.object.persisted? %>
+            <%= f.email_field :user_key, hide_label: true, disabled: f.object.persisted?, help: (t('.help') if f.object.new_record?) %>
           </td>
           <td><%= f.select :role, roles_for_select, hide_label: true %></td>
           <td></td>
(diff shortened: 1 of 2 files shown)


ggeisler:
@camillevilla Visually, I like Option 2. Specifically, the larger font size and the text being closer to the input. And code-wise I like using the `help-block` class versus using `small`, though I'm not sure that's important or my opinion about code is relevant. And as you suggest, with some extra CSS you could replicate the visual look of Option 2 with the HTML in Option 1.

Text-wise, it's a minor thing, but I'd prefer to avoid using "please". I'd rather be more direct with the language. So maybe just "Enter a valid email address".

jkeck:
👍 to Option 2 (and double 👍 if that is the approach this PR is currently doing)

PR #1878, opened 2018-01-10
Allow field specific display settings to be honored…

jkeck:
…on Blacklight Configuration initialization.

### DIFF
--- a/app/models/spotlight/blacklight_configuration.rb
+++ b/app/models/spotlight/blacklight_configuration.rb
@@ -319,7 +319,7 @@ module Spotlight
       return unless index_fields.blank?
 
       views = default_blacklight_config.view.keys | [:show, :enabled]
-      field.merge! Hash[views.map { |v| [v, !title_only_by_default?(v)] }]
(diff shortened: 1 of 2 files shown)


app/models/spotlight/blacklight_configuration.rb
```diff
@@ -319,7 +319,7 @@ def set_index_field_defaults(field)
       return unless index_fields.blank?
 
       views = default_blacklight_config.view.keys | [:show, :enabled]
-      field.merge! Hash[views.map { |v| [v, !title_only_by_default?(v)] }]
+      field.merge! Hash[views.map { |v| [v, field.send(v).nil? ? !title_only_by_default?(v) : field.send(v)] }]
```
cbeer: Is `field.merge! (views - field.keys).map { |v| [v, !title_only_by_default?(v)] }.to_h` any more satisfying/clear?

jkeck: Yeah, I like that a bit better.  Will update.  Thanks.

PR #1888, opened 2018-01-20
Add a configurable search box to browse categories

### LINKED ISSUE #1880: Add a (admin UI) configurable search box to browse categories
Taken from @ggeisler's https://github.com/sul-dlss/exhibits/issues/1025#issuecomment-353177081

---

Mockups for adding a search box to the browse category page are below. The mockups assume:

* Adding a search box to a browse category page is optional, and done on a per-browse category page basis (so a given exhibit can show it on some browse pages and not others).
* The default setting is to **not** show the search box (so there are no ramifications for existing exhibit browse pages).

(Also, one potential hurdle for implementation might be with clearly distinguishing the two search boxes on the same page for accessibility purposes. For reference, we do show two search boxes on the collection page in Arclight (e.g., https://arclight-demo.projectblacklight.org/catalog/umich-bhl-0234), though I can't remember if we have addressed the accessibility issue there yet.)

Edit: Because this ticket is in the Exhibits repo, I've mocked it up as a SUL exhibit, but maybe we should discuss whether, if we decide to implement it, it should be done in the Spotlight repo?

## Configuration
Add one setting to the Curation > Browse > Edit Browse Category description page, which enables the curator to display the search box on the browse category page:

browse-page-searchbox-admin-config

## Initial display of a browse category page with the search box enabled
If the configuration option is enabled for a browse category, the page displays a new section at the top of the results (note 1 in the mockup):

browse-page-searchbox

## Browse category page with a successful search
If the user enters a search term that matches some or all of the items in the browse category, the results area is updated as shown below. There are three pieces added:

1. A "clear search term" icon in the search box
2. A message notifying the user of how many items matched their term
3. A message that enables the user to easily broaden their search to the all exhibit items by clicking a link

browse-page-searchbox-results

## Browse category page with an unsuccessful search
If the terms the user searches for don't match any items in the browse category, no result items are shown, and the messages in notes 2 and 3 above are changed a bit:

browse-page-searchbox-no-results

camillevilla:
Fixes #1880 

This PR adds a configurable search box to browse categories. Users can provide a query to search within the browse category / saved search.

### Successful Search
successful_search

### No results
zero_results_search

## Todo:
- [x] test behavior for facet-only browse category
- [x] clear search link
- [x] link to search all items
- [x] add behavior for unsuccessful search
- [x] finish styling (clear results button)
- [x] add "1 of 5 results" 
- [x] accessibility check re: second search box
- [x] specs
- [x] maintain browse category total as results number
- [x] change text to use "browse category"
- [x] Fix JS
- [x] recenter search bar

### DIFF
--- a/app/assets/javascripts/spotlight/clear_form_button.js
+++ b/app/assets/javascripts/spotlight/clear_form_button.js
@@ -4,18 +4,24 @@ Spotlight.onLoad(function() {
 
 (function($) {
   $.fn.ClearFormButton = function() {
-    var clearBtn = this;
-    var input = $(clearBtn).parent().prev('input');
-    $(input).on('keyup', function() {
-      if (input.val() !== '') {
-        $(clearBtn).css('display', 'inline-block');
+    var $clearBtn = $(this);
+    var $input = $clearBtn.parent().prev('input');
+    var btnCheck = function(){
+      if ($input.val() !== '') {
+        $clearBtn.css('display', 'inline-block');
       } else {
-        $(clearBtn).css('display', 'none');
+        $clearBtn.css('display', 'none');
       }
+    };
+
+    btnCheck();
+    $input.on('keyup', function() {
+      btnCheck();
     });
-    $(clearBtn).on('click', function(event) {
+
+    $clearBtn.on('click', function(event) {
       event.preventDefault();
-      input.val('');
+      $input.val('');
     });
   };
 })(jQuery);
--- a/app/assets/javascripts/spotlight/clear_form_button.js
+++ b/app/assets/javascripts/spotlight/clear_form_button.js
@@ -1,0 +1,21 @@
+Spotlight.onLoad(function() {
+  $('.btn-reset').ClearFormButton();
+});
+
+(function($) {
+  $.fn.ClearFormButton = function() {
+    var clearBtn = this;
+    var input = $(clearBtn).parent().prev('input');
+    $(input).on('keyup', function() {
+      if (input.val() !== '') {
+        $(clearBtn).css('display', 'inline-block');
+      } else {
+        $(clearBtn).css('display', 'none');
+      }
+    });
+    $(clearBtn).on('click', function(event) {
+      event.preventDefault();
+      input.val('');
+    });
+  };
+})(jQuery);
--- a/app/helpers/spotlight/browse_helper.rb
+++ b/app/helpers/spotlight/browse_helper.rb
@@ -27,7 +27,7 @@ module Spotlight
     ##
     # Get parent results count of a browse category search
     def parent_search_count
-      Spotlight::Search.find(@search.id).count
+      @parent_search_count ||= Spotlight::Search.find(@search.id).count
     end
 
     private
--- a/app/helpers/spotlight/browse_helper.rb
+++ b/app/helpers/spotlight/browse_helper.rb
@@ -24,6 +24,12 @@ module Spotlight
       end
     end
 
+    ##
+    # Get parent results count of a browse category search
+    def parent_search_count
+      Spotlight::Search.find(@search.id).count
+    end
+
     private
 
     def view_available?(view)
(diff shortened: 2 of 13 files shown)


app/assets/javascripts/spotlight/clear_form_button.js
```diff
@@ -0,0 +1,21 @@
+Spotlight.onLoad(function() {
+  $('.btn-reset').ClearFormButton();
+});
+
+(function($) {
+  $.fn.ClearFormButton = function() {
+    var clearBtn = this;
+    var input = $(clearBtn).parent().prev('input');
+    $(input).on('keyup', function() {
```
jkeck: I wouldn't mind getting somebody a little more facile w/ jQuery to weigh in here, but a couple of things come to mind when looking over this.

1. Is `input` already a jQuery object at this point? (e.g. do we need to re-cast it as a jQuery object by calling `$(input)`).
1. Instead of casting `clearBtn` to a jquery object in a few places, can we do it when we assign the `clearBtn` var?
1. On the line below, can/should we use `this` which I _believe_ would be the input  in the `keypu` event scope?

One pattern I've seen (although admittedly haven't been good at using) is naming variables with a `$` when they are jQuery objects, just to know when we're dealing with one (e.g. `var $clearBtn = this`)

(FWIW, in a quick test drive of this branch I'm not getting the clear button, haven't tracked down the cause though)

app/helpers/spotlight/browse_helper.rb
```diff
@@ -24,6 +24,12 @@ def default_document_index_view_type
       end
     end
 
+    ##
+    # Get parent results count of a browse category search
+    def parent_search_count
+      Spotlight::Search.find(@search.id).count
```
jkeck: I get a bit nervous about this.  This triggers a solr request under the hood to get the document count.  Having this in a helper like this could trigger multiple page blocking solr requests w/o realizing it.

A few options I can think of initially.
1. Memoize this helper method
1. Move this logic into the `Spotlight::Search` class and memoize it there (could get into a weird state management issue though).
1. Others?

app/views/spotlight/searches/_form.html.erb
```diff
@@ -30,6 +30,10 @@
       <div role="tabpanel" class="tab-pane active" id="search-description">
         <%= f.text_field :title, control_col: "col-sm-5" %>
         <%= f.text_area :long_description, rows: 5 %>
+        <%= f.form_group :search_box, label: { text: t(:'.search_box.label'), class: nil }, help: nil do %>
+          <%= f.check_box :search_box, label: "" %>
+          <p class="help-block"><%= t(:'.search_box.help_block') %></p>
```
jkeck: Is this functionally different than setting ` t(:'.search_box.help_block')` as the `help` key for the `form_group` itself?  I think `bootstrap_form` may be able to handle this for us.

app/assets/javascripts/spotlight/clear_form_button.js
camillevilla: Thanks, @jkeck. I was able to take care of #1 and #2 and added `$` to the jQuery object variable names.

The clear button wasn't showing up until after a user starts typing in the form. 😅 I modified the function so that it checks for an input value when the page is first loaded. Let me know what you think.

app/views/spotlight/searches/_form.html.erb
camillevilla: 👍  good point. I've added the key to `form_group` and rebased.

app/helpers/spotlight/browse_helper.rb
camillevilla: Per our discussion, I went with option 1

camillevilla:
Alrighty, @jkeck. I've responded to each of your comments. Let me know how that looks and I can give this a squash 🎃

camillevilla:
@jkeck just making a quick CSS adjustment here. @ggeisler pointed out that the form is skewing left now 😅

PR #1909, opened 2018-02-21
Only copy over original for non-custom fields.

tpendragon:


app/models/spotlight/blacklight_configuration.rb
```diff
@@ -164,7 +164,7 @@ def blacklight_config
           config.facet_fields = Hash[config.facet_fields.sort_by { |k, _v| field_weight(facet_fields, k) }]
 
           config.facet_fields.each do |k, v|
-            v.original = v.dup
+            v.original = v.dup unless v.custom_field
```
jkeck: Would it be possible to include a test for this behavior?

tpendragon: Yeah, I can work on it this afternoon.

tpendragon:
@jkeck Added a test.

PR #1927, opened 2018-03-09
Implement language selector

### LINKED ISSUE #1911: Enable exhibit admin to select available locales/languages

To specify which languages are relevant for a given exhibit, the admin of the exhibit needs to:

- Select the locales/languages in which the exhibit should be viewable
- Control when those languages are available to exhibit visitors

The mockup below proposes to add a Languages tab to the Configuration > General page of the exhibit admin section to provide the controls to meet the requirements above. I suggest the Languages tab be inserted after the General tab, since it is likely to be used more than the other tabs on the page.

languages-tab
 
The numbered sections below correspond to the numbered orange note indicators in the mockup. (Also, note that while the mockup uses Stanford's color scheme, we should use the default Spotlight/Bootstrap colors; e.g., the red buttons should be the Bootstrap `.btn-primary` blue).

### 1. Language selection

Intro text for the top section:

> You can make the content and user interface of this exhibit available to exhibit visitors in multiple languages. Use the select box below to add a new language to the exhibit and then use the Curation > Translations page to provide translations for exhibit labels, browse categories, and curated page content.   

Only languages with existing (i.e., available to the app) locale files are available for selection in the Language menu. Sorted alphabetically in the menu.

### 2. Curation > Translations menu item

This new Translations menu item is not shown if there are no languages in the Current Languages section (to avoid confusion as to what it's for if no languages have been selected; alternatively, we should show it always, but include some text pointing to the General > Languages tab if there are no languages yet specified). As soon as the admin adds a language, the Translations menu item should be shown.

Although the Translations page (covered in other forthcoming tickets) provides the means for the exhibit curators to translate exhibit content, those translations aren't required. The exhibit admin might, for example, just want the standard Blacklight/Spotlight UI labels from the basic locale files to be shown, in which case they could just ignore the Translations page.

### 3. Current Languages section

Intro text:

> Each language listed below is available for translations on the Curation > Translations page. To enable exhibit visitors to change  the language of the exhibit, enable the Public checkbox. To hide translations from an available language from the exhibit, disable the Public checkbox.  

As soon as the admin has selected a language in the top section (Note 1), that language should be added to the table in this section. If no languages have been selected, omit the table in this section and show the text below instead of the intro text: 

> No languages have been added for translation. To add a language, make a selection above.  

### 4. Public checkbox

When a new language is added to the table, the Public checkbox for that language should be disabled by default (the assumption being that the exhibit curators will likely want to make some translations before enabling visitors to see the exhibit in that language).

When a language is added, even if it is not made public, it should be shown in the locale switcher when the current user is an **exhibit administrator or curator**. This will enable admins and curators to see how the exhibit site will look with any translations that have been made (or that exist in the base locale file for the selected language).

When the Public checkbox is enabled for a language (and the page is saved with the Save changes button), that language should then be available in the locale switcher to **all** exhibit users in the brand bar, which should enable an exhibit visitor to see any existing translations for that language as they navigate the exhibit.

If no languages have been added, do not show the locale switcher in the brand bar at all.

### 5. Removing a language

I suggest using `.btn-sm` to size the remove buttons in the languages table.

If the admin selects the “Remove” button for a language, show a confirmation modal:

languages-remove-confirmation

If the user confirms, remove the language from the Current Languages table. 

Open question: Do we also delete any translations made for that language? Or do we simply leave them in the database (since they won't be exposed anywhere if that language isn't selected as an available language), just in case the exhibit admin later changes her mind and adds that language back?

mejackreed:
Closes #1911 

This PR enables an exhibit admin to add, make public, and delete languages for an exhibit.

screen shot 2018-03-09 at 3 23 33 pm


ggeisler:
Looks good. From the screenshot it looks like the public checkbox isn't being centered, even though the `<td>` it is in is centered?

A bit more important, do you think we can update the select to have a default selected option of "Select language..."?

I realize my mockup is misleading so it's my fault, but I think it would be a bit clearer experience if the user saw "Select language..." in the menu when coming to the page rather than the first language in the list.

ggeisler:
Great, thanks for the extra work on the select!

(No problem with the checkbox, I get it. I can look into what it would take to center it in the Exhibits code down the road.)

config/locales/spotlight.en.yml
```diff
@@ -109,6 +115,15 @@ en:
       fields:
         facet:
           exhibit_tag: Exhibit Tags
+  locales:
```
cbeer: I'm uncomfortable with managing the language configuration in a yaml file; maybe this could move to application or engine configuration?

PR #1928, opened 2018-03-09
Update flaky JS tests

camillevilla:
They shall flap no more! 

This PR:
- treats some squeaky `expect` statements by either
  - nesting them inside of a `within` or
  - adding `visible: true` 
- disables image loading in Chrome headless
- ~~increases the max wait time in CI only~~

See #1974

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
 before_install:
   - gem update --system
   - gem install bundler
-  - google-chrome-stable --headless --disable-gpu --no-sandbox --remote-debugging-port=9222 http://localhost &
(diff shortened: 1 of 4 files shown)


mejackreed:
shall not pass

camillevilla:
Experimenting with using `#find`, which explicitly waits for the element to appear or the  `Capybara.default_max_wait_time`.
http://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Finders#find_field-instance_method

camillevilla:
sailor_mars_evil_spirits_be_gone

PR #1930, opened 2018-03-12
Add a locale picker into the user util links.

### LINKED ISSUE #1920: Add toggle for exhibit visitors to switch between languages
After one or more languages beyond the default language have been added to an exhibit (i.e., #1911) exhibit users need a way to switch languages, like this:

33725033-d5fd6e42-db25-11e7-95c3-f18a40bd4a0b

jkeck:
Closes #1920 

TODO:
- [ ] Make locales used in the locale switcher be their native language/script
- [ ] Sort locales
- [ ] Remove the currently selected locale from the drop down.


lib/spotlight/engine.rb
```diff
@@ -176,6 +176,17 @@ def self.blacklight_config
 
     Blacklight::Engine.config.inject_blacklight_helpers = false
 
+    config.i18n_locales = {
```
mejackreed: I wonder if we could manage this as a larger project? It feels unwieldy to manage the translation native languages in the engine config. What do you think about a spin off ticket for a small gem that did this?

jkeck: Would it make sense for us to have a 2nd or 3rd use case before making a decision on that kind of abstraction?  I'm not sure we have enough information to know if this is something that would be served well be its own gem, and how one might want to interact with it/extend it.

FWIW, this was pulled directly from how Wikipedia links to its different language specific sites (as suggested by @ggeisler).

mejackreed: Totally I get it, that is why I was just asking the question. It just seems arbitrary that we manage these native languages only for the locales that Blacklight provides. I def think this is the way to start this, but I see potential for others to use this also.

jkeck: Yeah, definitely agree, let's keep this in mind.  I agree it feels odd.

PR #1931, opened 2018-03-13
add basic implementation of general edit form

### LINKED ISSUE #1914: Enable curator to provide translations of general exhibit labels

When navigating to the Curation > Translations page, there should be five tabs, as shown in the mockup below. This ticket covers the General labels tab. The numbered sections below correspond to the numbered orange note indicators in the mockup. 

translation-general

### 1. Language buttons

Each language that has been added to the Current Languages section of the Configuration > General page should be represented by a button here. On page load the first button should be selected and the tabs on the page reflect the state of translation (progress counts and any existing translated field values) for that language.

When the user clicks a different language button, the page looks similar but loads the state of translation for that language. For example, here the user has clicked the Spanish button, which has only had one field translated so far:

translation-general-spanish

### 2. Progress counts

To give admins and curators a quick understanding of the translation progress for the selected language, each tab includes a progress indicator for the fields associated with that tab. For example, the General labels tab contains seven fields that could potentially be translated. In the first mockup above, there are three completed Italian translations, so the tab includes `[3/7]` as a progress indicator. I suggest using `#686868` for the text color of the progress indicator (I don't think we're using that color anywhere, but it's the lightest gray that meets AA color contrast level; Bootstrap's muted gray is too light).

### 3. Translatable fields

Each field that can be translated is shown with a field label at left and then a visually grouped block consisting of an input field for the translated field value and a help block below the input showing the current field value in the default language. Inputs are blank if no translated value exists; if a value does exist for a field, it's shown in the input so it can be edited.

Styling: Bootstrap's `panel panel-body` styling could possibly be used as a basis for the visual grouping, but we'd want to add `background-color: #f5f5f5; padding-bottom: 0;` to it. The field value shown below the input could be done as a `help-block` but I'd add `color: #555555; padding-left: 12px;` for contrast and alignment (there are Bootstrap variables for some of those property values, I think).

### 4. Completed translated icon

To help the curator more easily see which fields have been translated, show a check icon (e.g. `.glyphicon-ok` in `#80BF77`) next to each field with a value, after the page is saved. In other words, when this tabbed page is saved, we'd reload the page, updated to show 1) checkboxes next to each field with a value, and 2) currently accurate progress indicator in the General labels tab title (e.g., "[3/7]".

mejackreed:
Work outlined in #1914 

translations2

Fixes #1914. Unresolved items have been split off as additional tickets.

## Update guide
Make sure to update your `config/initializers/translation.rb` file to turn on the i18n-active_record backend.

## Fixture data
None, just create an exhibit

This is a proof of concept with some lingering questions: ✅ 

 - [x] How should the model attributes be translated throughout the application? Should we do something dynamic [like Globalize](https://github.com/globalize/globalize/blob/master/lib/globalize/active_record/act_macro.rb#L30-L51)?
 - [x] How do I know what the default language values are for some of the AR attributes (`title`, `subtitle`)? related to ^^ *Answer: Do not care at the moment*
 - [x] ~~These forms are going to be painful to build. What type of custom helpers/models can we build to help with this? https://github.com/projectblacklight/spotlight/compare/i18n-%F0%9F%8C%9D...add-general-edit-form?expand=1#diff-44ee77aa16a75e066de55c5be5448a46R28~~ split out to #1934
 - [x] ~~Should we put uniqueness validations on locale keys by language + exhibit?~~ Split off https://github.com/projectblacklight/spotlight/issues/1933
 - [x] ~~I'm using nest attributes here.. so after saving it redirects to `edit_exhibit_path`~~ resolved with controller changes

### DIFF
--- a/app/models/spotlight/ability.rb
+++ b/app/models/spotlight/ability.rb
@@ -26,7 +26,8 @@ module Spotlight
         Spotlight::Resource,
         Spotlight::Page,
         Spotlight::Contact,
-        Spotlight::CustomField
+        Spotlight::CustomField,
+        Translation
       ], exhibit_id: user.exhibit_roles.pluck(:resource_id)
 
       can :manage, Spotlight::Lock, by: user
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -120,6 +120,7 @@ Spotlight::Engine.routes.draw do
       end
     end
     post 'solr/update' => 'solr#update'
+    resource :translations, only: [:edit, :update]
   end
 
   get '/:exhibit_id' => 'home_pages#show', as: :exhibit_root
--- a/spec/models/spotlight/exhibit_spec.rb
+++ b/spec/models/spotlight/exhibit_spec.rb
@@ -299,4 +299,26 @@ describe Spotlight::Exhibit, type: :model do
   it 'is expected to be versioned' do
     is_expected.to be_versioned
   end
+  describe 'translatable fields' do
+    let(:persisted_exhibit) { FactoryBot.create(:exhibit, title: 'Sample', subtitle: 'SubSample') }
+    before do
+      FactoryBot.create(:translation, locale: 'fr', exhibit: persisted_exhibit, key: "#{persisted_exhibit.slug}.title", value: 'Titre français')
+      FactoryBot.create(:translation, locale: 'fr', exhibit: persisted_exhibit, key: "#{persisted_exhibit.slug}.subtitle", value: 'Sous-titre français')
+    end
+    after do
+      I18n.locale = 'en'
+    end
+    it 'has a translatable title' do
+      expect(persisted_exhibit.title).to eq 'Sample'
+      I18n.locale = 'fr'
+      persisted_exhibit.reload
+      expect(persisted_exhibit.title).to eq 'Titre français'
+    end
+    it 'has a translatable subtitle' do
+      expect(persisted_exhibit.subtitle).to eq 'SubSample'
+      I18n.locale = 'fr'
+      persisted_exhibit.reload
+      expect(persisted_exhibit.subtitle).to eq 'Sous-titre français'
+    end
+  end
 end
--- a/app/assets/stylesheets/spotlight/_spotlight.scss
+++ b/app/assets/stylesheets/spotlight/_spotlight.scss
@@ -30,6 +30,8 @@
 @import "spotlight/report_a_problem";
 @import "spotlight/exhibits_index";
 @import "spotlight/collapse_toggle";
+@import "spotlight/translations";
+@import "spotlight/utilities";
 
 
 
--- a/app/assets/stylesheets/spotlight/_translations.scss
+++ b/app/assets/stylesheets/spotlight/_translations.scss
@@ -1,0 +1,25 @@
+.translation-subheading {
+  border-bottom: 1px solid $navbar-default-border;
+  margin-bottom: 20px;
+}
+
+.form-group.translation-form {
+  .control-label {
+    padding-top: $panel-body-padding + $padding-base-vertical;
+  }
+
+  .glyphicon-ok {
+    color: $translation-available-color;
+    padding-top: $panel-body-padding + $padding-base-vertical;
+  }
+}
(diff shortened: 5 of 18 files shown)


config/routes.rb
```diff
@@ -120,6 +120,8 @@
       end
     end
     post 'solr/update' => 'solr#update'
+    resource :translations, only: [:edit, :update] do
+    end
```
jkeck: I think this block may be redundant.

app/models/spotlight/ability.rb
```diff
@@ -26,7 +26,8 @@ def initialize(user)
         Spotlight::Resource,
         Spotlight::Page,
         Spotlight::Contact,
-        Spotlight::CustomField
+        Spotlight::CustomField,
+        Translation
```
jkeck: FWIW, there is an ability spec if you want to throw in an expectation about curators being able to access translations.  We haven't been super consistent about adding things there, so don't consider this a blocker.

spec/models/spotlight/exhibit_spec.rb
```diff
@@ -299,4 +299,26 @@
...
     is_expected.to be_versioned
   end
+  describe 'translatable fields' do
+    let(:persisted_exhibit) { FactoryBot.create(:exhibit, title: 'Sample', subtitle: 'SubSample') }
+    before do
+      FactoryBot.create(:translation, locale: 'fr', exhibit: persisted_exhibit, key: "#{persisted_exhibit.slug}.title", value: 'Titre français')
+      FactoryBot.create(:translation, locale: 'fr', exhibit: persisted_exhibit, key: "#{persisted_exhibit.slug}.subtitle", value: 'Sous-titre français')
+    end
+    after do
+      I18n.locale = 'en'
```
jkeck: Another approach here (that doesn't require teardown) would be to do a `allow` / `expect` in the test (e.g. `expect(I18n).to receive(:locale).and_return(:fr)`).

Not a huge deal, but ran into this in my other test and found it a bit cleaner to stub instead of setup/teardown.

config/routes.rb
mejackreed: yep totally that is cruft

app/models/spotlight/ability.rb
mejackreed: I'll look into adding that.

app/controllers/spotlight/exhibits_controller.rb
```diff
@@ -89,8 +89,7 @@ def exhibit_params
         :description,
         :published,
         :tag_list,
-        contact_emails_attributes: [:id, :email],
-        languages_attributes: [:id, :public]
```
camillevilla: @mejackreed I think removing this line is affecting the Current Languages form. I can't update the language to "Public" now.

camillevilla: Interesting! It passes this test because it gets the  `
The exhibit was successfully updated` flash message. But when the page reloads the box is still unchecked and the ActiveRecord object is unchanged.

https://github.com/projectblacklight/spotlight/blob/21f62f6ce7edbd70cf6bd5440b2d69484a2ee6ad/spec/features/exhibits/language_create_edit_spec.rb#L28-L32

jkeck: Nice catch! Maybe we can do like a `expect(exhibit.languages.last).to be_public` or something like that as well.

mejackreed: Awesome catch @camillevilla , thanks for the recommendation @jkeck . Test added in dbc700e7ba1f76de21f1d7fa90448606793ec2a7 and an update removing this change (accidental) in 2059a7bd626815c13f5af3e8faf576b43d4777a2

app/models/spotlight/ability.rb
mejackreed: Added in 2059a7bd626815c13f5af3e8faf576b43d4777a2

config/routes.rb
mejackreed: Removed in 2059a7bd626815c13f5af3e8faf576b43d4777a2

PR #1940, opened 2018-03-15
Add ability to translate title/description of browse categories

### LINKED ISSUE #1917: Enable curator to provide translations of browse category fields

When navigating to the Curation > Translations page, there should be five tabs, as shown in the mockup below. This ticket covers the Browse categories tab. The numbered sections below correspond to the numbered orange note indicators in the mockup.

translations-browse

### 1. and 2. Column headers

The layout of the Browse categories tab is somewhat different from the General labels, Metadata field labels, and Search field labels tabs because the browse categories are very repetitive in terms of fields. They all have a title, and optionally a description and that's it (in terms of translatable fields). Also, since (in Stanford's experience at least) exhibits can have many browse categories but these often don't include a description, this layout is intended to only show what the translator is most likely to need to translate a bunch of browse categories efficiently.

The layout here is thus just two columns, where the first column (note 1 in the mockup) has a heading of "English language title", where "English" could in theory be whatever the default language of the exhibit is, although we might want to just hardcode English for now and deal with the possibility of a dynamic default language later.

The second column header (note 2 in the mockup) should reflect whatever the currently selected language is.

### 3. Browse category titles and inputs

Under the headers are rows of title/input pairs, where each title is the default language title of a browse category. 

### 4. Browse category description labels

If, and only if, a given browse category has a description value in the default language, show a Description label with a toggle icon as shown in note 4 of the mockup. In other words, if a given browse category does not have a description in the default language, we do not enable inputting a description in the translated language.

### 5. Browse category description translation
If there is a description in the default language, a translated description is still optional, but if the user opens the Description toggle an input for translation is shown like this:

translations-browse-open

There is no save button associated with the description field; the user must use the Save changes button for the page to save the description. The user should be able to have open multiple description sections at the same time (open one, enter a description, open the next one, enter a description, etc. and then save the page to save all translations).

See #1914 for styling suggestions that can be used for the open description section and showing the checkbox icon for any fields that have been translated.

jkeck:
Closes #1917 

## Open Questions
- What should we do if there are no browse categories? (Don't show the tab / Put some explanation/help text as the tab content?)
- I'm using the browse category's `slug` as part of the key (similar to how we do w/ the exhibit) so it can be fetched appropriately by the `Translatables` mixin.  This has the potential of causing an issue if you were to name a browse category the same thing as your exhibit. A few ways that we could solve this is adding some sort of normalization of the class name into the key or to have the Translation have a polymorphic relationship (not sure how that will play w/ how we've set things up so far though).
- Another issue w/ using the `slug` is that if a browse category is renamed, its slug is updated (this will break all translation keys as-is).  We've run into this in pages that include the Browse Category widget as well.  This makes me wonder if we should make the slugs static (similar to what we do w/ the exhibits themselves).

admin-ui

user-facing

### DIFF
--- a/app/views/spotlight/translations/_browse_categories.html.erb
+++ b/app/views/spotlight/translations/_browse_categories.html.erb
@@ -1,0 +1,68 @@
+<div role="tabpanel" class="tab-pane" id="browse">
+  <%= bootstrap_form_for current_exhibit, url: spotlight.exhibit_translations_path(current_exhibit), layout: :horizontal do |f| %>
+    <% # Add a hidden field for the language so the redirect knows how to come back here %>
+    <%= hidden_field_tag :language, @language %>
+    <div class="row">
+      <div class="col-xs-4 text-right">
+        <strong>
+          <%= t :'spotlight.exhibits.translations.browse_categories.default_language_column_label' %>
+        </strong>
+      </div>
+      <div class="col-xs-7">
+        <strong>
+          <%= t :'spotlight.exhibits.translations.browse_categories.translation_column_label', language: t("locales.#{@language}")  %>
+        </strong>
+      </div>
+    </div>
+
+    <% current_exhibit.searches.each do |search| %>
+      <% title_translation = Translation.find_or_initialize_by(exhibit: current_exhibit, key: "#{search.slug}.title", locale: @language) %>
+      <%= f.fields_for :translations, title_translation do |translation_fields| %>
+        <%= translation_fields.hidden_field :key %>
+        <%= translation_fields.hidden_field :locale %>
+        <div class="form-group">
+          <%= translation_fields.label :value, search[:title], class: 'control-label col-xs-4' %>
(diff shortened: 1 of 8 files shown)


app/views/spotlight/translations/_browse_categories.html.erb
```diff
@@ -0,0 +1,66 @@
+<div role="tabpanel" class="tab-pane" id="browse">
+  <%= bootstrap_form_for current_exhibit, url: spotlight.exhibit_translations_path(current_exhibit), layout: :horizontal do |f| %>
```
mejackreed: Would you mind adding a hidden field here for the language? https://github.com/projectblacklight/spotlight/pull/1940/files#diff-f6fd4ed34036df67d84f28e10b98eac8R3

This way the page should correctly reload this language tab after save

jkeck: Good catch, will add that.

PR #1941, opened 2018-03-15
Mark blank translations for deletion (before validation)

jkeck:


cbeer:
Empty out, meaning e.g. going to the form and deleting the value from the box?

mejackreed:
`allow_blank: false` as a validation? I'm not sure if this fits all of our potential scenarios though.

jkeck:
I would be worried that using `allow_blank: false` would throw a validation error as opposed to give us the behavior that we want.

It seems like deleting the translation itself would be the desired result, but wasn't sure about the best rails way to do that w/o adding a bunch of potentially complicated callback logic (although happy to explore further if this seems unacceptable).

cbeer:
Alternatively, I suppose, we could add some callback (or override the setter?) on the model to call `#marked_for_destruction` if the value is blank?

jkeck:
I couldn't get `#marked_for_destruction` to ever really work right, so I went w/ the first approach.  Let me know what ya'll think.

app/controllers/spotlight/translations_controller.rb
```diff
@@ -19,6 +19,13 @@ def update
 
     private
 
+    def translations_attributes_destroying_blanks
+      exhibit_params.to_h[:translations_attributes].map do |index, attrs|
+        attrs['_destroy'] = true if attrs['value'].blank?
+        [index, attrs]
```
jkeck: I was thinking I could guard against sending the `_destroy` key when there is no `id`, but this seems to workout just fine.

cbeer: Instead of `#map`, maybe `#transform_values`?

PR #1942, opened 2018-03-16
Redirect to the same Configuration > General tab after forms are saved

### LINKED ISSUE #1932: User should stay on Config > Languages tab when adding language or saving
Currently, when the user adds a language on the Configuration > Languages tab (https://github.com/projectblacklight/spotlight/pull/1927) they end up back on the Configuration > General tab. Same thing happens when the user makes a change to the Current Languages table and saves the page.

The user should stay on the Configuration > Languages tab when adding a language or selecting Save on the page. This would be consistent with the Save behavior on the other tabs within Configuration, and more helpful for the user who wants to add multiple languages or make a language public after adding it.

camillevilla:
Fixes #1932 

## Approach:
1. Pass a `tab` parameter through a hidden_field on each form within Configuration > General
2. Pass the `tab` parameter to the controller action's redirect 
2. In the view, apply the `active` class to the nav tab and `div.tab-content` that have an id that matches the `tab` param

Note: h/t to @mejackreed for providing [an example of this approach](https://github.com/projectblacklight/spotlight/pull/1931/files#diff-059f2c780562af6d663a2f0d5cd9038bR9)

### Before
`http://test.host/spotlight/exhibit-title-714/edit#language`
### After
`http://test.host/spotlight/exhibit-title-714/edit?tab=language`

## Config > General tabs

- Basic settings
- Languages
- Filter
- Import (unchanged; redirects to exhibit dashboard)
- Export (unchanged; stays on the same page after downloading)
- Delete (unchanged; redirects to exhibits#index)

### DIFF
--- a/app/controllers/spotlight/filters_controller.rb
+++ b/app/controllers/spotlight/filters_controller.rb
@@ -11,14 +11,14 @@ module Spotlight
       else
         flash[:alert] = @filter.errors.full_messages.join('<br/>'.html_safe)
       end
-      redirect_to spotlight.edit_exhibit_path @exhibit, anchor: 'filter'
+      redirect_to spotlight.edit_exhibit_path @exhibit, tab: 'filter'
     end
 
     def update
       unless @filter.update(filter_params)
         flash[:alert] = @filter.errors.full_messages.join('<br/>'.html_safe)
       end
-      redirect_to spotlight.edit_exhibit_path @exhibit, anchor: 'filter'
+      redirect_to spotlight.edit_exhibit_path @exhibit, tab: 'filter'
     end
 
     def filter_params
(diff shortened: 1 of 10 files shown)


app/controllers/spotlight/filters_controller.rb
```diff
@@ -11,14 +11,14 @@ def create
       else
         flash[:alert] = @filter.errors.full_messages.join('<br/>'.html_safe)
       end
-      redirect_to spotlight.edit_exhibit_path @exhibit, anchor: 'filter'
+      redirect_to spotlight.edit_exhibit_path @exhibit, params: { tab: 'filter' }
```
jkeck: I'm kind of surprised to the `params` key.  I wasn't aware that `params` was a special thing (I don't see any reference to it in the `redirect_to` documentation.  Does this offer any benefit over doing something like

```ruby
redirect_to spotlight.edit_exhibit_path(@exhibit, tab: 'filter')
```
?

camillevilla: Nope! No extra benefits. Looks like I might have been referencing an outdated snippet... I've updated this PR with the more concise syntax. ✂️

PR #1947, opened 2018-03-17
adds main menu navigation translation functionality

### LINKED ISSUE #1936: Add "main menu" to general translations section
Split out of #1914

Main menu section needs to be added.

translation-general

mejackreed:
Closes #1936 

Adds "main menu" translations to "General" translations.

*Note to reviewer* You may need to enable browse categories, about pages

menutranslation

### DIFF
--- a/app/models/spotlight/exhibit.rb
+++ b/app/models/spotlight/exhibit.rb
@@ -20,7 +20,7 @@ module Spotlight
 
     extend FriendlyId
     friendly_id :title, use: [:slugged, :finders]
-    validates :title, presence: true
+    validates :title, presence: true, if: -> { I18n.locale == I18n.default_locale }
     validates :slug, uniqueness: true
     validates :theme, inclusion: { in: Spotlight::Engine.config.exhibit_themes }, allow_blank: true
 
--- a/app/models/spotlight/main_navigation.rb
+++ b/app/models/spotlight/main_navigation.rb
@@ -2,11 +2,15 @@ module Spotlight
   ##
   # Exhibit navbar links
   class MainNavigation < ActiveRecord::Base
+    include Spotlight::Translatables
+
     belongs_to :exhibit, touch: true
     default_scope { order('weight ASC') }
     scope :browse, -> { find_by(nav_type: 'browse') }
     scope :about, -> { find_by(nav_type: 'about') }
+    scope :curated_features, -> { find_by(nav_type: 'curated_features') }
     scope :displayable, -> { where(display: true) }
+    translates :label
 
     def displayable?
       display?
@@ -22,6 +26,14 @@ module Spotlight
 
     def default_label
       I18n.t(:"spotlight.main_navigation.#{nav_type}")
+    end
+
+    private
+
+    ##
+    # Allows us to scope translations namespace.
+    def slug
+      ['main_navigation', nav_type].join('.')
     end
   end
 end
--- a/app/views/spotlight/translations/_general.html.erb
+++ b/app/views/spotlight/translations/_general.html.erb
@@ -66,6 +66,88 @@
         </div>
       <% end %>
     </div>
+    <div class='translation-main-menu'>
+      <h2 class='translation-subheading'>
+        <%= t('spotlight.exhibits.translations.general.main_menu.label') %>
+      </h2>
+
+      <% translation = Translation.find_or_initialize_by(exhibit: current_exhibit, key: "spotlight.curation.nav.home", locale: @language) %>
+      <%= f.fields_for :translations, translation do |translation_fields| %>
+        <%= translation_fields.hidden_field :key %>
+        <%= translation_fields.hidden_field :locale %>
+        <div class='form-group translation-form translation-main-menu-home'>
+          <%= translation_fields.label :value, t('spotlight.exhibits.translations.general.basic_settings.home'), class: 'control-label col-sm-2' %>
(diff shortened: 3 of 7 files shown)


app/models/spotlight/main_navigation.rb
```diff
@@ -23,5 +27,13 @@ def label_or_default
     def default_label
       I18n.t(:"spotlight.main_navigation.#{nav_type}")
     end
+
+    private
+
+    ##
+    # Allows us to scope translations namespace.
+    def slug
+      ['main_navigation', nav_type].join('.')
```
mejackreed: While this `main_navigation` isn't absolutely necessary, I thought it might help give context to someone who came across the key.

app/views/spotlight/translations/_general.html.erb
```diff
@@ -66,6 +66,88 @@
...
+      <% translation = Translation.find_or_initialize_by(exhibit: current_exhibit, key: "main_navigation.browse.label", locale: @language) %>
+      <%= f.fields_for :translations, translation do |translation_fields| %>
+        <%= translation_fields.hidden_field :key %>
+        <%= translation_fields.hidden_field :locale %>
+        <div class='form-group translation-form translation-main-menu-browse'>
+          <%= translation_fields.label :value, t('spotlight.exhibits.translations.main_menu.browse'), class: 'control-label col-sm-2' %>
+          <div class='col-md-8 panel panel-body panel-translation'>
+            <%= translation_fields.text_field_without_bootstrap :value, class: 'form-control' %>
+            <p class="help-block">
+              <%= current_exhibit.main_navigations.browse.label %>
```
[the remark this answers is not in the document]
mejackreed: Yeah I was thinking about this a bit. I wonder if we should just consider that the admin pages should only be rendered in the default locale instead of providing for a way for it to also be translated.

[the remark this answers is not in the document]
jkeck: I worry that could get unnecessarily complicated.

That's one of the reasons I added the Fallbacks module, so that in the absence of an i18n'd value you'll get the English version (which is also the way the behavior has been described in multiple places around the app by @ggeisler).

We've used the `[]` accessor in other parts of the General and Browse categories forms to make this work appropriately under this pattern.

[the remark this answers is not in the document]
mejackreed: Ah that makes total sense. For some reason I had not realized that a multilingual admin UI was expected, but it makes sense why we would want that especially with multiple curators/translators.

camillevilla:
@jkeck so I found a thing, but I'm not sure it needs to be addressed in this PR. We're not generating a translation for the `Home` link in the breadcrumbs. That logic is controlled by the `#attach_breadcrumbs` method, which some controllers `super` up to while others implement their own. Switching out `spotlight.exhibits.breadcrumb` for `spotlight.curation.nav.home` should do the trick, but it seems like there's enough going here ¯\\_(ツ)_/¯ Should I just open up a new ticket? 

### About page
home breadcrumb link is not translated in about page

https://github.com/projectblacklight/spotlight/blob/e28229790f175506a37a5604591fcac07b662d84/app/controllers/spotlight/about_pages_controller.rb#L18-L19

https://github.com/projectblacklight/spotlight/blob/e28229790f175506a37a5604591fcac07b662d84/app/controllers/spotlight/pages_controller.rb#L130-L136

### Browse page
home breadcrumb link is not translated in browse page

https://github.com/projectblacklight/spotlight/blob/e28229790f175506a37a5604591fcac07b662d84/app/controllers/spotlight/browse_controller.rb#L58-L61

app/models/spotlight/exhibit.rb
```diff
@@ -20,7 +20,7 @@ class Exhibit < ActiveRecord::Base
 
     extend FriendlyId
     friendly_id :title, use: [:slugged, :finders]
-    validates :title, presence: true
+    validates :title, presence: true, if: -> { I18n.locale == I18n.default_locale }
```
jkeck: I'm adding this because otherwise we were getting a validation error saving translations (through the exhibit via `nested_attributes`) and .

However; this can have the affect of allowing a site administrator to empty out the Exhibit's title if they are on that page in a non-default locale).  There are a few things we could do about this including changing the validation or disabling the normal exhibit label input when viewing in a non-default locale), but I'm not sure how much we would need to worry about it.

I would be happy to break dealing w/ this out into a separate ticket (although I think this change now would be useful because the form is effectively broken otherwise).

jkeck:
@camillevilla ah, yeah, I had noticed that at one point.  Yeah, let's make that into a new ticket. 👍

app/models/spotlight/exhibit.rb
jkeck: @cbeer or @ggeisler have any thoughts on this?

ggeisler: I would be fine with going with this approach now. But ticketing a more long-term solution to do validation for the translated values seems like a good idea.

jkeck: Just to make sure I'm being clear, I'm not suggesting we do any validations for the non-default language. I'll add more details in a new ticket shortly.

ggeisler: Oh, I think I see. `presence` only needs to be enforced in the default locale because if it's not there in the translated language we just use the default locale version.

PR #1955, opened 2018-03-21
Translate 'Home' breadcrumb

### LINKED ISSUE #1953: "Home" breadcrumb link is not being translated for Browse categories, Feature pages, and About pages
See #1936 // This work should be merged into the i18n-🌝feature branch.

We're not generating a translation for the `Home` link in the breadcrumbs. That logic is controlled by the `#attach_breadcrumbs` method, which some controllers `super` up to while others implement their own. Switching out `spotlight.exhibits.breadcrumb` for `spotlight.curation.nav.home` should do the trick. The latter can draw on translations in the `Main Menu` section.

### About page
home breadcrumb link is not translated in about page

https://github.com/projectblacklight/spotlight/blob/e28229790f175506a37a5604591fcac07b662d84/app/controllers/spotlight/about_pages_controller.rb#L18-L19

https://github.com/projectblacklight/spotlight/blob/e28229790f175506a37a5604591fcac07b662d84/app/controllers/spotlight/pages_controller.rb#L130-L136

### Browse page
home breadcrumb link is not translated in browse page

https://github.com/projectblacklight/spotlight/blob/e28229790f175506a37a5604591fcac07b662d84/app/controllers/spotlight/browse_controller.rb#L58-L61

camillevilla:
Closes #1953

This PR changes the i18n key for the "Home" breadcrumb so that it can utilize translations provided by the curator. 

### About page
about_page

### Browse categories
browse_categories

### Feature page
feature_page

### DIFF
--- a/app/controllers/spotlight/browse_controller.rb
+++ b/app/controllers/spotlight/browse_controller.rb
@@ -56,7 +56,7 @@ module Spotlight
     end
 
     def attach_breadcrumbs
-      add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: @exhibit.title), @exhibit
+      add_breadcrumb t(:'spotlight.curation.nav.home', title: @exhibit.title), @exhibit
       add_breadcrumb(@exhibit.main_navigations.browse.label_or_default, exhibit_browse_index_path(@exhibit))
     end
 
--- a/app/controllers/spotlight/pages_controller.rb
+++ b/app/controllers/spotlight/pages_controller.rb
@@ -129,9 +129,9 @@ module Spotlight
 
     def attach_breadcrumbs
       if view_context.current_page? '/'
-        add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: current_exhibit.title), main_app.root_path
+        add_breadcrumb t(:'spotlight.curation.nav.home', title: current_exhibit.title), main_app.root_path
       else
-        add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: current_exhibit.title), spotlight.exhibit_root_path(current_exhibit)
+        add_breadcrumb t(:'spotlight.curation.nav.home', title: current_exhibit.title), spotlight.exhibit_root_path(current_exhibit)
       end
     end
 


jkeck:
A few questions I have w/ this change.

1. It looks there are a few other places where `spotlight.exhibits.breadcrumb` is used in breadcrumbs around the app, should we change them all?
1. If we do change those, should we remove `spotlight.exhibits.breadcrumb` from our locale file?

camillevilla:
@jkeck Most of the instances of `spotlight.exhibits.breadcrumb` are for breadcrumbs within the Admin / Curator UI. Here's an example:

example_admin_breadcrumbs

I think we'd want to retrain this key so that it _only_ gets translated with the rest of the admin UI via a contributed locale file. If we change these keys to `spotlight.curation.nav.home`, we'll end up with a mix of languages. Something like:

inconsistent mix of translated elements

FWIW I'll take another look at the list and make sure the remaining instances are in the admin UI.

camillevilla:
**The good news:** 
Indeed, the appearances of `spotlight.exhibits.breadcrumb` in the following controllers are for admin UI breadcrumbs:
- Appearances
- Contacts - add exhibit
- Custom fields
- dashboards
- exhibits
- Metadata configurations
- resources
- roles
- Search configurations
- searches
- Tags

**The bad news**
It looks like the admin UI and the user UI are connected to the same `attach_breadcrumbs` method for both Feature pages and About pages. Thus we get:

### Feature - user
feature_user-facing

### Feature - admin
feature_admin-facing

### About - user
about_user-facing

### About - admin
about_admin-facing

jkeck:
Yeah, I was worried about that.  I wonder if we'll be better off just being consistent in the link we use at the beginning of the breadcrumb.

On a side-note (that this made me notice), it looks like there are going to be other Breadcrumbs that we present to end-users that are from Spotlight (since Blacklight doesn't have any breadcrumb functionality).  These are end user-facing strings that are provided by Spotlight, and if we need them to be internationalized, we'll either need to provide a UI to do so, or the keys for all the languages we intend to support.  _sigh_

camillevilla:
@jkeck Yeahhhhh...I ran across an instance of this in the Catalog controller and was going to open up a separate ticket to discuss the headache. 😱

Here's an example that has `spotlight.exhibits.breadcrumb` (Home) and `spotlight.exhibits.index` (Search Results). 


genre__map_-_test-fr_-_blacklight_resultats_de_recherche

camillevilla:
Alrighty! I just pushed an update to this PR. I'm splitting off the issue about breadcrumbs in search results to #1991 because it involves adding a new field to the Translations form.

PR #1956, opened 2018-03-22
Add the Search field labels tab to the Translations configuration

### LINKED ISSUE #1916: Enable curator to provide translations of search-related field labels

When navigating to the Curation > Translations page, there should be five tabs, as shown in the mockup below. This ticket covers the Search field labels tab.

The Search field labels tab should include every field that is shown in the three tabs of the Configuration > Search page, with a section headings for each tabbed page, as shown in the mockup below. The layout and styling of the fields should be very similar to the General labels tab; see #1914 for details.

The Field-Based Search Options section should show the editable fields on the Configuration > Search, Options tab. The Facets section should show the facet fields on the Configuration > Search, Facets tab. And the Sort Fields section should show the editable fields on the Configuration > Search, Results tab. 

translations-search-fields

jkeck:
Closes #1916 

## Config > Translations > Search field labels 
search-field-labels

### Field-based search fields
translated_field_based_search

### Facet fields
translated_facet_fields

### Sort fields
translated_sort_fields

### DIFF
--- a/app/views/spotlight/translations/_search_fields.html.erb
+++ b/app/views/spotlight/translations/_search_fields.html.erb
@@ -3,6 +3,34 @@
     <% # Add a hidden field for the language so the redirect knows how to come back here %>
     <%= hidden_field_tag :language, @language %>
 
+    <div class='translation-field-based-search-fields'>
+      <h2 class='translation-subheading'>
+        <%= t('spotlight.exhibits.translations.search_fields.field_based_search_fields.label') %>
+      </h2>
+
+      <% current_exhibit.blacklight_config.search_fields.select { |_, config| config.if }.each do |key, search_config| %>
+        <% translation = Translation.find_or_initialize_by(exhibit: current_exhibit, key: "blacklight.search.fields.search.#{key}", locale: @language) %>
+        <%= f.fields_for :translations, translation do |translation_fields| %>
+          <%= translation_fields.hidden_field :key %>
+          <%= translation_fields.hidden_field :locale %>
+          <div class='form-group translation-form'>
+            <%= translation_fields.label :value, t("spotlight.search.fields.search.#{key}", locale: I18n.default_locale), class: 'control-label col-sm-2' %>
+            <div class='col-md-8 panel panel-body panel-translation'>
+              <%= translation_fields.text_field_without_bootstrap :value, class: 'form-control' %>
+              <p class="help-block">
+                <%= search_config.label %>
+              </p>
+            </div>
+            <div class='col-md-2'>
+              <% if translation.value.present? %>
+                <span class='glyphicon glyphicon-ok'></span>
+              <% end %>
+            </div>
+          </div>
+        <% end %>
+      <% end %>
+    </div>
+
     <div class='translation-sort-fields'>
       <h2 class='translation-subheading'>
         <%= t('spotlight.exhibits.translations.search_fields.sort_fields.label') %>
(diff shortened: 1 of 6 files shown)


app/views/spotlight/translations/_search_fields.html.erb
```diff
@@ -0,0 +1,68 @@
...
+          <%= translation_fields.hidden_field :locale %>
+          <div class='form-group translation-form'>
+            <%= translation_fields.label :value, t("spotlight.search.fields.search.#{key}", locale: I18n.default_locale), class: 'control-label col-sm-2' %>
+            <div class='col-md-8 panel panel-body panel-translation'>
+              <%= translation_fields.text_field_without_bootstrap :value, class: 'form-control' %>
+              <p class="help-block">
+                <%= search_config.label %>
+              </p>
+            </div>
+            <div class='col-md-2'>
```
jkeck: @ggeisler not sure if you have any thoughts on this. I've been copying this grid layout for the form, but it might be a good idea to call attention to the layout as we've implemented it thus far.

I think we've done a 1 column checkbox/status column in the browse categories.  Also, we've made it so that the status column never wraps under the input/panel in browse categories (but as we've implemented this here, we get the columns stacking on smaller viewports).  This may not be a huge deal because the dashboard is already not very responsive, but thought I would just call it out now before we proliferate the pattern further.

ggeisler: @jkeck Yeah, this is a good thing to bring up. The 1-column approach seems fine, **if** we don't ever get that column wrapped under the input it is associated with, because that might be hard for the user to interpret. When you say:

> Also, we've made it so that the status column never wraps under the input/panel in browse categories (but as we've implemented this here, we get the columns stacking on smaller viewports).

I'm not sure if you mean the check is always next to the input for browse categories, but not for the other tabs? If so, is the technique used for browse categories feasible in the other tabs?

Otherwise, as you say, we have many responsive issues in general, so we could put this off and just put it in that bucket for the future, but I do wonder if we should deal with this now, while we're in the middle of the translation feature. 

It would take some adjustment to existing elements, but I wonder if an approach that just adds a fixed-width `<span>` (so the space is reserved whether or not there is a check icon to display) next to the input (in the same `col` rather than adding the `col-1`) is worth switching to? That way we could get the check to always stay with the associated input (I don't have a branch with translations running, so just using the General settings form as an example):

configuration_-_general___test_exhibit_-_blacklight-2
---
configuration_-_general___test_exhibit_-_blacklight

This approach might be more problematic if the check icon `span` needed to be wider, but we'd only need ~15px for it (with `margin-left: 10px` or something).

But again, if the browse categories already keeps the check icon next to the input, however that is done might be just as effective as the above idea.

jkeck: Sorry, I wasn't being clear.  We're using `xs` column sizes on the browse category inputs to achieve this, and we could do something similar here.

It might be a good idea to take a quick look over how the current form is responding in different viewports since we're kinda doing our own custom thing here.  Perhaps we should split that out into a separate issue and not let that hold up the work here.

jkeck: #1957

ggeisler: @jkeck Responded in #1957. I would've understood what you were saying here (and why my suggestion above made no sense) had I actually looked at the translation tabs code; I was just in the middle of something else and being lazy... but hopefully we are on the right track with the approach in #1957.

PR #1959, opened 2018-03-23
Add a Metadata field labels tab to the Translations configuration...

jkeck:
...admin UI.

Closes #1915 

## Admin UI
md-admin-ui

## Search Results
yh941cp6284-result

## Details/Show PAGE
yh941cp6284

Issue #1919, opened 2018-02-28
Enable curator to provide translations of home, feature, and about page content


ggeisler:
(Part of #1912, a larger feature aimed at providing multilingual exhibits.)

Mockups and details forthcoming.

PR #1964, opened 2018-03-24
Redirect to the same Config > Translations tab after submitting form

### LINKED ISSUE #1958: Have translations configuration UI to return to the tab that the form was submitted from
Similar to the pattern in #1942

camillevilla:
Closes #1958 

This PR enhances redirect behavior for the Config > Translations form. After a user submits a translation form (e.g. "Search fields"), they will be sent back to the same tab.

**Note:** This only implements redirects for General, Search fields, and Browse categories. This behavior will need to be added to Metadata fields (#1959) and Pages (#1919) once available. 

## Demo
translations-tab-redirects

### DIFF
--- a/spec/features/exhibits/translation_editing_spec.rb
+++ b/spec/features/exhibits/translation_editing_spec.rb
@@ -71,6 +71,17 @@ describe 'Translation editing', type: :feature do
   describe 'Search field labels' do
     before { visit spotlight.edit_exhibit_translations_path(exhibit, language: 'fr') }
 
+    it 'redirects to the same form tab' do
+      click_link 'Search field labels'
+      within('#search_fields', visible: true) do
+        fill_in 'Everything', with: 'Tout'
+        click_button 'Save changes'
+      end
+
+      expect(page).to have_css '.nav-pills li.active', text: 'French'
+      expect(page).to have_css '.nav-tabs li.active', text: 'Search field labels'
+    end
+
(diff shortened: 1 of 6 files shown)


spec/features/exhibits/translation_editing_spec.rb
```diff
@@ -71,6 +71,17 @@
...
 
+    it 'redirects to the same form tab' do
+      click_link 'Search field labels'
+      within('#search_fields', visible: true) do
+        fill_in 'Everything', with: 'Tout'
+        click_button 'Save changes'
+      end
+
+      expect(page).to have_css '.nav-pills li.active', text: 'French'
+      expect(page).to have_css '.nav-tabs li.active', text: 'Search field labels'
```
camillevilla: @mejackreed what do you think about this (a separate test for redirects) vs. throwing in three one-liner expectations? e.g. adding `expect(page).to have_css '.nav-tabs li.active', text: 'Browse categories'` to the existing tests for: 
- Field-based search fields
- facet fields
- sort fields

mejackreed: I think thats a great idea. One thing that might be nice to keep is the no javascript on the added tests.

camillevilla: Neat. I've moved the spec for the Browse categories tab out of the JS tests.

Issue #1972, opened 2018-04-03
Spotlight-specific locale files?


ggeisler:
For the current i18n work we're relying on the Blacklight locale files for translation of UI labels. The Blacklight locale files cover the majority, but not quite all, of the labels the user sees in an exhibit UI. (Bolded words in the lists below are the actual labels that would need translations.)

Do we need to supplement the Blacklight locale files with Spotlight locale files to account for any of the labels below?

### Not covered but visible to all users
- **Contacts** sidebar heading on About page (when the exhibit has one or more published contacts)
- Mouseover titles for **Gallery**, **Slideshow**, and **Masonry** view option buttons on search results page (when those views are enabled). These are from https://github.com/projectblacklight/blacklight-gallery
- Mouseover title for **Map** view option button. This is from https://github.com/sul-dlss/blacklight_heatmaps
- **Feedback** brandbar menu item
- Feedback form (**Contact Us**, **Name**, **Email**, **Message**, **Cancel**, **Send**)

### Not covered but possibly out of scope
We've deemed the administration UI out of scope for this first round of i18n work so perhaps these could be considered out of scope, although they aren't actually in the administration UI itself:

- **Save this search** button on search results page (not in administration, but the user must be a curator or administrator to see the button)

Logged-in user menu (curator users)
* **Exhibit dashboard**
* **Change password**
* **Sign out**

Logged-in user menu (admin users)
* **Site administration**
* **Create new exhibit**

anarchivist:
> Mouseover title for Map view option button. This is from https://github.com/sul-dlss/blacklight_heatmaps

I think this is from `blacklight-maps` instead.

anarchivist:
These are:

**Blacklight-Gallery**
```
blacklight.search.view.gallery
blacklight.search.view.masonry
blacklight.search.view.slideshow
blacklight_gallery.catalog.document_slideshow.header
blacklight_gallery.catalog.grid_slideshow.missing_image
blacklight_gallery.catalog.modal_slideshow.counter
```

**Blacklight-Maps**
```
blacklight.maps.title
blacklight.search.view.maps
```

**Spotlight**
```
activemodel.attributes.spotlight/contact_form.email
activemodel.attributes.spotlight/contact_form.message
activemodel.attributes.spotlight/contact_form.name
cancel
helpers.action.cancel
submit.contact_form.create
submit.contact_form.created
spotlight.about_pages.contacts.header
spotlight.header_links.contact
[...]

PR #1973, opened 2018-04-03
[WIP] Add i18n tasks to help aid in assessing translations available

mejackreed:
Tasks available in `i18n-tasks` may be helpful in #1972 and other work.


jkeck:
Wondering if we should try and add this after we merge the i18n-🌝branch since this is doing some normalization of the i18n keys (and the spotlight.en.yml might be really difficult to merge)

PR #1976, opened 2018-04-03
disable jQuery animations when Rails.env = test

mejackreed:


app/views/layouts/spotlight/spotlight.html.erb
```diff
@@ -27,6 +27,7 @@
     <%= description %>
     <%= twitter_card %>
     <%= opengraph %>
+    <%= javascript_tag '$.fx.off = true;' if Rails.env.test? %>
```
mejackreed: I don't know where the best place to interject this would be. We could also pop it into a `*.js.erb` thing?

cbeer: Yeah, I wondered about that. I think we did `.js.erb` elsewhere, but I can't find it now.

mejackreed: Seems like this is a pattern that others use?

https://github.com/thoughtbot/suspenders/blob/master/templates/_javascript.html.erb#L5-L10
https://github.com/thoughtbot/suspenders/pull/705/files

Issue #1974, opened 2018-04-03
Address various test suite performance/consistency issues


jkeck:
Potential areas of improvement identified 

* Convert some feature tests to unit tests
   * Some feature specs could probably be view specs, etc.
   * ZPR link!! 
* Analyze slowest tests, classify types of tests for refactoring
* Explore options for mocking responses from our infrastructure (webmock)
* Cut out some Sir Trevor testing if it doesn’t directly pertain to our code?
* Swarm the top 2 or 3 sir-trevor-related tests that cause the most trouble
* Cut down on instances of FactoryBot.create
* Remove “sleep”s from tests (async), see options for killing animations in embed

mejackreed:
Slowest tests on travis:

```
Top 10 slowest example groups:
  About page
    16.2 seconds average (32.4 seconds / 2 examples) ./spec/features/about_page_spec.rb:1
  Slideshow
    12.96 seconds average (12.96 seconds / 1 example) ./spec/features/slideshow_spec.rb:1
  Featured Pages Blocks
    12.74 seconds average (25.49 seconds / 2 examples) ./spec/features/javascript/blocks/featured_pages_block_spec.rb:2
  Link to Search Block
    11.76 seconds average (23.52 seconds / 2 examples) ./spec/features/javascript/blocks/link_to_search_block_spec.rb:1
  Featured Browse Category Block
    11.65 seconds average (23.3 seconds / 2 examples) ./spec/features/javascript/blocks/featured_browse_categories_block_spec.rb:2
  Search Result Block
    11.42 seconds average (11.42 seconds / 1 example) ./spec/features/javascript/blocks/search_result_block_spec.rb:2
  Multi image selector
    6.58 seconds average (6.58 seconds / 1 example) ./spec/features/javascript/multi_image_select_spec.rb:1
  Horizontal rule block
    4.78 seconds average (4.78 seconds / 1 example) ./spec/features/javascript/blocks/rule_block_spec.rb:1
  Uploaded Items Block
    4.65 seconds average (9.29 seconds / 2 examples) ./spec/features/javascript/blocks/uploaded_items_block_spec.rb:1
  Reindex Monitor
    4.54 seconds average (4.54 seconds / 1 example) ./spec/features/javascript/reindex_monitor_spec.rb:1
```
https://travis-ci.org/projectblacklight/spotlight/builds/361895492

mejackreed:
Instead of pending, should we `skip` tests we know aren't actively going to change?

[...]

PR #1980, opened 2018-04-04
try adding an additional page expectation to address flappyness

mejackreed:
Part of #1974


mejackreed:
It didn't work

sad puppy

PR #1983, opened 2018-04-04
try bundler and sprockets caching

mejackreed:
Discussed here: https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,3 +31,12 @@ script:
   - bundle exec rake
 
 jdk: oraclejdk8
+
+cache:
+  ##
+  # Caching bundler here in a hope to reduce Travis test time
+  bundler: true
+  ##
+  # We don't really know how this will work, but lets see.
+  directories:
+    - .internal_test_app/tmp/cache/assets/sprockets


mejackreed:
Previously we enabled: https://github.com/projectblacklight/spotlight/commit/865607343a4d392c6d5478ac359e2e0a540a840e#diff-354f30a63fb0907d4ad57269548329e3

Then subsequently disabled:
https://github.com/projectblacklight/spotlight/commit/a51a2d752f64885ac673c44f4fc4799ca662d28c#diff-354f30a63fb0907d4ad57269548329e3

/shrug

.travis.yml
```diff
@@ -31,3 +31,8 @@ script:
   - bundle exec rake
 
 jdk: oraclejdk8
+
+cache:
+  bundler: true
+  directories:
+    - .internal_test_app/tmp/cache/assets/test/sprockets
```
jkeck: So I'm a bit unfamiliar w/ how this works.  Will asset fingerprinting bust this cache?  (and maybe we can add a comment here, just so if we try to remove it again, we know what we're potentially getting into).

Thanks!

.travis.yml
```diff
@@ -31,3 +31,12 @@ script:
   - bundle exec rake
 
 jdk: oraclejdk8
+
+cache:
+  ##
+  # Caching bundler here in a hope to reduce Travis test time
+  bundler: true
+  ##
+  # We don't really know how this will work, but lets see.
```
jkeck: 😂

PR #1987, opened 2018-04-05
disabled exhibit title edit in non default language

### LINKED ISSUE #1954: Exhibit Title Validation in non-default languages
In our current implementation, an exhibit admin could technically empty out an exhibit's title (meaning in the standard admin UI's title field) and save the exhibit successfully **if doing so under a non-default locale**.  This would typically throw a validation error since we validate the presence of the title but currently won't in our i18n feature branch because we've made this validation only active under the default locale.

We could do something like disable the title input in the standard admin UI when being viewed in a non-default locale.  We could also look at changing the validation to be more sophisticated.  I'm sure there are other options as well.

mejackreed:
Fixes #1954 

# default locale
screen shot 2018-04-05 at 11 48 01 am
# non default locale
screen shot 2018-04-05 at 11 47 54 am

### DIFF
--- a/app/helpers/spotlight/languages_helper.rb
+++ b/app/helpers/spotlight/languages_helper.rb
@@ -19,5 +19,14 @@ module Spotlight
 
       languages.reject { |language| language.locale.to_s.casecmp(I18n.locale.to_s).zero? }.sort_by(&:to_native)
     end
+
+    ##
+    # Can determine whether the current page is using the application's default
+    # locale
+    # @return [Boolean]
+    def default_language?
+      return true unless params[:locale]
+      params[:locale].to_s == I18n.default_locale.to_s
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


app/helpers/spotlight/languages_helper.rb
```diff
@@ -19,5 +19,14 @@ def locale_selecter_dropown_options
...
     end
+
+    ##
+    # Can determine whether the current page is using the application's default
+    # locale
+    # @return [Boolean]
+    def default_language?
+      return true unless params[:locale]
+      params[:locale].to_s == I18n.default_locale.to_s
+    end
```
jkeck: Is this offering something better than what `I18n.locale == I18n.default_locale`?

I think that would allow us to avoid casting to string and we can remove the guard clause (because we'll always have a locale).

mejackreed: Nope.. I guess I didn't realize that was the same thing

mejackreed: My test doesn't pass now, even when passing in strings as params. 

hmm..

jkeck:
Looks like this'll need to be rebased and have its i18n keys normalized.

PR #1988, opened 2018-04-05
Add the Pages tab/dashboard to the Translations admin

### LINKED ISSUE #1919: Enable curator to provide translations of home, feature, and about page content

jkeck:
Closes #1919

## TODO
- [x] Add link to create a page locale specific page variant from an existing page
- [x] ~~Add checkbox for updating the publish state of  translated pages~~ #2001
- [x] Add `Edit`, `Recreate`, and `Delete` links to the Actions column
- [x] Ensure that the progress indicator in the tab is being updated appropriately
- [x] ~~Handle automatic Home Page creation for non-default locales~~ #2002
- [x] Add "Outdated page warning"
- [x] Ensure that the feature page link/dropdown in Main Navigations renders for the appropriate locale.
- [x] Handle how to switch from different language pages using the locale switcher (if we're allowing for slugs to differ between translated pages)
- [x] ~~Potentially write a migration if we're changing friendly-id scopes for pages.~~ #2003
- [x] Address data synchronization issues of weight & parent relationship
- [x] Address creating a translated parent page where the translated child already exists.

## Questions
- [x] Can/should we fall-back to english pages when a translated version has not been created/published?  (And if so, does this mean we don't necessarily need to have the special behavior around the home page)
- [x] Where should the user be redirected after clicking the `Create one now` link?  To the page edit screen or back to the pages translation dashboard?
- [x] Should non-default locale exhibit slugs be internationalized?
  - _(e.g. Should the URL to a Spanish translated page that derived from `/en/exhibit/page-title` be `/es/exhibit/page-title` or should we support `/es/exhibit/titulo-pagina`)_
- [x] Data synchronization: should we sync weights and/or parent relationships from the default locale page to the translated pages (otherwise there will be no way to re-order/update hierarchy of translated pages)

## Progress so far
pages-dashboard


ggeisler:
Looking good @jkeck! 

> Can/should we fall-back to english pages when a translated version has not been created/published? (And if so, does this mean we don't necessarily need to have the special behavior around the home page)

This is certainly up for debate but my thought was that in the case of pages, no, we shouldn't fall back to the English page if no translated page has been created (besides the home page). This approach fits with the idea that when it comes to pages, the curator can create a very different exhibit in the translated language from the default language (widgets can be different, some pages might not be shown in one language or the other, etc.). And if the curator did want to intentionally show the English as a fall-back page they could just create the translated page and not translate anything in it.

> Where should the user be redirected after clicking the Create one now link? To the page edit screen or back to the pages translation dashboard?

Let's try keeping them on the pages translation tab. It's true the curator might often want to edit the page they just created, but it might be annoying and disorienting to be dropped into a page in edit mode if that was not their intended workflow. The edit action is right there in the dashboard so hopefully if editing immediately is the intended workflow, it will be pretty obvious and quick to proceed.

> Should non-default locale exhibit slugs be internationalized?

I don't have a strong opinion either way. I wonder what others think? I hadn't expected the slug to be internationalized because that seemed complicated, but now that I think about it we just make slugs from the page title so I guess it wouldn't be that difficult?

app/controllers/spotlight/about_pages_controller.rb
```diff
@@ -5,6 +5,10 @@ class AboutPagesController < PagesController
     load_and_authorize_resource through: :exhibit, instance_name: 'page'
     before_action :attach_breadcrumbs, except: [:update_contacts]
 
+    def clone
+      super
```
jkeck: I don't understand why we have to do this, but otherwise we get an undefined action error (or something along those lines).  Would love to get an extra pair of eyes on this.

app/controllers/spotlight/home_pages_controller.rb
```diff
@@ -29,11 +29,20 @@ def show
...
+    end
+
     private
 
     alias search_action_url exhibit_search_action_url
     alias search_facet_url exhibit_search_facet_url
 
+    def load_locale_specific_page
+      # Can we infer the resource name here via cancan?
+      @page = Spotlight::HomePage.for_locale.find_by(exhibit: current_exhibit)
```
jkeck: Not sure if I should worry about this or not...

app/controllers/spotlight/pages_controller.rb
```diff
@@ -85,6 +89,20 @@ def update_all
       redirect_back fallback_location: spotlight.exhibit_dashboard_path(@exhibit), notice: notice
     end
 
+    def clone
+      destroy_pre_existing_page_for_locale
```
jkeck: Happy to do this in a different way.  This is basically what gives us the `Recreate` functionality.

app/controllers/spotlight/pages_controller.rb
```diff
@@ -149,5 +181,17 @@ def breadcrumb_to_exhibit_root(key)
     def page_params
       params.require(controller_name.singularize).permit(allowed_page_params)
     end
+
+    def redirect_page_to_related_locale_version
```
jkeck: This is how we're handling if the translated page and default language page slugs change (which is likely given that the slug is derived from the title).  Happy to look into alternatives here too.

app/models/spotlight/page.rb
```diff
@@ -93,5 +101,26 @@ def should_display_title?
...
+      translated_pages.for_locale(locale).first
+    end
+
+    def clone_for_locale(locale)
+      dup.tap do |np|
+        np.locale = locale
+        np.default_locale_page = self
+        np.published = false
+
+        if !top_level_page? && (parent_translation = parent_page.translated_page_for(locale)).present?
```
jkeck: I could be easily convinced to find a better abstraction here.

app/models/spotlight/page.rb
```diff
@@ -93,5 +101,26 @@ def should_display_title?
...
+          np.parent_page = parent_translation
+        end
```
jkeck: NOTE: This requires that the parent translated page be created before the child page.  We could also do the inverse where we check when a top level page is created if there have been any child pages for the given language created that are associated with the default language page (meaning that the child page had been created before the parent page).  Not sure what folks think about this or if this is just the wrong approach in general.

PR #2020, opened 2018-04-23
add page migration task to new locales

### LINKED ISSUE #2003: Write a data migration of some sort (rake task?) for new page slug scope
Broken out of #1988 

We're updating the scope of page slugs to include the locale (that way many pages in an exhibit can have the same slug as long as they have different locales) and this scope is stored in the database.  We should provide some code that can help people migrate to the version of Spotlight w/ internationalization support.

mejackreed:
Fixes #2003 

A few questions here...

 - [x] Should this be for all exhibits? .. yep
 - [x] ~~Should we limit this to a certain type of page? (e.g `where(locale: nil)` ) or something similar?~~ ... yep.. and its really just setting the `FriendId::Slug`

### DIFF
--- a/lib/migration/page_language.rb
+++ b/lib/migration/page_language.rb
@@ -1,0 +1,27 @@
+module Migration
+  ##
+  # This migration sets the Spotlight::Page locale to the I18n.default_locale.
+  # Needed for migrating Exhibits to internationalization work.
+  class PageLanguage
+    def self.run
+      new.run
+    end
+
+    def initialize; end
+
+    def run
+      migrate_pages
+    end
+
+    private
+
+    def migrate_pages
+      FriendlyId::Slug.where(sluggable_type: 'Spotlight::Page').find_each do |slug|
+        unless /locale:\w+/ =~ slug.scope
+          slug.scope += ",locale:#{I18n.default_locale}"
+          slug.save
+        end
+      end
+    end
+  end
+end
--- a/lib/tasks/spotlight_tasks.rake
+++ b/lib/tasks/spotlight_tasks.rake
@@ -69,6 +69,12 @@ namespace :spotlight do
     end
   end
 
+  desc 'Migrate page\'s FriendlyId::Slug to a scoped language'
+  task migrate_pages_language: :environment do
+    require 'migration/page_language'
+    Migration::PageLanguage.run
+  end
+
   def prompt_to_create_user
     Spotlight::Engine.user_class.find_or_create_by!(email: prompt_for_email) do |u|
       puts 'User not found. Enter a password to create the user.'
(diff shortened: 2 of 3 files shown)


lib/migration/page_language.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    attr_reader :language
+
+    private
+
+    def migrate_pages
+      Spotlight::Page.find_each do |page|
+        page.locale = language
+        page.save
```
jkeck: Our migration to add the column defines a `default` so _I think_ the existing pages get set to the default locale (I'm happy to verify this though).

I think the thing that will cause issues for us is that the Friendly ID Slug doesn't know about the new locale scope that was added.  I suppose that this would technically address that (although I imagine we could just save/touch the record if my assumption above is true).

Another option would be to update the slugs themselves in the migration (although I obv. haven't really explored that much).  Would be happy to 🍐up on this w/ ya if you're interested.

lib/tasks/spotlight_tasks.rake
```diff
@@ -69,6 +69,16 @@ namespace :spotlight do
     end
   end
 
+  desc 'Migrate Pages to a scoped language'
+  task :migrate_pages_language, [:language] => :environment do |_, args|
+    if args[:language]
+      require 'migration/page_language'
+      Migration::PageLanguage.run args[:language]
+    else
+      STDERR.puts "\nUsage: rake spotlight:migrate_pages_language[language]\n\n Example: rake spotlight:migrate_pages_language[en]\n\n"
```
jkeck: Would there be any use/value to us falling back to `I18n.default_locale` in this case?

lib/migration/page_language.rb
```diff
@@ -0,0 +1,29 @@
+module Migration
+  ##
+  # This migration sets the Spotlight::Page locale to the provided one. Needed
+  # for migrating Exhibits to internationalization work.
+  class PageLanguage
+    def self.run(language)
```
jkeck: 👏 I like wrapping this up into a PORO that can be run from the rake task.

lib/migration/page_language.rb
mejackreed: 🍐 sounds good here. I'm not sure I understand the full implications of the friendly id scope updates.

lib/migration/page_language.rb
```diff
@@ -0,0 +1,27 @@
...
+    def run
+      migrate_pages
+    end
+
+    private
+
+    def migrate_pages
+      FriendlyId::Slug.where(sluggable_type: 'Spotlight::Page').find_each do |slug|
+        unless /locale:\w+/ =~ slug.scope
+          slug.scope += ",locale:#{I18n.default_locale}"
```
jkeck: Only thing I can think of here is we may want to do something like `slug.scope = [slug.scope, "locale:#{I18n.default_locale}"].join(',')` to guard against any blank column.

I would be **extremely** surprised if any of our scope columns for pages where blank, so this is more of a comment than a change request.

PR #2025, opened 2018-04-25
fix rails deprecation warnings

mejackreed:
Updates to fix Rails deprecation warnings


camillevilla:
Woot! Throwing a copy of the error message in here for future find-ability 🔎
```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. 
The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). 
To maintain the current behavior, use `saved_change_to_attribute?` instead. 
```

PR #2035, opened 2018-04-27
Fix async form observer

mejackreed:
Refactor form-observer to be useful when serving javascript as async

This refactor does a few things:
 - Removes the form observer methods from the global state
 - creates this as a singleton jQuery type plugin/module
 - Documents some strange edge cases / handles same event instantiation

Unforuntately this is not compatible to running our tests in an async
way. But manual smoke testing has confirmed this working.

To test, set:
https://github.com/projectblacklight/spotlight/blob/master/app/views/layouts/spotlight/spotlight.html.erb#L24

```diff
- <%= javascript_include_tag "application" %>
+ <%= javascript_include_tag "application", async: true %>
```
You will also need to set:

```diff
# .internal_test_app/config/environments/development.rb
- config.assets.debug = true
+ config.assets.debug = false
```

### DIFF
--- a/app/assets/javascripts/spotlight/form_observer.js
+++ b/app/assets/javascripts/spotlight/form_observer.js
@@ -1,64 +1,86 @@
 Spotlight.onLoad(function() {
-  serializeObservedForms(observedForms());
-});
-// All the observed forms
-function observedForms(){
-  return $('[data-form-observer]');
-}
-// Serialize all observed forms on the page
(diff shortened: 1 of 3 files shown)


jkeck:
I know this isn't in the actual commit message, but should we update the PR's description to remove the info about the logger (for posterity)?

mejackreed:
Nice catch @jkeck . Done

PR #2046, opened 2018-05-01
Add the current locale to the cache key for searches/browse categories.

jkeck:
I believe this is causing browse category titles to be cached and not changed when switching locales on the browse category landing page.


cbeer:
Hm. I'm surprised something like that isn't part of the default.

jkeck:
Most of our caching happens on the admin side, so I don't think we typically see it.

jkeck:
Just tested this out in our stage environment

## Before

before

## After

after

PR #2058, opened 2018-05-04
modifies the about page link_to syntax in sidebar to use consistent f…

### LINKED ISSUE #2057: Translations of about pages are not getting linked to consistently
If I switch to another language (e.g. `pt-BR`) and navigate to the about pages, the pages end up loading in their default languages. This is a case where the original pages and the translations are both published.

35qt6k6udf
xrlruzzrwg

mejackreed:
…eature page link

Without this, and when using path based locale scopes, about page links
will link to the default page language.

Fixes #2057

### DIFF
--- a/app/views/spotlight/about_pages/_sidebar.html.erb
+++ b/app/views/spotlight/about_pages/_sidebar.html.erb
@@ -3,8 +3,8 @@
   <h4 class="nav-heading"><%= current_exhibit.main_navigations.about.label_or_default %></h4>
   <ul class="nav sidenav">
     <% current_exhibit.about_pages.for_locale.published.each do |page| %>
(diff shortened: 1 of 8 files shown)


app/views/spotlight/about_pages/_sidebar.html.erb
```diff
@@ -4,7 +4,7 @@
   <ul class="nav sidenav">
     <% current_exhibit.about_pages.for_locale.published.each do |page| %>
       <li class="<%= 'active' if current_page? [page.exhibit, page] %>">
```
cbeer: Is the `current_page?` still ok?

mejackreed: Good point.. fixed

PR #1336, opened 2015-12-11
Rescue exceptions from current_page_context and current_browse_context

cbeer:

PR #2062, opened 2018-05-07
Use documents returned from #get_previous_and_next_documents_for_sear…

jkeck:
…ch instead of directly from response.

The documents array returned from #get_previous_and_next_documents_for_search handles the nil first/last documents appropriately for previous/next pagination.

https://github.com/projectblacklight/blacklight/blob/v6.15.0/app/controllers/concerns/blacklight/search_helper.rb#L98-L100

Fixes sul-dlss/exhibits#450

## Before
before

## After
after


jkeck:
@cbeer it might be helpful to get your 👀 on this.  This was introduced in #1336 but I believe we may be okay w/ the next or previous document being `nil`.

cbeer:
🤷‍♂️

PR #2063, opened 2018-05-09
Log entries to replace reindexing entries

ives1227:
The changes here are to allow additions to the Spotlight Dashboard for jobs other than reindexing jobs (for example, we use it for our harvester status).  The changes include some renaming from reindexing to job_entry as well as adding a job_type field.  The default job_type is called 'Reindexing' but can be set to any value (such as 'Harvesting').  

There are two migrations associated.  One to rename from reindex to job entry and the other to add the job_type field.


camillevilla:
@ives1227 thanks for contributing! We'll hopefully have time to review this next week.

ives1227:
Going to resubmit after upgrading to newest Spotlight.  This is severely outdated at this point.

PR #2066, opened 2018-05-22
TinyMCE Sir Trevor widget

wwelling:
screen shot 2018-05-22 at 11 24 41 am

screen shot 2018-05-22 at 8 58 24 am


app/assets/javascripts/spotlight/blocks/wysiwyg_block.js
```diff
@@ -0,0 +1,56 @@
...
+
+    onBlockRender: function() {
+      var wb = this;
+      tinyMCE.init({
+        selector: "div.tmce-" + wb.blockID,
+        inline: false,
+        inline_styles: true,
+        menubar: false,
+        plugins: [
+          'advlist autolink lists link charmap print preview hr anchor pagebreak',
```
jkeck: It would be great if this was somehow more easily configurable by adopters.

app/assets/javascripts/spotlight/blocks/wysiwyg_block.js
```diff
@@ -0,0 +1,56 @@
...
+          'searchreplace wordcount visualblocks visualchars code',
+          'insertdatetime nonbreaking table contextmenu directionality',
+          'paste textcolor colorpicker textpattern codesample'
+        ],
+        toolbar: 'undo redo | insert | styleselect | bold italic | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link | forecolor backcolor | table codesample',
```
jkeck: Similarly to above, it would be great if we could make this easily configurable.

wwelling:
I couldn't agree more. It would be great to have the TinyMCE plugins and toolbar configurable, either from file or through UI. Also, it makes sense to have this as an optional install via a generator. As you may know, here at A&M we apply all customizations via a generator. It would be easy enough to produce a TinyMCE generator to enable the WSYIWYG Sir Trevor block. Additionally, we add the mirador gem via a generator. This is a bit more involved requiring database migrations. Let me know how you would like to proceed and I can work on getting some other pull requests in as an alternative to this.

wwelling:
Closing the PR as it has become stale. I don't see any time where I can afford making it configurable as desired.

PR #2067, opened 2018-05-29
Change the content column in the spotlight_pages table from…

jkeck:
…text to mediumtext.

### DIFF
--- a/db/migrate/20180529225807_change_spotlight_pages_content_column_to_medium_text.rb
+++ b/db/migrate/20180529225807_change_spotlight_pages_content_column_to_medium_text.rb
@@ -1,0 +1,5 @@
+class ChangeSpotlightPagesContentColumnToMediumText < ActiveRecord::Migration[5.1]
+  def change
+    change_column :spotlight_pages, :content, :text, limit: 16_777_215
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20180529225807_change_spotlight_pages_content_column_to_medium_text.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeSpotlightPagesContentColumnToMediumText < ActiveRecord::Migration[5.2]
```
cbeer: Just one stupid question; will specifying `[5.2]` here mean that Rails 5.1 won't be able to run the migration?

cbeer: Answer: yes; Rails refuses to run a migration that's tagged with a greater version. I think we need to specify `5.1` instead.

jkeck:
👍

db/migrate/20180529225807_change_spotlight_pages_content_column_to_medium_text.rb
jkeck: Yeah, I just did whatever rails generating for me w/o thinking too much about it. 😳

PR #2078, opened 2018-06-21
Add an easily extendable PageConfigurations object to replace the…

jkeck:
… current hard-coded data hash.

This is intended to allow apps or plugins providing SirTrevor widgets to inject their own configurations into the page.

### DIFF
--- a/app/models/spotlight/page_configurations.rb
+++ b/app/models/spotlight/page_configurations.rb
@@ -1,0 +1,93 @@
+module Spotlight
+  ##
+  # PageConfigurations is a simple class to gather and return all
+  # configurations needed for the various SirTrevor widgets.
+  # A downstream application (or gem) can inject its own data through the Spotlight's engine config like so
+  # Spotlight::Engine.config.page_configurations = {
+  #   'my-key': 'my_val'
+  # }
+  # You can also pass anything that responds to #call (eg. a lambda or custom ruby class)
+  # as the value and it will be evaluted within the PageConfiguration context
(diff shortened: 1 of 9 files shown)


jkeck:
Not married to this pattern/approach, so happy to explore other options.

app/models/spotlight/page_configurations.rb
```diff
@@ -0,0 +1,91 @@
+module Spotlight
+  ##
+  # PageConfigurations is a simple class to gather and return all
+  # configurations needed for the various SirTrevor widgets.
+  # A downstream application can inject its own data like so
+  # Spotlight::PageConfigurations.local_params = {
```
mejackreed: Where is the recommended place for this to happen in a Spotlight application?

jkeck: Most likely in an initializer.  Should I add that to this comment?

mejackreed: Ah gotcha.. sure / maybe.. i was wondering if these were like per widget configs , but I guess they are global which is fine

cbeer: If we're putting them in an initializer, does it make sense to add them as engine configuration, just for the sake of keeping everything in one place?

jkeck: Sure, that makes sense.

jkeck: I've updated the code and relevant comment, let me know if there should be addl. changes.

PR #2080, opened 2018-06-29
Add the page type to the friendly-id scope so that pages of different…

jkeck:
… types can have the same name/slug.

### DIFF
--- a/app/models/spotlight/home_page.rb
+++ b/app/models/spotlight/home_page.rb
@@ -5,7 +5,7 @@ module Spotlight
   # Exhibit home page
   class HomePage < Spotlight::Page
     extend FriendlyId
-    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale] do |config|
+    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale type] do |config|
       config.reserved_words&.concat(%w[update_all])
     end
 
--- a/app/models/spotlight/page.rb
+++ b/app/models/spotlight/page.rb
@@ -9,7 +9,7 @@ module Spotlight
     extend FriendlyId
     # Note: This configuration also needs to be duplicated on the
     # STI models ({Spotlight::AboutPage}, {Spotlight::FeaturePage}, {Spotlight::HomePage})
-    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale], treat_reserved_as_conflict: true do |config|
+    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale type], treat_reserved_as_conflict: true do |config|
       config.reserved_words&.concat(%w[update_all contacts])
     end
 
--- a/lib/migration/add_page_type_to_friendly_id_scope.rb
+++ b/lib/migration/add_page_type_to_friendly_id_scope.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Migration
+  ##
+  # Different types of pages are accessed through their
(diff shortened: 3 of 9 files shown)


app/models/spotlight/home_page.rb
```diff
@@ -3,7 +3,7 @@ module Spotlight
   # Exhibit home page
   class HomePage < Spotlight::Page
     extend FriendlyId
-    friendly_id :title, use: [:slugged, :scoped, :finders], scope: [:exhibit, :locale]
+    friendly_id :title, use: [:slugged, :scoped, :finders], scope: [:exhibit, :locale, :type]
```
jkeck: I don't think we need this here, but it felt nice to have for consistency?

app/models/spotlight/page.rb
```diff
@@ -6,7 +6,7 @@ class Page < ActiveRecord::Base
     MAX_PAGES = 50
 
     extend FriendlyId
-    friendly_id :title, use: [:slugged, :scoped, :finders, :history], scope: [:exhibit, :locale]
+    friendly_id :title, use: [:slugged, :scoped, :finders, :history], scope: [:exhibit, :locale, :type]
```
jkeck: I honestly don't know why we have `friendly_id` on the base class as well as the subclasses but `/shrug`

lib/tasks/spotlight_tasks.rake
```diff
@@ -75,6 +75,12 @@ namespace :spotlight do
     Migration::PageLanguage.run
   end
 
+  desc 'Migrate page\'s FriendlyId::Slug to a scoped page type'
+  task migrate_pages_slug_type: :environment do
+    require 'migration/add_page_type_to_friendly_id_scope'
+    Migration::AddPageTypeToFriendlyIdScope.run
```
jkeck: I'm doing it this way, because if we add a migration in the Spotlight codebase it will get run before Friendly IDs migrations are run and will be problematic for new installs.  Not sure if there is a better way to handle that sort of thing.

app/models/spotlight/page.rb
cbeer: I think it was a problem with friendly_id + STI?

lib/migration/add_page_type_to_friendly_id_scope.rb
```diff
@@ -0,0 +1,33 @@
...
+      slugs.find_each do |slug|
+        next if /type:\w+/ =~ slug.scope
+
+        slug.scope = new_scope(slug)
+        slug.save
+      end
+    end
+
+    def new_scope(slug)
+      (slug.scope.split(',') << new_scope_part(slug)).sort.join(',')
```
jkeck: https://github.com/norman/friendly_id/blob/dbbdd5f202c63909dd5de6426be332cbe832105d/lib/friendly_id/scoped.rb#L121

app/models/spotlight/page.rb
jkeck: Probably.

thatbudakguy: the [friendly_id docs](http://norman.github.io/friendly_id/file.Guide.html#Setting_Up_FriendlyId_in_Your_Model) say:
> A note about single table inheritance (STI): you must extend FriendlyId in all classes that participate in STI, both your parent classes and their children.

jcoyne:
@thatbudakguy do you still care about this work?  Do you want to rebase it? Otherwise can we close it?

thatbudakguy:
Seems like the underlying issue was resolved, so feel free to close it. I don't even remember what this was for

PR #2096, opened 2018-08-16
make store_dir root a configuration option

dunn:


dunn:
Travis timed out trying to download Solr.

dunn:
Could I get someone to restart the build?

jkeck:
I'm not able to rebase this branch.  Can you address the rubocop errors?

dunn:
The weird thing is that I don't see those errors locally, and they're only cropping up on one of the builds, which is why I wanted to re-run Travis first to see if it was a fluke.

jkeck:
Can you try removing your `Gemfile.lock` and your `.internal_test_app`, run `bundle install` and `bundle exec rake` again afterwards.  That will more closely emulate what travis is doing (since it rebuilds from scratch on every build).  I was just able to reproduce that by checking out your fork.

dunn:
Oh, I see it too now, but I'm even seeing it on `master`.  Should probably be a separate PR since it's unrelated.

dunn:
All green now.

app/uploaders/spotlight/featured_image_uploader.rb
```diff
@@ -9,7 +9,7 @@ def extension_whitelist
     end
 
     def store_dir
-      "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
+      "#{Spotlight::Engine.config.upload_dir}/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
```
jkeck: Quick question about this.  It was previously "uploads/...." and was already putting the records in the `public` directory.  Will changing this from "uploads/..." to "public/uploads/..." cause a redundant `public` directory inside the already existing `public` directory, or does CarrierWave figure that out for us?  Just trying to ensure that this wont cause any issues for upgrading existing spotlight apps.

dunn: Previously `public` was hardcoded here, and I removed it when I made the change you're highlighting: https://github.com/projectblacklight/spotlight/pull/2096/files#diff-7d4f437731e1a8742d3acb9a9b78b3c7L54

I think that's the only place that uses `store_dir`, so changing it to explicitly include the `public` base directory should be OK, right?  Are there any other uses that I overlooked?

jkeck: Okay, I just spun up a test app using your fork of spotlight and uploaded an image.  That image ended up in `/Projects/spotlight/dunn-sl/.internal_test_app/public/public/uploads/spotlight/featured_image/image/1` (notice the `/public/public`.

This is what I'm trying to avoid here.  Previously it was just `uploads` (which put an uploads directory in `public`), and not it is `public/uploads` which puts it in a redundant `public` directory within `public`.

dunn:
I think this fixes the `public` problem, but now a JS test is failing:
```
  2) Uploaded Items Block users can toggle individual images to not display
     Got 0 failures and 2 other errors:

     2.1) Failure/Error: raise ActionController::RoutingError, "No route matches [#{env['REQUEST_METHOD']}] #{env['PATH_INFO'].inspect}"

          ActionController::RoutingError:
            No route matches [GET] "/Users/alex/clones/spotlight/public/uploads/spotlight/attachment/file/1/800x600.png"

rspec ./spec/features/javascript/blocks/uploaded_items_block_spec.rb:13 # Uploaded Items Block users can upload images with text
```

It looks like this is happening because the path to the image fixture is expected to also be a valid route: https://github.com/projectblacklight/spotlight/blob/master/spec/features/javascript/blocks/uploaded_items_block_spec.rb#L21

But I don't know enough about rspec or the Spotlight JS code; is this just a matter of rewriting the spec?

dunn:
I'm starting to think CarrierWave just isn't designed for `store_dir` to be customized like this; closing for now.

jkeck:
🙏 ActiveStorage

Issue #1827, opened 2017-08-29
Migrate references column from Strings to integers


jcoyne:
We no longer need strings after: https://github.com/projectblacklight/spotlight/pull/1815

This https://github.com/mbleigh/acts-as-taggable-on/issues/853 is forcing us to make the change.

So we no longer need this custom column: https://github.com/projectblacklight/spotlight/blob/7ab587601351e10a403b01770055c13b646b29cc/db/migrate/20140206152809_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb#L13

We can revert to:
https://github.com/mbleigh/acts-as-taggable-on/blob/f5222d8afaec1337a9d76b27fcc503a9125fc85d/db/migrate/1_acts_as_taggable_on_migration.rb#L17

dunn:
Can https://github.com/sul-dlss/dlme/pull/227 be easily adapted for Spotlight?

dunn:
Or would it be as simple as adding a new migration, like:
```ruby
class ChangeTypeOfTaggingsTaggable < ActiveRecord::Migration[5.2]
  def change
    change_column :taggable_id, :integer
  end
end

Issue #1433, opened 2016-01-29
Postgres type mismatch error with taggable


ronan-mch:
> PG::UndefinedFunction: ERROR: operator does not exist: character varying = integer LINE 1: ...= 'Spotlight::Exhibit') AND (taggings.taggable_id IN(SELECT ... ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.

``` SQL
SELECT COUNT(*) FROM "tags" JOIN (SELECT taggings.tag_id FROM "taggings" WHERE (taggings.taggable_type = 'Spotlight::Exhibit') AND (taggings.taggable_id IN(SELECT spotlight_exhibits.id FROM "spotlight_exhibits" WHERE ('t'='t') AND "spotlight_exhibits"."published" = 't' ORDER BY weight ASC)) GROUP BY taggings.tag_id) AS taggings ON taggings.tag_id = tags.id
```

To reproduce: Use postgres, create multiple exhibitions with a tag, return to home screen. Or from the rails console: `Spotlight::Exhibit.all.all_tags`

ronan-mch:
Seems to arise from this line:
https://github.com/sul-dlss/spotlight/blob/master/app/views/spotlight/exhibits/index.html.erb#L19

which refers to the `all_tags` method from `acts-as-taggable`: https://github.com/mbleigh/acts-as-taggable-on/blob/fac92a9114d20bf71595c1e4b1e8ec622fc2f190/lib/acts_as_taggable_on/taggable/collection.rb#L55

dunn:
This is still present when running 2.2.0 and PSQL 9.2.

dunn:
So it is; @jcoyne noticed the same thing a while back https://github.com/mbleigh/acts-as-taggable-on/issues/853

PR #2100, opened 2018-08-23
db: change taggable_id to integer

dunn:
Fixes GH-1433.
Closes GH-1827.

Tests are failing locally with:
```
== 20180814221815 ChangePaperTrailVersionsObjectColumnToMediumText: migrating =
-- adapter_name()
   -> 0.0000s
-- change_column(:versions, :object, :text, {:limit=>1073741823})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Could not find table 'versions'
/Users/alex/clones/spotlight/.internal_test_app/db/migrate/20180814221815_change_paper_trail_versions_object_column_to_medium_text.rb:9:in `change'
/Users/alex/.rbenv/versions/2.4.4/bin/bundle:23:in `load'
/Users/alex/.rbenv/versions/2.4.4/bin/bundle:23:in `<main>'

### DIFF
--- a/db/migrate/20180822233601_change_type_of_taggings_taggable.rb
+++ b/db/migrate/20180822233601_change_type_of_taggings_taggable.rb
@@ -1,0 +1,8 @@
+class ChangeTypeOfTaggingsTaggable < ActiveRecord::Migration[5.1]
+  def change
+    change_column :taggings,
+                  :taggable_id,
+                  :integer,
+                  using: 'taggable_id::integer'
+  end
+end

### CODE REVIEW

**Comment by dunn** [comment]
Seeing the same error on Travis, even though this is an unrelated migration I'm adding.

**Comment by dunn** [comment]
I'm confused as to why https://github.com/projectblacklight/spotlight/blob/master/lib/generators/spotlight/templates/migrations/20180814221815_change_paper_trail_versions_object_column_to_medium_text.rb is necessary in the first place; it looks like PaperTrail already sets that limit on the specified field: https://github.com/paper-trail-gem/paper_trail/blob/6c34a3dd5a5f8c1b042f458b7727c9d3bbf81a50/lib/generators/paper_trail/install/templates/create_versions.rb.erb#L5-L17

**Comment by mejackreed** [comment]
I'm seeing an error in the travis build with this migration here: https://travis-ci.org/projectblacklight/spotlight/jobs/419446701#L2448

**Comment by dunn** [comment]
Now that's a different error than I'm seeing locally.  Does that syntax not work for sqlite?

**Comment by dunn** [comment]
Turns out the syntax was just wrong.

**Comment by dunn** [comment]
Anything more needed here?

**Comment by jkeck** [comment]
This seems reasonable to me, although I'm not really familiar w/ PG to know if this the way to go to solve the error outlined.  @mejackreed thoughts?

**Comment by mejackreed** [comment]
Thanks @dunn for your patience. This seems reasonable to me as well, especially since acts-as-taggable-on is kind of forcing our hand here. I've pulled this into one of our Spotlight applications and everything seems to be working as expected still.

**Comment by elrayle** [comment]
We had problems with our data where IDs were NULL.  Posting the error we saw when running this migration in case others run into a similar problem.

### ERROR

```
== 20200414165044 ChangeTypeOfTaggingsTaggable: migrating =====================
-- change_column(:taggings, :taggable_id, :integer, {:using=>"taggable_id::integer"})
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Data truncated for column 'taggable_id' at row 7: ALTER TABLE `taggings` CHANGE `taggable_id` `taggable_id` int DEFAULT NULL
```

### Problem Exploration

```
select * from taggings where taggable_id is NULL;
select * from taggings where tagger_id is NULL;
```

We had 5 rows where taggable_id was NULL and another 41 rows where tagger_id is NULL.

It is unclear why these values are NULL.  To resolve this problem, the rows in taggings with the NULL values were deleted.  Then the migration completed without error.

PR #2110, opened 2018-10-23
Fix project build

mejackreed:
Also done upstream in Blacklight


camillevilla:
Changes merged by #2109

PR #2113, opened 2018-11-27
Increase the maximum number of pages

cbeer:
Fixes https://github.com/sul-dlss/vatican_exhibits/issues/320


ggeisler:
1000! I better not mention that to the curator... Thanks.

cbeer:
50 was supposed to be an absurd number, so 🤷‍♂️

PR #2118, opened 2019-01-04
Add UniversalViewer

little9:
This commit adds a `package.json` so that
is possible to install `universalviewer`
with yarn in Rails 5.1+ applications.

To install uv in your development
enviroment run `yarn run install-uv`.

In production there should be a capistrano task that
moves the uv assets to the `/public` folder.

In the catalog controller for the spotlight app
the `universal_viewer_default` partial should be used
instead of `openseadragon`. This
uses the `<iframe>` version of UniversalViewer. It's
done this way because including UV directly is not
compatible with jQuery almond and other JS included
in `spotlight.js`.

Documentation on the options available for the viewer
is here:

https://github.com/UniversalViewer/universalviewer/wiki/Embedding#hash-parameters

There is a `ManifestService` and a `#iiif_manifest` helper that are
used in views.


little9:
This is being used in a spotlight app at:
https://github.com/ucsblibrary/geniza

and can be viewed at:
https://geniza-qa.curationexperts.com/spotlight/the-anna-s-c-blake-manual-training-school/catalog/9-30

little9:
Thanks for the feedback @mejackreed !

Doing it this way wouldn't change any of spotlight defaults. In the spotlight app, you would need to 
change the `catalog_controller` to use the universalviewer view instead of the openseadragon one. (See https://github.com/ucsblibrary/geniza/blob/master/app/controllers/catalog_controller.rb#L13) You'd also need to install uv with `yarn run install-uv`. In production you would need a deploy script to do that as well. (See https://github.com/ucsblibrary/geniza/blob/master/lib/capistrano/tasks/install_uv.rake) So this commit would allow you to install it and use it if you wanted it, but there would need to be additional configuration in the spotlight app. 

My reasoning behind doing it this way was to avoid checking in the UV JS into a repo where it would need to be updated manually. 

Do you have an example of a plugin I can look at for this kind of integration?

mejackreed:
Spotlight itself is a plugin for Blacklight so that general pattern holds. A fairly small, Blacklight plugin that might be similar is [BlacklightHeatmaps](https://github.com/sul-dlss/blacklight_heatmaps).

It has a [custom generator ](https://github.com/sul-dlss/blacklight_heatmaps/blob/master/lib/generators/blacklight_heatmaps/install_generator.rb) which I could I see something similarly being used for Spotlight-UV (or even Blacklight-UV)?

PR #2120, opened 2019-01-09
Generalize job log

ives1227:
This request has a few minor changes to allow for more generalization:
1. Addition of a DB column called 'job_type' which defaults to 'Reindexing'. 
2. Renaming of DB columns from using 'reindexing' to 'job_item'
3. Update of Dashboard view to generalize the area for background jobs to be called 'Item Jobs' instead of 'Reindexing Jobs'.  

This allows other background jobs (or other viable information) to be logged and displayed.  One such example that we use it for is during harvesting of items via OAI (or Solr), it displays the progress of the harvesting.  

New Display:
image


ggeisler:
Thanks for this update @ives1227. I'll leave it to more technical people to comment on most of this but I have a couple minor suggestions related to the UI updates.

I know to a developer "Item job" is very common and descriptive term for what's going on, but the vast majority of people (at least at Stanford) who will view this section of the UI are likely exhibit curators or product owners, to whom "item job" is likely to sound a bit technical or jargony. I suggest:

- Rename the section header "Recent Item Processing Activity"
- Rename the second column header "Activity"

I don't love "item processing" so if you or anyone else has another suggestion that would be great, but I think it's at least a term that will feel more descriptive to non-technical people. Harvesting and reindexing, I think, reasonably fit into the concept of "item processing".

ives1227:
Closing to prep a new one without breaking changes.

PR #2124, opened 2019-01-11
Add a helper for getting a IIIF manifest URL

little9:
This commit adds a service and helper for retrieving the IIIF manifest
from a SolrDocument. Currently trying to access this URL
for an item on a view is kind of awkward:

```
document['response']['response']['docs'][0]['iiif_manifest_url_ssi']
```

With this service and helper, you can access it in views by doing:

```
iiif_manifest(manifest_service: Spotlight::ManifestService.new(document: document))
```

The functionality of the service could be expanded to additional
checking and other useful things in the future.

### DIFF
--- a/app/services/spotlight/manifest_service.rb
+++ b/app/services/spotlight/manifest_service.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Spotlight
+  # Service that is initilized with a SolrDocument from an item in spotlight and
+  # returns the IIIF manifest URL
+  class ManifestService
+    attr_reader :document
+
+    # @param document [SolrDocument] a SolrDocument from an item in Spotlight with a IIIF manifest
+    def initialize(document:)
+      @document = document
+    end
+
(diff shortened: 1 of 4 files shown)


little9:
@mejackreed This was in my other UV focused commit, but this is general enhancement that 
I think would be useful for spotlight in general. Could you take a look at this?

app/services/spotlight/manifest_service.rb
```diff
@@ -0,0 +1,19 @@
...
+    attr_reader :document
+
+    # @param document [SolrDocument] a SolrDocument from an item in Spotlight with a IIIF manifest
+    def initialize(document:)
+      @document = document
+    end
+
+    # @return [String] the URL to a IIIF manifest
+    def url
+      document.response.dig(:response, :docs, 0, :iiif_manifest_url_ssi)
```
cbeer: Is this the same as `document[:iiif_manifest_url_ssi]`?

little9: ahh! Thanks

PR #2135, opened 2019-02-12
Upgrade (and pin) rubocop to ~> 0.64.0

jcoyne:


app/models/spotlight/solr_document_sidecar.rb
```diff
@@ -6,9 +6,9 @@ class SolrDocumentSidecar < ActiveRecord::Base
 
     acts_as_taggable
 
-    belongs_to :exhibit, required: true
+    belongs_to :exhibit, optional: false
```
jkeck: Is this addressing a rubocop rule?

jcoyne: Yeah, back around Rails 5 they flipped from `required` to `optional`.

PR #2137, opened 2019-02-26
Update Spotlight to Blacklight 7 + Bootstrap 4

cbeer:
This passes all tests when combined with some upstream PRs in blacklight:

- [x] https://github.com/projectblacklight/blacklight/pull/2075
- [x] https://github.com/projectblacklight/blacklight/pull/2071

More CSS work is needed to make it visually better. There's also some work to update (or rewrite) css for some of the pages that attached additional classes to the `#content` element.


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
-# `rubocop --auto-gen-config --no-auto-gen-timestamp`
```
dunn: It's not important but I do like this flag for minimizing the diff.

jkeck:
There are a few bugs here and there, but we'll begin 🔪'n them as part of our current work cycle.  As it stands, this is a good starting point to push forward with.  A 2.x release branch was created to backport things for release in the 2.x line.

PR #2141, opened 2019-03-05
Allow dots to appear in the identifier

### LINKED ISSUE #2140: Allow periods in the identifier
When I have an object imported with an identifier of `https://gallica.bnf.fr/ark:/12148/bpt6k91073505`
Then when visit the index page and I see the object, and I click on it, I get:
```
No route matches [GET] "/library/catalog/https:%2F%2Fgallica.bnf.fr%2Fark:%2F12148%2Fbpt6k91073505"
```

I believe this is because the identifier regex that rails uses by default doesn't include `.`. This can be fixed by adding to https://github.com/projectblacklight/spotlight/blob/master/config/routes.rb#L54:

```ruby
ALLOW_DOTS = /([^\/])+?/
resources :solr_documents, except: [:index], path: '/catalog', controller: 'catalog', constraints: { id: ALLOW_DOTS } do
  ...
```

jcoyne:
Fixes #2140

### DIFF
--- a/blacklight-spotlight.gemspec
+++ b/blacklight-spotlight.gemspec
@@ -20,7 +20,7 @@ these collections.)
   s.required_ruby_version = '~> 2.3'
 
   s.add_dependency 'rails', '~> 5.1'
-  s.add_dependency 'blacklight', '~> 6.3'
+  s.add_dependency 'blacklight', '~> 6.20'
   s.add_dependency 'autoprefixer-rails'
   s.add_dependency 'cancancan'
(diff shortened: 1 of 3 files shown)


cbeer:
I'm pretty 👎 on hard-coding it (IIRC, it breaks some things, like using the format extensions in URLs?); can this be configurable, perhaps?

jcoyne:
@cbeer I've added configuration for this. Does this work for you?

jcoyne:
We need blacklight 6 to support this too.

PR #2144, opened 2019-03-07
Encapsulate the Spotlight config

jcoyne:
All values set on Spotlight::Engine.config are actually stored
in a class variable on Railtie::Configuration. So these pollute
that namespace.  All other railties are encapsulating their configs
onto a single entry in that scope.  This change makes Spotlight do
likewise.

This probably needs to be a major revision

### DIFF
--- a/app/components/spotlight/analytics/dashboard_component.rb
+++ b/app/components/spotlight/analytics/dashboard_component.rb
@@ -22,11 +22,11 @@ module Spotlight
       end
 
       def min_date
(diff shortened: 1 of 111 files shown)


hackartisan (commented):
We looked at this in a PR review ensemble. We're interested in whether this could be made more backwards compatible.

- could we check first for the spotlight namespaced config, then fall back if it's not found?
- could we use an initializer to pull the configs we expect to find into a spotlight namespace?
- other ideas?

PR #2146, opened 2019-03-12
Validate the FeaturedImage object if provided.

cbeer:
It seems that, with optional associations, if an associated object is invalid it'll just get dropped. For uploaded objects, if the image is invalid we want to alert the user and cancel the upload.

### DIFF
--- a/app/models/spotlight/resources/upload.rb
+++ b/app/models/spotlight/resources/upload.rb
@@ -4,7 +4,7 @@ module Spotlight
     ##
     # Exhibit-specific resources, created using uploaded and custom fields
     class Upload < Spotlight::Resource
-      belongs_to :upload, class_name: 'Spotlight::FeaturedImage', optional: true
+      belongs_to :upload, class_name: 'Spotlight::FeaturedImage', optional: true, validate: true
 
       # we want to do this before reindexing
       after_create :update_document_sidecar


app/models/spotlight/resources/upload.rb
```diff
@@ -4,7 +4,7 @@ module Resources
     ##
     # Exhibit-specific resources, created using uploaded and custom fields
     class Upload < Spotlight::Resource
-      belongs_to :upload, class_name: 'Spotlight::FeaturedImage', optional: true
+      belongs_to :upload, class_name: 'Spotlight::FeaturedImage', optional: true, validate: true
```
jcoyne: You might want to drop a comment in here about why we need to validate the upload.

cbeer:
```suggestion
     # Force Rails to validate uploads in order to report errors back to the user.
      belongs_to :upload, class_name: 'Spotlight::FeaturedImage', optional: true, validate: true
```

cbeer: @jcoyne : 🤷‍♂️ ?

jcoyne: Ugh. I thought that got into this PR.

PR #2151, opened 2019-03-28
teach Spotlight how to apply themes site-wide

dunn:
Exhibit-level theme settings override the global theme setting; when an exhibit has no theme set it will inherit from the site.

TODO update wiki


dunn:
🍏

mcritchlow:
@mejackreed - I know you all are busy, but any chance we can get a review on this today or early next week?

Having this  upstream would unblock a number of theming tasks we have going in our sprints with UCSD/UCSB. Thanks!

mejackreed:
Yep its on my list today. One thing I'm thinking through is the backwards compatibility piece and if this will require a major version release and if there are ways to mitigate that.

mcritchlow:
@mejackreed - Thank you! Yeah that makes perfect sense.

app/helpers/spotlight/application_helper.rb
```diff
@@ -135,17 +135,24 @@ def available_view_fields
       current_exhibit.blacklight_configuration.default_blacklight_config.view.to_h.reject { |_k, v| v.if == false }
     end
 
-    def exhibit_stylesheet_link_tag(tag)
-      if current_exhibit_theme && current_exhibit.theme != 'default'
-        stylesheet_link_tag "#{tag}_#{current_exhibit_theme}"
+    def themed_stylesheet_link_tag(tag)
```
mejackreed: The method name change here is a breaking change. Could we provide backwards compatibility and deprecation warnings so adopters know how to update?

app/helpers/spotlight/application_helper.rb
```diff
@@ -135,17 +135,24 @@ def available_view_fields
...
+        stylesheet_link_tag "#{tag}_#{current_theme}"
       else
-        Rails.logger.warn "Exhibit theme '#{current_exhibit_theme}' not in white-list of available themes: #{Spotlight::Engine.config.exhibit_themes}"
+        Rails.logger.warn "Exhibit theme '#{current_theme}' not in white-list of available themes: #{Spotlight::Engine.config.exhibit_themes}"
         stylesheet_link_tag(tag)
       end
     end
 
-    def current_exhibit_theme
-      current_exhibit.theme if current_exhibit && current_exhibit.theme.present? && Spotlight::Engine.config.exhibit_themes.include?(current_exhibit.theme)
```
mejackreed: Similarly the method name change here would be breaking. Could you provide backwards compatibility and deprecation warnings?

app/helpers/spotlight/main_app_helpers.rb
```diff
@@ -37,17 +37,24 @@ def presenter(document)
       end
     end
 
-    def exhibit_stylesheet_link_tag(tag)
-      if current_exhibit_theme && current_exhibit.theme != 'default'
-        stylesheet_link_tag "#{tag}_#{current_exhibit_theme}"
+    def themed_stylesheet_link_tag(tag)
```
mejackreed: Similarly the method name change here would be breaking. Could you provide backwards compatibility and deprecation warnings?

app/helpers/spotlight/main_app_helpers.rb
```diff
@@ -37,17 +37,24 @@ def presenter(document)
...
       else
-        Rails.logger.warn "Exhibit theme '#{current_exhibit_theme}' not in white-list of available themes: #{Spotlight::Engine.config.exhibit_themes}"
+        Rails.logger.warn "Exhibit theme '#{current_theme}' not in white-list of available themes: #{Spotlight::Engine.config.exhibit_themes}"
         stylesheet_link_tag(tag)
       end
     end
 
-    def current_exhibit_theme
-      current_exhibit.theme if current_exhibit && current_exhibit.theme.present? && Spotlight::Engine.config.exhibit_themes.include?(current_exhibit.theme)
+    def current_theme
```
mejackreed: Similarly the method name change here would be breaking. Could you provide backwards compatibility and deprecation warnings?

app/views/spotlight/sites/edit.html.erb
```diff
@@ -15,6 +15,17 @@
           <div class="row col-md-12">
             <%= f.text_field :title %>
             <%= f.text_field :subtitle %>
+            <% if Spotlight::Engine.config.exhibit_themes.many? %>
+              <p class="instructions"><%= t(:'.site_theme.help') %></p>
+              <%= f.form_group :theme, label: { text: t(:'.site_theme.label') } do %>
+                <% Spotlight::Engine.config.exhibit_themes.each do |theme| %>
+                  <div class="col-md-6">
+                    <%= image_tag "spotlight/themes/#{theme}_preview.png", width: 100, height: 100 %>
+                    <%= f.radio_button :theme, theme, label: t(:".site_theme.#{theme}", default: theme.to_s), inline: true %>
```
mejackreed: Is it possible `titleize` the string here?

mcritchlow:
@mejackreed happy to do a call today to discuss if useful.

app/helpers/spotlight/application_helper.rb
```diff
@@ -136,16 +136,35 @@ def available_view_fields
     end
 
     def exhibit_stylesheet_link_tag(tag)
-      if current_exhibit_theme && current_exhibit.theme != 'default'
-        stylesheet_link_tag "#{tag}_#{current_exhibit_theme}"
+      Deprecation.warn 'exhibit_stylesheet_link_tag has been deprecated and will be removed.  '\
```
mejackreed: I've noticed that these deprecation warnings are not using the supplied message. I think we may need to pass an instance of `self` to them.

```ruby
      Deprecation.warn self, 'exhibit_stylesheet_link_tag has been deprecated and will be removed.  '\
```

Could we also pass the message that this will be removed in Spotlight 3.0?

app/helpers/spotlight/application_helper.rb
```diff
@@ -136,16 +136,35 @@ def available_view_fields
     end
 
     def exhibit_stylesheet_link_tag(tag)
-      if current_exhibit_theme && current_exhibit.theme != 'default'
```
mejackreed: This seems to me to be another backwards compatibility issue.

Current behavior:

If no theme is configured, then it uses `application.css` for the default theme

New behavior:

If no theme is configured, tries to use `application_default.css`

dunn: That shouldn't be happening.  This test is supposed to ensure that it's still `application.css`: https://github.com/projectblacklight/spotlight/pull/2151/files#diff-e1e92e3b842cf6d2c8b95afc21c06044R71

How can I reproduce what you're seeing?

mejackreed: Here are the steps I'm following:

1. `rm -rf .internal_test_app`
1. Recreate a new spotlight app
1. Configure two themes 

```ruby
# .internal_test_app/config/initializers/spotlight_initializer.rb
Spotlight::Engine.config.exhibit_themes = ['default', 'fancy']
```

Then after starting up application and choosing a site theme as "default", no css is found to render that.

Screen Shot 2019-04-16 at 12 53 24 PM

dunn: One thing I did that probably just ended up being confusing was to make `default` be the name of a theme; it's doesn't mean that no theme is used.  I did that because (as far as I can see), once you select a theme there's no way to revert to using no theme.  So I added a reference to a fallback 'default' theme as a workaround.

I can start working on the root problem (letting admins un-theme exhibits), but I'd rather that be a subsequent PR.

dunn:
renamed the test themes for clarity cc @mejackreed

mejackreed:
Thanks for submitting this pull request and the fixes to previous comments.

We are going to suggest an alternative approach now to enable similar functionality in your Spotlight site that hopefully preserves the intent of the approach in this PR and gives you similar functionality. There are a few reasons for this, but the main one is to unblock this type of work so Surfliner can move forward.

We suggest the following:

1. Enable per-deployment `config/initializers/spotlight_initializer.rb` settings that could be configured using a deployment settings, config, or environment variable.

```diff
# https://gitlab.com/surfliner/surfliner/blob/master/starlight/config/initializers/spotlight_initializer.rb#L3
# CONFIGURED_THEMES=default,ucsb
# DEFAULT_THEME=ucsb
- Spotlight::Engine.config.exhibit_themes = %w[default ucsb ucsd]
+ Spotlight::Engine.config.exhibit_themes = ENV["CONFIGURED_THEMES"].split(',')
```

2. Go ahead and override `app/views/layout/spotlight/spotlight.html.erb#L19-L23` in the Surfliner codebase. Provide some custom logic there to select the default configured theme for the the deployment.

```diff
   <% if current_exhibit %>
      <%= exhibit_stylesheet_link_tag "application" %>
    <% else %>
-       <%= stylesheet_link_tag "application" %>
+      <%= stylesheet_link_tag "application<%= ENV["DEFAULT_THEME"] %>" %>
    <% end %>
```

3. Submit a PR to Spotlight that extracts the overriden stylesheet logic into a helper (similar to #2151) so that you can rely on overriding that helper rather than the layout long term.

Reasons why we want to start with this approach:
1. This hopefully unblocks your team ASAP.
1. We are still trying to understand if there are applications for site wide themes that aren’t institutionally based, and other community uses for this feature.
1. This approach preserves compatibility with a future site-wide theme feature if that happens
1. Gary found some enhancements to be made in the Site administration homepage in this PR. We will add these in an issue for future work.

dunn:
@mejackreed I don't follow.  Are you suggesting that my existing implementation be updated to use environment variables instead of hardcoding new themes into the initializer?  If so, that's a good idea and I can do it.

Or do you mean that the entire PR should be replaced with a minimal patch that uses an environment variable to change what CSS is rendered, without touching the `Site` model as I have?  If that's what you mean then if multiple themes are passed in via environment variables, there's no way to tell Spotlight which to use.

mejackreed:
> Are you suggesting that my existing implementation be updated to use environment variables instead of hardcoding new themes into the initializer? If so, that's a good idea and I can do it.

Yes this one.

dunn:
Updated to use the environment variable `SPOTLIGHT_THEMES`.  Will update the wiki accordingly.

mejackreed:
Sorry, I think we may have cross communicated here. I assumed when you said "existing implementation" you meant your application, Surfliner.

no-reply:
@mejackreed thanks for the consideration.

Could I request a meeting with folks on the project team to chat about what the best way for our project to contribute might be? We've been aiming to avoid owning a wildly divergent codebase, instead working to distill our requirements into community-focused contributions. Clearly, this isn't working out the way we'd hoped, and I could use some guidance understanding how our team should proceed.

mejackreed:
@no-reply sure thing

Shoot me an email @ pjreed@stanford.edu and let's set it up.

PR #1986, opened 2018-04-05
Update RIIIF to 2.0

jcoyne:

PR #2152, opened 2019-03-28
travis: run beta chrome

dunn:
Hopefully will fix https://github.com/projectblacklight/spotlight/pull/1986#issuecomment-477765303


dunn:
I am defeated; @jcoyne maybe we should just drop 5.1.

Issue #1678, opened 2017-01-30
Items in a deleted exhibit persist


sfdi:
Items in an exhibit persist when the exhibit is deleted. If a "new" exhibit is then created with the same URL slug, those items appear in that exhibit. Is this expected behavior?

escowles:
Relevant here: https://github.com/pulibrary/pomegranate/pull/204 (deleting items from Solr when an Exhibit is deleted, which would avoid this issue)

sfdi:
@cbeer in your comment [here](https://github.com/pulibrary/pomegranate/pull/204/commits/35e9400aa5f90ebedb794c447daa010eea0172f8#r98511158) did you mean *upstream* model? To my knowledge if we want an exhibit deletion to cascade to most of its assets, there are a number of changes to those models that would be necessary.

PR #2153, opened 2019-03-28
When a resource is deleted, remove its index object too

jcoyne:


app/models/spotlight/resource.rb
```diff
@@ -96,6 +98,18 @@ def commit
         Rails.logger.warn "Unable to commit to solr: #{e}"
       end
 
+      # Called when the record is destroyed.
+      def remove_from_index
+        return unless write?
+
+        blacklight_solr.delete_by_query(
+          "#{Spotlight::Engine.config.resource_global_id_field}:#{to_global_id}"
```
cbeer: I think there's probably another necessary component to this to deal with solr documents that are generated into the index from multiple resources. To start with, we may need to do additional bookkeeping to track resources -> solr ids in the database.

jcoyne: So, we need to check that resource_global_id_field is only a single element and that single element is this resource?

cbeer: And do the work to make sure `resource_global_id_field` is multi-valued when necessary.

hackartisan:
This pull request is currently inactive and is being closed. If you would like to discuss this work, please discuss on a new or existing issue. Thank you!

chrisrlc:
Related to currently open issues: https://github.com/projectblacklight/spotlight/issues/3516 and https://github.com/projectblacklight/spotlight/issues/1678

PR #2158, opened 2019-04-01
Setup SpotlightAnalytics to be able to set anonymizeIp for Google Ana…

mejackreed:
…lytics in configuration


lib/spotlight/engine.rb
```diff
@@ -189,6 +189,7 @@ def self.blacklight_config
     config.ga_email = nil
     config.ga_analytics_options = {}
     config.ga_page_analytics_options = config.ga_analytics_options.merge(limit: 5)
+    config.ga_anonymize_ip = false # false for backwards compatibility
```
jcoyne: should this also be added to the template configuration?

mejackreed: Probaby.. just rebased and pushed up this change. Nice catch!

PR #2164, opened 2019-04-24
render_document_class: avoid modifying frozen strings

dunn:

PR #2167, opened 2019-05-13
render_document_class: avoid modifying frozen strings

cbeer:
Includes #2164, with updates to work with latest chrome + chromedriver-helper.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ matrix:
 before_install:
   - gem update --system
   - gem install --no-document bundler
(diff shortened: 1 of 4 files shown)


dunn:
Maybe we should drop 5.1 from the test matrix?
```
     Selenium::WebDriver::Error::WebDriverError:
       Unable to find chromedriver. Please download the server from
       https://chromedriver.storage.googleapis.com/index.html and place it somewhere on your PATH.
       More info at https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver.

PR #2176, opened 2019-06-20
Add the document's csrf token to the AJAX request for uploading images.

jkeck:
This seems to handle an issue in some spotlight installations where we're getting an invalid authenticity token error when uploading images using for cropping in thumbnails/mastheads.

See sul-dlss/vatican_exhibits#415

### DIFF
--- a/app/assets/javascripts/spotlight/crop.es6
+++ b/app/assets/javascripts/spotlight/crop.es6
@@ -245,6 +245,8 @@ export default class Crop {
   }
 
   uploadFile() {
+    // Set a ujs adapter to support both rails-ujs and jquery-ujs
+    var ujs = typeof Rails === 'undefined' ? $.rails : Rails;
     var url = this.fileInput.data('endpoint')
     // Every post creates a new image/masthead.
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/spotlight/crop.es6
```diff
@@ -238,9 +238,11 @@ export default class Crop {
   }
 
   // Get all the form data with the exception of the _method field.
+  // Append the document's authenticity token to the form data.
   getData() {
     var data = new FormData(this.form[0]);
     data.append('_method', null);
+    data.append('authenticity_token', $('meta[name="csrf-token"]').prop('content'));
     return data;
```
cbeer: In the sir-trevor code, we use `ujs.csrfToken()` to grab the CSRF token.

cbeer: Oh, and it looks like we might set the auth token in a request header?

```
      headers: {
        'X-CSRF-Token': ujs.csrfToken() || ''
      },
```

jkeck: 👌

PR #2178, opened 2019-06-21
Fixes an issue where translated pages of pages that have UUIDs from F…

mejackreed:
…riendlyID have the incorrect slug

FriendlyID "conveniently" removes the slug for us.

Fixes https://github.com/sul-dlss/vatican_exhibits/issues/411

### DIFF
--- a/app/models/spotlight/page.rb
+++ b/app/models/spotlight/page.rb
@@ -135,6 +135,7 @@ module Spotlight
         np.locale = locale
         np.default_locale_page = self
         np.published = false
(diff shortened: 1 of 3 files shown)


jkeck:
dafuq

mejackreed:
A quick note for future finders of this issue. This fix will only prevent the problem from happening in the future. You will need to manually remediate the data by setting the translated page's slug to its `default_locale_page`'s slug.

Fun! I know :P

PR #2181, opened 2019-07-12
role management: don't apply 'pending' when invite_pending? is false

dunn:


app/views/spotlight/admin_users/index.html.erb
```diff
@@ -13,7 +13,9 @@
           <tr>
             <td class="<%= 'invite-pending' if user.invite_pending? %>">
               <%= user.email %>
-              <span class='label label-warning pending-label'><%= t('.pending') %></span>
+              <% if user.invite_pending? %>
```
jkeck: I believe we're attempting to handle this w/ css (w/ the assumed benefit being we can easily toggle a class w/ js if this we're dealing w/ this client side).

https://github.com/projectblacklight/spotlight/blob/64aeab656e4667325c1bccb598c5cc6a589cf502/app/assets/stylesheets/spotlight/_exhibit_admin.scss#L89-L97

Is this not working for you?  I'm totally happy to entertain other approaches here.

dunn: Maybe I misunderstood the purpose of this `span`.  Does the 'pending' here mean something other than a pending invite?

jkeck: I believe that is the purpose of this `span`.  This is a bit of confusing code (and I don't really recall why it ended up like this), but the parent `td` gets an `invite-pending` class if the user has an invite pending.

By default the `pending-label` span is hidden by using `display: none`. That css is overriden when the parent `td` has the `invite-pending` class and it is instead set explicitly to display by using `display: inline`.

dunn: Oh, I see the problem.  Our campus Bootstrap theme applies a `display` to the `.label` class that overrides this.  So I can fix this on our end.

PR #2183, opened 2019-07-24
fix redirects after uploading resources

dunn:
Followup to GH-1942.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -370,12 +370,13 @@ Rails/HasManyOrHasOneDependent:
     - 'app/models/spotlight/resource.rb'
     - 'app/models/spotlight/site.rb'
 
-# Offense count: 3
+# Offense count: 4
 # Configuration parameters: Include.
 # Include: app/controllers/**/*.rb
 Rails/LexicallyScopedActionFilter:
   Exclude:
     - 'app/controllers/spotlight/catalog_controller.rb'
+    - 'app/controllers/spotlight/resources/upload_controller.rb'
     - 'app/controllers/spotlight/resources_controller.rb'
 
 # Offense count: 3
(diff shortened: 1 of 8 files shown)


dunn:
```
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
Ignoring license option: BCL -- using GPLv2+CE by default
install-jdk.sh 2019-07-17
Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

cbeer:
I would like @ggeisler to take a look at this too; at least in the beginning of spotlight, sending users from the add items page back to the items dashboard was an intentional decision (and, hence, the "save and add another" button on the page where we expected some batch work). If we're keeping users on the add item page, there might be some additional work needed to indicate status and the like?

dunn:
@ggeisler @cbeer sorry for the delay, and the lack of clarity.

Part of that was because I misunderstood the acceptance criteria for the Surfliner ticket that prompted this work: https://gitlab.com/surfliner/surfliner/issues/223

I've updated this PR to limit the changes to two things:
1. Ensure the `Add and continue adding` button actually redirects back to the single-item upload tab (it wasn't, at least for us).
2. Bring the redirect of the CSV Upload into line with the others (have it redirect to the item admin screen, rather than the main tab of the item upload screen).

ggeisler:
@dunn Okay, great, the context from the surfliner ticket helps me a lot, as does your explanation of the two things you said this PR is now limited to.

Based on my understanding of those, this PR sounds good to me, just speaking from the UX point of view.

PR #2184, opened 2019-08-01
ExhibitExportSerializer: remove confirmation token

dunn:
This avoids the edge case where exporting an exhibit with a contact email then
importing it into the same Spotlight instance causes this sort of error:
```
F, [2019-07-29T12:32:28.849634 #15168] FATAL -- : [157062cd-0728-432e-8acb-252372f37ce0] ActiveRecord::RecordNotUnique (PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_spotlight_contact_emails_on_confirmation_token"
DETAIL:  Key (confirmation_token)=(R6A-skseGTy41-9EHs5T) already exists.
: INSERT INTO "spotlight_contact_emails" ("exhibit_id", "email", "confirmation_token", "confirmed_at", "confirmation_sent_at", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"):
```


dunn:
Looks like the same Travis error as in #2183

cbeer:
Rebased and fixed in #2189.

PR #2191, opened 2019-08-06
Explicitly require webdrivers

cbeer:


mejackreed:
Do we need to do this thing? https://github.com/sul-dlss/exhibits/commit/60fb800fd65853da97cabff2bdc956a6abef0436

PR #2201, opened 2019-08-15
Refactor SolrDocumentSidecar to use the custom field slug instead of …

cbeer:
…the solr field name.

Unblocks #2200

### DIFF
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -197,7 +197,8 @@ module Spotlight
     end
 
     def custom_field_params
-      current_exhibit.custom_fields.writeable.pluck(:field)
+      current_exhibit.custom_fields.writeable.pluck(:slug) +
+        current_exhibit.custom_fields.writeable.pluck(:field) # for backwards compatibility
     end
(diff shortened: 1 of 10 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -197,7 +197,8 @@ def uploaded_resource_params
     end
 
     def custom_field_params
-      current_exhibit.custom_fields.writeable.pluck(:field)
+      current_exhibit.custom_fields.writeable.pluck(:slug) +
```
cbeer: Using the old name in strong params may be unnecessary, but done out of an abundance of caution. 🤷‍♂

app/models/spotlight/custom_field.rb
```diff
@@ -92,7 +93,7 @@ def index_fields_config
     end
 
     def should_generate_new_friendly_id?
-      super || persisted?
+      new_record? && slug.blank?
```
cbeer: We used to update the slug all the time, and the solr field name only if something critical changed... now that we're using the slug for data storage, this should make it more like the field name.

PR #2203, opened 2019-08-15
allow IIIF url helpers to be reconfigured

dunn:
This change allows Spotlight applications to use an external (non-riiif) IIIF server for image derivative requests.

It does not teach Spotlight to send uploaded resources to that server; currently using this configuration option requires that the images already exist on the IIIF server.

### DIFF
--- a/app/controllers/spotlight/featured_images_controller.rb
+++ b/app/controllers/spotlight/featured_images_controller.rb
@@ -16,7 +16,10 @@ module Spotlight
     private
 
     def tilesource
-      riiif.info_url(@featured_image.id)
+      Spotlight::Engine.config.iiif_url_helpers.info_url(
+        @featured_image.id,
+        host: request.host_with_port
+      )
     end
 
     # The create action can be called from a number of different forms, so
--- a/app/models/spotlight/featured_image.rb
+++ b/app/models/spotlight/featured_image.rb
@@ -53,8 +53,7 @@ module Spotlight
     def set_tilesource_from_uploaded_resource
       return if iiif_tilesource
 
-      riiif = Riiif::Engine.routes.url_helpers
-      self.iiif_tilesource = riiif.info_path(id)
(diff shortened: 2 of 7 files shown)


dunn:
As an example, this is what I added to my `spotlight_initializer.rb`:
```ruby
module CantaloupeHelpers
  IIIF_BASE_URL = ENV["IIIF_BASE_URL"] || "http://localhost:8182/iiif/2"

  def self.info_url(id, _options)
    "#{IIIF_BASE_URL}/#{id}/info.json"
  end

  def self.info_path(id, _options)
    "#{IIIF_BASE_URL}/#{id}/info.json"
  end

  def self.image_path(id,
                      region: "full",
                      size:,
                      rotation: "0",
                      quality: "default",
                      format: "jpg")
    "#{IIIF_BASE_URL}/#{id}/#{region}/#{size}/#{rotation}/#{quality}.#{format}"
  end
end

Spotlight::Engine.config.iiif_url_helpers = CantaloupeHelpers
```

cbeer:
This PR doesn't seem to be working for me out-of-the-box. When I upload an image, I'm getting an error trying to view it. `master` seems to work fine, though. Do you see the same?

One thing that makes me uncomfortable about this PR, and may be related to the problem above, is by using `Riiif::Engine.routes.url_helpers` directly, I'm a little worried we end up losing some of the Rails routing flexibility that we get when we use `controller.riiif` (mainly, the ability to mount the RIIIF application at an arbitrary path). I'm not sure it's a show-stopper (because we were clearly inconsistent in going through the controller's already-defined routes) but it does give me pause.

dunn:
@cbeer I'm not seeing that error; image uploads are working for me, both in a stock instance and in our Surfliner instance (which uses the manifests instead of fetching the images directly).

I'm not sure I understand your concern about the routes; doesn't `Riiif::Engine.routes.url_helpers` just pick up on where the engine has been mounted by the downstream application?

PR #2204, opened 2019-08-19
ExhibitExportSerializer: ignore theme when unavailable

dunn:
When the target instance does not have the same theme, errors occur: https://gitlab.com/surfliner/surfliner/issues/145#note_203830948

### DIFF
--- a/app/serializers/spotlight/exhibit_export_serializer.rb
+++ b/app/serializers/spotlight/exhibit_export_serializer.rb
@@ -34,6 +34,10 @@ module Spotlight
       property prop
     end
 
+    property :theme, setter: lambda { |fragment:, represented:, **|
+      represented.theme = fragment if Spotlight::Engine.config.exhibit_themes.include? fragment
+    }
+
     collection :searches, populator: ->(fragment, options) { options[:represented].searches.find_or_initialize_by(slug: fragment['slug']) },
                           class: Spotlight::Search do
       (Spotlight::Search.attribute_names - %w(id scope exhibit_id masthead_id thumbnail_id)).each do |prop|


cbeer:
Hm. I'd rather we serialize it on export, but be more permissive when importing.

dunn:
@cbeer I tried to figure out how to do that, but AFAICT the serializer never has access to the exhibit objects themselves, so I'm not sure if it's possible to conditionally turn off that field.

dunn:
Oh, looks like there is a way: http://trailblazer.to/gems/representable/3.0/function-api.html#overview

I'll update this PR.

dunn:
Fully awake now, figured out how to include the theme conditionally.

app/serializers/spotlight/exhibit_export_serializer.rb
```diff
@@ -34,6 +34,10 @@ class ExhibitExportSerializer < Roar::Decorator
       property prop
     end
 
+    property :theme, if: lambda { |options|
+      Spotlight::Engine.config.exhibit_themes.include? options[:doc]['theme']
```
cbeer:
```suggestion
      Spotlight::Engine.config.exhibit_themes.include? options[:input]
```

This is just a guess from skimming through the doc

app/serializers/spotlight/exhibit_export_serializer.rb
```diff
@@ -34,6 +34,10 @@ class ExhibitExportSerializer < Roar::Decorator
       property prop
     end
 
+    property :theme, if: lambda { |options|
```
cbeer: I think `if` applies for both serializing + deserializing. It looks like `setter` may be just for serializing? Something like:

`property :theme, setter: ->(fragment:, represented:, **) { represented.theme = fragment if Spotlight::Engine.config.exhibit_themes.include? fragment }`

dunn: We need it for deserializing though, right?  The serialized export will include the theme since the source instance had it, so the target instance needs to be able to ignore it.

cbeer: Indeed -- I was thinking there's no need to check during serialization.

dunn: Oh, so `setter` is for deserialization, and `getter` for serialization? Am I reading http://trailblazer.to/gems/representable/3.0/function-api.html#getter right?

cbeer: Yes, I believe that's correct.

dunn: righto, updated

PR #2209, opened 2019-09-10
Use the IIIF rules for handling multilingual metadata in a manifest

cbeer:
Fixes https://github.com/sul-dlss/exhibits/issues/1429

### DIFF
--- a/app/models/spotlight/resources/iiif_manifest.rb
+++ b/app/models/spotlight/resources/iiif_manifest.rb
@@ -70,7 +70,7 @@ module Spotlight
         return unless title_fields.present? && manifest.try(:label)
 
         Array.wrap(title_fields).each do |field|
-          solr_hash[field] = json_ld_value(manifest.label)
+          solr_hash[field] = metadata_class.new(manifest).label
         end
       end
 
@@ -94,13 +94,6 @@ module Spotlight
 
           hash[field.field] = value
         end
-      end
-
-      def json_ld_value(value)
-        return value['@value'] if value.is_a?(Hash)
-        return value.find { |v| v['@language'] == default_json_ld_language }.try(:[], '@value') if value.is_a?(Array)
-
-        value
       end
 
       def create_sidecars_for(*keys)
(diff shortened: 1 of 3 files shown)


app/models/spotlight/resources/iiif_manifest.rb
```diff
@@ -221,13 +218,81 @@ def manifest_level_metadata
...
 
         def manifest_fields
           %w(attribution description license)
         end
+
+        # rubocop:disable Metrics/CyclomaticComplexity, Metrics/AbcSize, Metrics/MethodLength, Metrics/PerceivedComplexity
+        def json_ld_value(value)
+          case value
+          # In the case where multiple values are supplied, clients must use the following algorithm to determine which values to display to the user.
+          when Array
```
mejackreed: Thanks for the helpful comments in this method

camillevilla:
Some context on this PR:
- Previously, trying to import a IIIF manifest with multilingual metadata would cause indexing errors and break reindexing for an exhibit. 
  - Example: https://api.digitale-sammlungen.de/iiif/presentation/v2/bsb00017921/manifest
- Spotlight does not support displaying multilingual metadata values yet
  - This PR uses IIIF Presentation API rules to select a single language for displaying metadata values (see below)
  - Metadata values will not change if you toggle locales. (Metadata **labels** will change if translations are provided via the curator UI). 
- The rules for handling multilingual metadata changes in Presentation v3. These rules are captured in the comments of [IiifManifest#json_ld_value](https://github.com/projectblacklight/spotlight/pull/2209/files#diff-5f3208df72ea2ea53e3b55826a01e6e4R230)
  - [IIIF Presentation v2.1 - JSON-LD Considerations: Language of Property Values](https://iiif.io/api/presentation/2.1/#json-ld-considerations)
  -[IIIF Presentation v3.0.0-Beta - JSON-LD Considerations: Language of Property Values](https://iiif.io/api/presentation/3.0/#44-language-of-property-values)

Q: Why did we contribute this upstream? 

A: Stanford is investigating using Mirador 3 instead of OSD for IIIF import items (sul-dlss/exhibits#1429). While this contribution does not implement full support for multilingual description, it fixes a bug that was preventing curators from reindexing exhibits.

PR #2225, opened 2019-11-01
Don't use the blacklight_7 branch of blacklight-oembed…

jkeck:
…(it has been merged, deleted, and a new release cut)


cbeer:
I have a similar commit in #2223 to fix the build there. We'll see who can get to 💚 first.

PR #2231, opened 2019-11-04
Blacklight 7 changed #search_facet_url to #search_facet_path

cbeer:


spec/controllers/application_controller_spec.rb
```diff
@@ -21,7 +21,7 @@
       describe '#search_facets_url' do
```
mejackreed: Should we update this string?

cbeer: 👍 ; it's nice that the string was wrong in the first place.

mejackreed: Yup!

PR #2247, opened 2019-11-08
Enable HTML lang attribute to use I18n.locale value

mejackreed:
Backport to v2.x release

Slightly different than #2246 as we are assuming the Blacklight helper is not available. 🤷‍♂

### DIFF
--- a/app/views/layouts/spotlight/spotlight.html.erb
+++ b/app/views/layouts/spotlight/spotlight.html.erb
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" class="no-js">
+<%= content_tag :html, class: 'no-js', **html_tag_attributes do %>
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


cbeer:
Should we go ahead and backport the blacklight helper upstream?

mejackreed:
likely the responsible thing :) https://github.com/projectblacklight/blacklight/pull/2206

app/views/layouts/spotlight/spotlight.html.erb
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" class="no-js">
+<%= content_tag :html, class: 'no-js', **html_tag_attributes do %>
```
cbeer: Should we preserve compatibility with pre-Blacklight 6.21 deployments?

```suggestion
<%= content_tag :html, class: 'no-js', **(try(:html_tag_attributes) || {}) do %>
```

PR #2271, opened 2019-11-13
Wrap the browse category search box in a row; part of #2262

### LINKED ISSUE #2262: Browse detail page: update description and search box input/label alignment
A couple of issues with optional elements of a browse category detail page:

Screen Shot 2019-11-12 at 5 00 06 PM

- [x] A: Previously we avoided unnecessarily long text lines by applying `column-width: 20em;` and `column-gap: 3em;` to the long description text. We don't seem to be doing that now.

- [x] B: If you turn on "Display search box" for a browse category the search input and its associated label are not well aligned. Previously we showed the label inline with the input.

- [x] ...and the label was in a heavier font-weight (unless that was done in Exhibits).

Example showing how we used to do this things:

Screen Shot 2019-11-12 at 5 01 03 PM

cbeer:
Screen Shot 2019-11-15 at 08 28 43

Fixes #2262

### DIFF
--- a/app/assets/stylesheets/spotlight/_browse.scss
+++ b/app/assets/stylesheets/spotlight/_browse.scss
@@ -106,8 +106,11 @@ $image-overlay-bottom-margin: 36px;
 
 .search-box-container {
   margin: 0 auto;
-  text-align: center;
   width: 100%;
+
+  .col-form-label {
+    text-align: right;
+  }
 
   .browse-search-form {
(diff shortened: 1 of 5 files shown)


ggeisler:
@camillevilla is correct about the border. Here's an example of the goal (more or less):

Screen Shot 2019-11-14 at 5 46 00 PM

PR #2284, opened 2019-11-14
Use the site title in the email, if it is provided

### LINKED ISSUE #2282: Update subject line of feedback submission form
This is pretty minor since presumably Spotlight adopters will customize the Feedback email details for their institution, but in case they don't, it seems like it's worth updating the current subject line text from "Blacklight Contact Form" to "Spotlight Contact Form". 

I'd actually suggest "Spotlight Feedback" might be more appropriate, given the label of the form trigger is "Feedback" and not "Contact".

Screen Shot 2019-11-13 at 3 43 41 PM

cbeer:
Fixes #2282

### DIFF
--- a/app/models/spotlight/contact_form.rb
+++ b/app/models/spotlight/contact_form.rb
@@ -20,16 +20,24 @@ module Spotlight
     def headers
       {
         to: to,
-        subject: "#{I18n.t(:'blacklight.application_name')} Contact Form",
+        subject: I18n.t(:'spotlight.contact_form.subject', application_name: application_name),
         from: %("#{name}" <#{email}>),
-        cc: current_exhibit.contact_emails.join(', ')
+        cc: contact_emails.join(', ')
       }
     end
 
     private
 
+    def application_name
+      current_exhibit&.title || Spotlight::Site.instance.title || I18n.t(:'blacklight.application_name')
+    end
+
     def to
-      Spotlight::Engine.config.default_contact_email || current_exhibit.contact_emails.first.to_s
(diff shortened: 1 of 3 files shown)


ggeisler:
Okay, sounds good. But can we still make the switch to "Feedback" rather than "Contact" that I suggested in #2282? In fact, given your idea to use the exhibit title if available rather than the application name, I'd suggest we go with this:

```
subject: "%{application_name} exhibit feedback"
```
(lowercase "exhibit feedback" so it is more distinct from the app name/exhibit title)

So example rendered subject lines would be:

- Spotlight exhibit feedback
- Stanford University Libraries exhibit feedback
- Stanford Geological Survey Collection exhibit feedback
- The Bob Fitch Photography Archive exhibit feedback

etc. I know it's not a huge difference but I think the change would make for a slightly more informative subject line than "... Contact Form" and is more consistent with the button label used to access the form.

cbeer:
👍

app/models/spotlight/contact_form.rb
```diff
@@ -20,16 +20,24 @@ class ContactForm
...
+      current_exhibit&.title || Spotlight::Site.instance.title || I18n.t(:'blacklight.application_name')
+    end
+
     def to
-      Spotlight::Engine.config.default_contact_email || current_exhibit.contact_emails.first.to_s
+      Spotlight::Engine.config.default_contact_email || contact_emails.first.to_s
+    end
+
+    def contact_emails
+      current_exhibit&.contact_emails || []
```
camillevilla: 😎 Safety first!

PR #2306, opened 2019-11-18
Remove the nav-link class from the Create Exhibit button.

jkeck:
_I'm pretty sure this was a larger button previously, feel free to 🔪 this if we decide it's not a good idea._

## Before
before

## After
after


ggeisler:
Good call @jkeck. I noticed that at some point but never got back to write it up.

PR #2328, opened 2019-12-09
[BACKPORT] Add Spotlight::ValidityChecker for checking if a job…

jkeck:
…(say, indexing) still needs to be run to avoid doing duplicative work


cbeer:
I think so; we saw this on some other backports due to some kind of caching issue.

PR #2335, opened 2019-12-10
Fix bug in SearchRresults widget

### LINKED ISSUE #2332: Search Results widget produces exception
If I try to save the Search Results widget on a feature page:

Screen Shot 2019-12-09 at 3 59 58 PM

I get this:

Screen Shot 2019-12-09 at 3 59 30 PM

jkeck:
Blacklight has moved the `search_results` method from where it was. 

Fixes #2332

### DIFF
--- a/app/controllers/spotlight/pages_controller.rb
+++ b/app/controllers/spotlight/pages_controller.rb
@@ -118,6 +118,7 @@ module Spotlight
 
     def undo_link
       return unless can? :manage, @page
+      return if @page.versions.blank?
 
       view_context.link_to(t(:'spotlight.versions.undo'), revert_version_path(@page.versions.last), method: :post)
     end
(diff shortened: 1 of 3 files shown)


app/controllers/spotlight/pages_controller.rb
```diff
@@ -116,8 +116,17 @@ def _prefixes
       @_prefixes ||= super + ['catalog']
     end
 
+    # Add a Page specific search_results method that takes user params as
+    # an option and passes that off to the search service to get results
+    # @param [Hash] the query parameters
+    # @return [Object] the search results object from the configured search service
+    def search_results(user_params)
```
jkeck: I could easily be convinced to use a method name that doesn't match the API of the search_service (like `block_search_results` or something like that), but kept it the same for now.

PR #2341, opened 2019-12-12
Removes the ability to spoof from emails.

mejackreed:
… from address

Related to https://github.com/sul-dlss/dlme/issues/626

Fixes https://github.com/sul-dlss/dlme/issues/813

`Net::SMTPFatalError: 554 Message rejected: Email address is not verified. The following identities failed the check in region US-WEST-2:`

If you use something like Amazon's Simple Email Service (SES) you can't just spoof the `from` header, but must use a verified address. This pull request removes spotlight from spoofing froms.

Adopters should use the ActionMailer `default_options[:from]` https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration

### DIFF
--- a/app/mailers/spotlight/contact_mailer.rb
+++ b/app/mailers/spotlight/contact_mailer.rb
@@ -2,7 +2,7 @@
 
 module Spotlight
   # Mailer for reporting problems to the application contact and/or exhibit administrator
-  class ContactMailer < ActionMailer::Base
+  class ContactMailer < ApplicationMailer
     def report_problem(contact_form)
       @contact_form = contact_form
       mail(@contact_form.headers)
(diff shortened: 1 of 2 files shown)


app/mailers/spotlight/contact_mailer.rb
```diff
@@ -2,7 +2,7 @@
 
 module Spotlight
   # Mailer for reporting problems to the application contact and/or exhibit administrator
-  class ContactMailer < ActionMailer::Base
+  class ContactMailer < ApplicationMailer
```
mejackreed: Is this controversial? I'm not sure? I'd offer this is the most straightforward way I could think of for a mailer to inherit application defaults without needing to go and create custom configurations for it.

cbeer: 👍 from me; I think this is generated by default these days.

PR #2343, opened 2019-12-12
Add locale to mailer invite

mejackreed:
Fixes an issue where invitations sent in multi-lingual exhibits raise an
error.

Fixes https://github.com/sul-dlss/dlme/issues/810

### DIFF
--- a/app/views/devise/mailer/invitation_instructions.html.erb
+++ b/app/views/devise/mailer/invitation_instructions.html.erb
@@ -7,7 +7,7 @@
 %>
 <p><%= t("spotlight.#{key}.invitation_instructions.hello", email: @resource.email) %></p>
 
-<p><%= t("spotlight.#{key}.invitation_instructions.someone_invited_you", role: @resource.roles.first.role, exhibit_name: @resource.roles.first.resource.title, url: spotlight.exhibit_home_page_url(@resource.roles.first.resource, locale: I18n.locale)) %></p>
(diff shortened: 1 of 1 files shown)


app/views/devise/mailer/invitation_instructions.html.erb
```diff
@@ -7,7 +7,7 @@
 %>
 <p><%= t("spotlight.#{key}.invitation_instructions.hello", email: @resource.email) %></p>
 
-<p><%= t("spotlight.#{key}.invitation_instructions.someone_invited_you", role: @resource.roles.first.role, exhibit_name: @resource.roles.first.resource.title, url: spotlight.exhibit_home_page_url(@resource.roles.first.resource)) %></p>
+<p><%= t("spotlight.#{key}.invitation_instructions.someone_invited_you", role: @resource.roles.first.role, exhibit_name: @resource.roles.first.resource.title, url: spotlight.exhibit_home_page_url(@resource.roles.first.resource, locale: I18n.locale)) %></p>
```
cbeer:
```suggestion
<p><%= t("spotlight.#{key}.invitation_instructions.someone_invited_you", role: @resource.roles.first.role, exhibit_name: @resource.roles.first.resource.title, url: spotlight.exhibit_home_page_url(@resource.roles.first.resource, locale: nil)) %></p>
```

I'm not sure if that'll work, but at least it won't inflict the inviting users locale on the invitee?

Actually, I wonder if setting the locale (to nil?) in the action mailer config would work too? 🤷‍♂

mejackreed: My thinking was that we wanted to inflict the inviting users locale to the invitee. 🤷‍♂ 🤷‍♂ 🤷‍♂ 

No idea.

PR #2349, opened 2020-01-15
Fixes issue where search-btn in input group can be off

mejackreed:
See https://github.com/sul-dlss/exhibits/issues/1478

## Before
Screen Shot 2020-01-15 at 11 50 03 AM

## After
Screen Shot 2020-01-15 at 11 50 43 AM

### DIFF
--- a/app/assets/stylesheets/spotlight/_blacklight_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_blacklight_overrides.scss
@@ -8,6 +8,10 @@
   .input-group {
     width: 100%;
   }
+  
+  .search-btn {
+    max-height: 2.375rem;
+  }
 
   .search_field {
     background: transparent;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/spotlight/_blacklight_overrides.scss
```diff
@@ -8,13 +8,9 @@
...
-
-  .input-group-addon {
-    &.for-search-field {
-      background-color: $input-bg;
-      border-radius: $border-radius 0 0 $border-radius;
-      width: 15ch;
-    }
+  
+  .search-btn {
+    max-height: 2.375rem;
```
cbeer: Just curious, where does 2.375 come from?

mejackreed: Its the calculated rem value for 38px when using a base of 16px. We could use 38px here instead but if someone changes the default font size i think it would be wrong.

PR #2350, opened 2020-01-15
Retain capitialization for header navbar

mejackreed:
Removed in https://github.com/projectblacklight/spotlight/commit/1a4f3654758c9d02b978b48b3ca802dfada05264

Screen Shot 2020-01-15 at 12 02 06 PM

Addressing https://github.com/sul-dlss/exhibits/issues/1493


mejackreed:
I'm not sure if we removed this to better support Arabic or not.

mejackreed:
Doesn't seem to be.. since we are manually doing this in DLME

PR #2367, opened 2020-01-21
Quick spike for searching across exhibits

cbeer:


cbeer:
Abandoned in favor of https://github.com/sul-dlss/exhibits/pull/1615. There's just too many local assumptions that are hard to accommodate.

PR #2368, opened 2020-01-22
Update solr documents embed column layout logic

mejackreed:
Fixes https://github.com/sul-dlss/exhibits/issues/1208

Only uses two column layout when there is no text. This aims to improve the layout for feature pages.

Screen Shot 2020-01-21 at 4 58 10 PM


ggeisler:
Based on the screenshot, it seems like this matches the goal in my comment in https://github.com/sul-dlss/exhibits/issues/1208#issuecomment-401400129.

If I remember correctly, the problem to be solved here was specific to the Embed widget (as I assume the solution is) but if you know differently @jkeck let us know.

PR #2371, opened 2020-01-22
Rails 6 compatibility

dunn:
Building on top of https://github.com/projectblacklight/spotlight/pull/2221


dunn:
I don't know what to make of these errors, since I don't see any `+` operators in the relevant files: https://travis-ci.org/projectblacklight/spotlight/jobs/640666697#L3570

dunn:
Here's where the error actually occurs: https://github.com/bootstrap-ruby/bootstrap_form/blob/master/lib/bootstrap_form/form_group.rb#L37

dunn:
https://github.com/bootstrap-ruby/bootstrap_form/pull/555/files

dunn:
The page editing tools don't seem to be loading correctly:
Screen Shot 2020-01-23 at 10 10 33 AM

There have been issues with Sprockets while upgrading to Rails 6, so some of the JS assets might not be included correctly.

dunn:
We might need to make changes to support Webpacker, as Blacklight did: https://github.com/projectblacklight/blacklight/pull/2163/files

dunn:
I think the problem is that the generated Rails 6 app is insisting on using webpack even though it isn't configured.  So maybe we should just get it set up while we're at it.

dunn:
afaict the problem now is that the assets in `vendor` aren't being loaded on rails 6, but i can't figure out why

jkeck:
@dunn, we're looking at releasing 3.0.0 soon hopefully, and I think it would be nice to get this into that release if possible.  Are you all in a position to validate that this works w/ rails 6 in production in the near future?

We won't be able to update our production instance to rails 6 until our next work cycle most likely, and I'm not 100% sure when that will be.

dunn:
In my local testing this is now blocked by the need for a new release of iiif-image-api with the recent commit that unpins its Rails versions.  I'm still having trouble getting added to the list of people who can cut releases so if someone else can do that it would help out.

jkeck:
Huh, I guess I have perms.  Once samvera-labs/iiif-image-api#5 is merged I can cut a release.

dunn:
https://github.com/curationexperts/riiif/pull/133

dunn:
this is new and surprising (only on rails 6):
```
  81) Spotlight::ExhibitImportExportService should round-trip data with a thumbnail is copied
      Failure/Error:
        exhibit.as_json(
          {
            except: %i[id slug site_id],
            include: {}.merge(
              if_include?(:config,
                          main_navigations: {
                            except: %i[id exhibit_id]
                          },
                          contact_emails: {
                            except: %i[id exhibit_id confirmation_token]

      SystemStackError:
        stack level too deep
      # ./app/services/spotlight/exhibit_import_export_service.rb:303:in `raw_json'
      # ./app/services/spotlight/exhibit_import_export_service.rb:176:in `call'
      # ./app/services/spotlight/exhibit_import_export_service.rb:176:in `block in as_json'
      # ./app/services/spotlight/exhibit_import_export_service.rb:175:in `each'
      # ./app/services/spotlight/exhibit_import_export_service.rb:175:in `inject'
      # ./app/services/spotlight/exhibit_import_export_service.rb:175:in `as_json'
      # ./spec/services/spotlight/exhibit_import_export_service_spec.rb:92:in `block (3 levels) in <top (required)>'
      # ./spec/services/spotlight/exhibit_import_export_service_spec.rb:86:in `block (3 levels) in <top (required)>'
      # ./spec/services/spotlight/exhibit_import_export_service_spec.rb:349:in `block (4 levels) in <top (required)>'
```

dunn:
even stranger, running that spec file in isolation gives a different error:
```
  27) Spotlight::ExhibitImportExportService should round-trip data with a thumbnail is copied
      Failure/Error: (super.constantize if defined?(super)) || default_document_type

      NameError:
        wrong constant name {:registered_extensions=>[{:module_obj=>{:_dependencies=>[], :_included_block=>{}}, :condition_proc=>{}}, {:module_obj=>{}, :condition_proc=>nil}, {:module_obj=>{}, :condition_proc=>nil}, {:module_obj=>{}, :condition_proc=>nil}], :unique_key=>"id"}
      # ./app/models/spotlight/solr_document_sidecar.rb:40:in `document_type'
      # ./app/models/spotlight/solr_document_sidecar.rb:36:in `document'
      # ./app/services/spotlight/exhibit_import_export_service.rb:124:in `block in from_hash!'
      # ./app/services/spotlight/exhibit_import_export_service.rb:122:in `each'
      # ./app/services/spotlight/exhibit_import_export_service.rb:122:in `from_hash!'
      # ./app/models/spotlight/exhibit.rb:103:in `import'
      # ./spec/services/spotlight/exhibit_import_export_service_spec.rb:86:in `block (3 levels) in <top (required)>'
      # ./spec/services/spotlight/exhibit_import_export_service_spec.rb:349:in `block (4 levels) in <top (required)>'
```

dunn:
The failures on Rails 6 are caused by https://github.com/projectblacklight/spotlight/pull/2507 cc @cbeer

jkeck:
Just as another data point on that `SystemStackError: stack level too deep` error, it doesn't seem to occur when running the test in isolation (or even along w/ all the other tests in the services directory).  It only seems to happen when running along w/ other tests (which is somewhat concerning).

cbeer:
I had a not-dissimilar error working on #2507 and fixed it with:

https://github.com/projectblacklight/spotlight/pull/2507/files#diff-3d8f083f8b6b996d9f2aeef14bb8d985R292

Maybe there's something similar happening in Rails 6?

lib/generators/spotlight/install_generator.rb
```diff
@@ -10,6 +10,25 @@ class Install < Rails::Generators::Base
     class_option :solr_update_class, type: :string, default: 'Spotlight::SolrDocument::AtomicUpdates'
     class_option :mailer_default_url_host, type: :string, default: '' # e.g. localhost:3000
 
+    # we're not (yet) using webpacker, so we need to re-add sprockets functionality
+    def add_js
+      return unless ENV['RAILS_VERSION'].to_i == 6
```
jkeck, line 15: We may want to use `Rails.version` here instead of `ENV['RAILS_VERSION']`.

At the point of executing a generator we should have a rails context and know the version that is installed. I believe typical users won't be passing the `RAILS_VERSION` env var when installing spotlight and will use whatever the most recent version of rails they have is.

jkeck:
@dunn hopefully #2515 will resolve https://github.com/projectblacklight/spotlight/pull/2371#issuecomment-609970917 and https://github.com/projectblacklight/spotlight/pull/2371#issuecomment-609975020

dunn:
https://github.com/projectblacklight/spotlight/pull/2371/commits/df47557b11ab2bd9823ea85b5780f6bb26e0f8af ensures that the standard widgets come first

lib/generators/spotlight/install_generator.rb
dunn: fixed with https://github.com/projectblacklight/spotlight/pull/2371/commits/379001852c36b95188ed44b5a2ecf5c16ce97aec

dunn:
`./spec/services/spotlight/exhibit_import_export_service_spec.rb:403` is now failing, because of discrepancies in the `updated_at` times of the exported objects.  Is that time value more precise than it used to be?
```diff
        :title => "Exhibit Title 1017",
        :translations => [],
       -:updated_at => "2020-04-29T21:29:00.099Z",
       +:updated_at => "2020-04-29T21:29:00.210Z",
        :weight => 50,
```

dunn:
requested changes made @jkeck

dunn:
@jkeck @cbeer if I bring this back up to date, can I expect it to be reviewed?

jkeck:
@dunn sorry, was away on vacation for a bit.  I believe the widget order was the only thing that jumped out at me.  I can find some time to do another review/smoke test in the next week or so.

dunn:
In order to achieve backwards compatibility, we'd have to be able to define `SirTrevorRails::Blocks.custom_block_types` without using a generated initializer, right?  Is that possible?

Otherwise we'll have to add `spotlight:install` to the upgrade notes (assuming re-running it doesn't clobber the user's modified files?).

dunn:
I don't really know anything about FriendlyID, but is the gist that specifying `:finders` at the model level like this isn't sufficient anymore? https://github.com/projectblacklight/spotlight/blob/35274d784a8d512cbf2ab11210a6d57106573009/app/models/spotlight/about_page.rb#L8

jkeck:
@dunn WRT the friendly ID stuff, I was meaning that I found myself needing to enable finders explicitly in our app (the thing that should enable ActiveRecord objects to be able to find things by the slug automagically when using `.find` instead of having to use `.finders.find` everywhere.  If nothing is jumping out at you, then no worries.

I have an idea about how to handle the configs in a possibly backwards compatible way and will push up a PR w/ a possible route.  I will `@` you on it in case you have any thoughts.

dunn:
sounds good, thanks very much!

PR #2372, opened 2020-01-23
Allow downstream applications to set exhibit specific themes

jkeck:
This is a stab at an approach that would allow us to resolve sul-dlss/exhibits#1462 downstream without changing behavior in Spotlight proper.

If this approach is agreeable, I'll update https://github.com/projectblacklight/spotlight/wiki/Creating-a-theme to include details of how to configure this downstream.

### DIFF
--- a/app/models/spotlight/exhibit.rb
+++ b/app/models/spotlight/exhibit.rb
@@ -5,6 +5,7 @@ module Spotlight
   ##
   # Spotlight exhibit
   class Exhibit < ActiveRecord::Base
+    class_attribute :themes_selector
     include Spotlight::ExhibitAnalytics
     include Spotlight::ExhibitDefaults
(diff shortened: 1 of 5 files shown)


app/models/spotlight/exhibit.rb
```diff
@@ -5,6 +5,7 @@ module Spotlight
   ##
   # Spotlight exhibit
   class Exhibit < ActiveRecord::Base
+    class_attribute :themes_selector
```
jkeck: Would happily entertain other ideas for naming this.

jkeck: This could also theoretically go in the engine config, went back and forth and landed here, but not for any reason other than the intention is to have the ability to be exhibit specific.

PR #2373, opened 2020-01-23
update rubocop to 0.79

dunn:


dunn:
🎊 https://github.com/bootstrap-ruby/bootstrap_form/pull/555

camillevilla:
After chatting with @dunn and seeing some more examples of various [RSpec/PredicateMatcher](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/PredicateMatcher) configurations, we decided to stick with the default config (`Strict:true, EnforcedStyle: inflected`).

PR #2374, opened 2020-01-24
Limit search across results to exhibits that the user has access to

jkeck:
Just pushing up some code that I have been poking around with.


cbeer:
Rebased and merged into the branch.

PR #2375, opened 2020-01-24
Add a custom search across form when being used in a sidebar/large screen

jkeck:
large-screen
small-screen

### DIFF
--- a/app/views/spotlight/search_across/_search_form.html.erb
+++ b/app/views/spotlight/search_across/_search_form.html.erb
@@ -1,0 +1,33 @@
+<div class="facets-header">
+  <h2 class="facets-heading"><%= t('spotlight.search_across.results.heading') %></h2>
+</div>
+
+<div class="d-lg-none">
+  <%= render 'catalog/search_form', presenter: presenter %>
+</div>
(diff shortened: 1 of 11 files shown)


app/views/spotlight/search_across/_search_form.html.erb
```diff
@@ -0,0 +1,36 @@
+<% if render_search_across_form? %>
+  <div class="facets-header">
+    <h2 class="facets-heading"><%= t('spotlight.search_across.results.heading') %></h2>
+  </div>
+
+  <div class="d-lg-none">
+    <%= render 'catalog/search_form', presenter: presenter %>
```
jkeck: I really don't like having to do this, but I don't see how to get the responsive input grouping described in the ticket without necessitating changing the structure of the elements at different breakpoints.

Happy to look at alternatives/pair up on hacking away at this.

jkeck: It's also worth mentioning that the default form doesn't do exactly what is described in the design (but is ohhhh so close)

cbeer:
Merged into  the branch.

PR #2377, opened 2020-01-27
Remove top-border for translation page tables

mejackreed:
Fixes https://github.com/sul-dlss/exhibits/issues/1554

Bootstrap 4 adds for a `.table.table-striped` a top border (BS3 did not have a top border). This reverts for the translation pages the previous table behavior (no top border). 

@ggeisler was there a better way this should be handled that is more globally relevant?

## Before
Screen Shot 2020-01-27 at 2 36 17 PM

## After
Screen Shot 2020-01-27 at 2 35 52 PM

### DIFF
--- a/app/assets/stylesheets/spotlight/_translations.scss
+++ b/app/assets/stylesheets/spotlight/_translations.scss
@@ -14,6 +14,10 @@
   .browse-category-description {
     margin-bottom: $padding-large-vertical * 2;
   }
+
+  .table.table-striped th {
+    border-top: 0;
+  }
 }
 
 .translation-subheading {


ggeisler:
@mejackreed 

> @ggeisler was there a better way this should be handled that is more globally relevant?

I noted in the ticket that I thought the team already fixed this in other places, so if that's true I would look to that solution (https://github.com/projectblacklight/spotlight/pull/2363 to fix https://github.com/sul-dlss/exhibits/issues/1530).

mejackreed:
That solution is a bit different as that is different and relies not on the table structure layout. I think this might be the way to go then.

ggeisler:
Okay, sounds good (they looked like similar problems on the surface but I didn't look past that).

PR #2382, opened 2020-01-28
Rework sort field panels to be wider and put the translated sort …

jkeck:
…fields into a card body (instead of next to the title in parens).

Closes sul-dlss/exhibits#1550

## Before
Screen Shot 2020-01-28 at 12 34 31 PM

## After
Screen Shot 2020-01-28 at 12 35 07 PM

### DIFF
--- a/app/views/spotlight/search_configurations/_sort.html.erb
+++ b/app/views/spotlight/search_configurations/_sort.html.erb
@@ -3,10 +3,10 @@
   <p class="instructions"><%= t(:'.help') %></p>
 
   <%= f.fields_for :sort_fields, @blacklight_configuration.blacklight_config.sort_fields.keys do |vt| %>
-    <ol class="dd-list col-md-7 disabled-sort-option sort_fields_admin">
+    <ol class="dd-list col-10 disabled-sort-option sort_fields_admin">
       <li>
         <div class="panel">
-          <div class="panel-content">
+          <div class="card">
             <div class="card-header">
               <h3 class="h6 card-title">
                 <%= vt.fields_for default_field.key, default_field do |sort| %>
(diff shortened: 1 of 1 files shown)


jkeck:
@ggeisler the padding that I ended up w/ by default seems to be a bit larger than in the mockup (not sure if that's a Bootstrap 3 vs 4 thing or not).  Is what's in the screenshot above 👌?

ggeisler:
@jkeck. Nice. I think we can go with this default padding for now. 

I don't think it's ideal, just because for any page where we list panels like this that are intended to be reordered via drag and drop I think making them as compact as possible (while still being readable) is a good aim. But I'd like to do this in as global way as possible (i.e., across all Admin panels) so maybe we can wait on refining that detail. It's a nice improvement as is.

Off-hand (I don't have my mockup available any longer) I think `0.5rem` top and bottom spacing would be ideal (it's probably using `1rem` now).

app/views/spotlight/search_configurations/_sort.html.erb
```diff
@@ -36,10 +38,12 @@
                       <a href="#edit-in-place" class="field-label edit-in-place"><%= config.label %></a>
                       <%= sort.hidden_field :label, {data: {:"edit-field-target" => "true"}} %>
                     </h3>
-                    (<%= translate_sort_fields(config) %>)
                     <%= sort.hidden_field :weight, {value: index, data: {property: "weight"}} %>
                   <% end %>
                 </div>
+                <div class="card-body bg-white">
```
jkeck: I explicitly added this `bg-white` class since the `dd3-content` class got moved up (otherwise this background was gray-ish as well).  Happy to 🔪 this, but this kept the original design.

PR #2388, opened 2020-01-31
upgrade sir-trevor to 0.6.1

dunn:
includes a breaking change: https://github.com/madebymany/sir-trevor-rails#upgrade-guide-to-v060

Extracted from #2371

### DIFF
--- a/app/assets/javascripts/spotlight/admin/sir-trevor/block_controls.js
+++ b/app/assets/javascripts/spotlight/admin/sir-trevor/block_controls.js
@@ -10,7 +10,9 @@
     img.setAttribute('role', 'img');
 
     var use = document.createElement('use');
-    use.setAttributeNS('https://www.w3.org/1999/xlink', 'href', SirTrevor.config.defaults.iconUrl + "#" + block.icon_name);
+    use.setAttributeNS('https://www.w3.org/1999/xlink',
+                       'href',
+                       SirTrevor.config.defaults.iconUrl + "#" + block.icon_name);
     img.appendChild(use);
     el.appendChild(img);
     el.appendChild(document.createTextNode(block.title()));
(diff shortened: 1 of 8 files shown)


dunn:
https://github.com/madebymany/sir-trevor-rails/issues/59

dunn:
i don't consider this a breaking change for spotlight users, since they shouldn't have to make any changes to their application when upgrading, but if that's a concern i can leave it as part of #2371

spec/features/javascript/block_controls_spec.rb
```diff
@@ -27,12 +27,14 @@
...
+
       within(first('.st-controls-group')) do
-        expect(page).to have_content 'Standard widgets'
+        expect(page).to have_content 'Exhibit item widgets'
         expect(page).to have_css('.st-block-controls__button')
       end
+
       within(all('.st-controls-group').last) do
-        expect(page).to have_content 'Exhibit item widgets'
+        expect(page).to have_content 'Standard widgets'
```
jkeck: Hmm, this is a bit odd that these sections were reversed (if I'm reading this change correctly).

Any idea what's going on there?  I think this will be an issue for us if these widget sections were swapped.

dunn: should be fixed now; i guess it had something to do with how block sections are defined now

dunn:
it's worth noting that the vendored copy of the sir-trevor js hasn't been updated in a few years: 435eaf8a382a54a60f6fd21b30c61770032c7444 (0.8.2 is the current latest version).

dunn:
i'm not sure what to make of these test failures.  when i put sir-trevor in the gemspec, the widget picker doesn't load; when i put it the gemfile.extras, this happens.

dunn:
I'm closing this in favor of #2371 which includes this version change.

PR #2390, opened 2020-01-31
Align and bold form col label ala BS3

mejackreed:
Fixes https://github.com/sul-dlss/exhibits/issues/1549

Screen Shot 2020-01-31 at 3 32 16 PM
Screen Shot 2020-01-31 at 3 31 57 PM

### DIFF
--- a/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
@@ -100,3 +100,8 @@ label.radio {
     padding-left: 0;
   }
 }
+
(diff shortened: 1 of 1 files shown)


cbeer:
I'm not thrilled with doing this in spotlight; it feels like bootstrap made a global choice that we don't agree with and we'd be better off providing our own theme locally to override it.

mejackreed:
I'd be fine with that approach. :+1:

mejackreed:
I'd be fine with that approach. :+1:

mejackreed:
Going to go ahead and close this and look at doing this locally.

PR #2400, opened 2020-02-06
Address Custom Field form field layout in add/edit forms 

jkeck:
Closes sul-dlss/exhibits#1600

### Add
record-add-before

### Edit
record-edit-before

### Add
record-add-after

### Edit
record-edit-after

### DIFF
--- a/app/views/spotlight/custom_fields/form_group/_text.html.erb
+++ b/app/views/spotlight/custom_fields/form_group/_text.html.erb
@@ -1,14 +1,23 @@
-<div class="form-group">
+<% inline = local_assigns[:inline].presence %>
+<% label_col = f&.label_col %>
+<% control_col = f&.control_col %>
+<% offset_col = label_col.sub('col', 'offset') %>
+
+<div class="form-group <%= "row" if inline %>">
   <% Array(value || '').each_with_index do |v, index| %>
-    <%= f.label field.slug, field.label, namespace: "f#{index}",class: "control-label #{'sr-only' if index > 0}" %>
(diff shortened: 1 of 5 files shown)


app/views/spotlight/custom_fields/form_group/_text.html.erb
```diff
@@ -1,14 +1,23 @@
-<div class="form-group">
+<% inline = local_assigns[:inline].presence %>
+<% label_col = f&.label_col %>
+<% control_col = f&.control_col %>
+<% offset_col = label_col.sub('col', 'offset') %>
```
jkeck: Not super happy with having to do this here.  Happy to entertain additional ideas.

ggeisler:
@jkeck These updates look good, but I have one concern about the item edit form. From your screenshot it looks like a metadata field ("Description"), not part of the form, is displaying to the right of the form fields.

I checked another item on -stage, different from one I used to identify the problem in the issue, and it looks like the whole read-only part of the item page is not displaying in a new row separate from the edit form:

Screen Shot 2020-02-06 at 12 11 34 PM

So while I didn't call this out in the ticket, I wonder if you can update things to force a new row after the edit form, so we don't get any side-by-side display of the form and the read-only part of the item page? Or I could ticket that separately.

ggeisler:
For reference, it looks like we currently do something like I suggested above, on -prod:

Screen Shot 2020-02-06 at 12 18 00 PM

jkeck:
My screenshots were for an uploaded item that I didn't upload any content for (so it didn't show a viewer) which is why it looked so odd).

However; as noticeable in your screenshot, something is way off w/ the edit screen.  I've added a commit (72f9028) that aims to resolve that.

## Before
edit-before

## After
edit-after

PR #2401, opened 2020-02-06
Add some bootstrap classes and custom css for a better responsive …

jkeck:
…exhibit navbar.

Closes sul-dlss/exhibits#1618

## Small Screen Before
sm-before

## Small Screen After
sm-after

## Medium Screen Before/After (unchanged)
md-before-n-fater

### DIFF
--- a/app/assets/stylesheets/spotlight/_exhibit_navbar.scss
+++ b/app/assets/stylesheets/spotlight/_exhibit_navbar.scss
@@ -1,0 +1,10 @@
+.exhibit-navbar {
+  .navbar-nav,
+  .exhibit-search-form {
+    width: 100%;
+
+    @media (min-width: breakpoint-min("md")) {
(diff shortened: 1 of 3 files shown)


ggeisler:
@jkeck Do you think it's feasible to add `1rem` top margin as a media query at the `sm` size and below, to provide some separation between the last menu item and the searchbox? I think for touch especially those targets are too close together to be ideal.

I did mention that in the ticket, just so you don't think I'm always adding new requirements to your PRs:

`I'd add top-margin: 1rem; to the searchbox nav in this context, to provide some additional separation between it and the last menu item (I did that in the screenshot above).`

jkeck:
@ggeisler I've updated to add a top margin of the bootstrap spacer below medium.

updated

ggeisler:
Looks good @jkeck thanks.

PR #2404, opened 2020-02-07
Update bootstrap styling to take into account that the grid layout is…

jkeck:
… also done by flexbox now.

Closes sul-dlss/exhibits#1624

## Before
item-grid-before

## After
item-grid-after

### DIFF
--- a/app/assets/stylesheets/spotlight/_pages.scss
+++ b/app/assets/stylesheets/spotlight/_pages.scss
@@ -110,12 +110,9 @@
   }
 
   .items-col {
+    display: flex;
     margin-bottom: 1em;
     z-index: 1;
-    display: flex;
-    flex-flow: row wrap;
-    justify-content: space-between;
-    align-content: stretch;
 
     @media (max-width: breakpoint-max("md")) {
       margin-bottom: $line-height-lg;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/spotlight/_pages.scss
```diff
@@ -102,6 +102,9 @@
 }
 
 .item-grid-block {
+  margin-left: -15px;
+  margin-right: -15px;
```
jkeck: Not sure if there is a variable that would be better to use here.  The other thing that might be a possibility would be to keep the `row` class but also add an explicit `d-block`, but not sure if that's any more straight forward.

jkeck: Maybe `-($grid-gutter-width / 2)`?

ggeisler:
@jkeck Are the top and bottom rows different heights? You checked that off in the ticket so I'm happy to trust you, just doesn't look like it from your "After" screenshot.

jkeck:
Ah, no, they're not.  I didn't touch that particular part of the css, but I'll look into what's going on there. (I glanced over that section and mistook the height to mean something else).

jkeck:
Can't slip anything passed you @ggeisler 

## Updated
Updated

PR #2414, opened 2020-02-10
Update file upload spacing

### LINKED ISSUE #2411: Image upload radio/input buttons need vertical spacing
Before the BS4 upgrade, the radio buttons and associated file upload buttons in the admin pages used for uploading an image (masthead, thumbnail, avatar) each got top margin via this SCSS selector:

```scss
  div.radio + input[type="file"] {
    margin-top: $padding-large-vertical;
  }
```

That selector still exists but the radio/input buttons are now within a `<div class="form-check">` so they don't match the selector looking for `div.radio`. As a result, we don't get the top margin we need to separate the radio/input buttons from the elements above them:

Screen Shot 2020-02-07 at 5 48 45 PM

I don't want to try to fix this myself because I'm not sure if there was a good reason the class name is different.

Radio/input buttons used for purposes other than uploading an image are not affected by this issue.

mejackreed:
Fixes #2411 

## Before
Screen Shot 2020-02-10 at 11 39 21 AM

## After
Screen Shot 2020-02-10 at 3 16 31 PM

### DIFF
--- a/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
@@ -71,10 +71,6 @@ h1,h2,h3,h4,h5,h6 {
   margin-top: 0;
 }
 
-label.radio {
-  font-weight: normal;
-}
-
 #main-container .nav {
   margin-bottom: $line-height-lg * 1rem;
 }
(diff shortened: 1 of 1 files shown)


ggeisler:
@mejackreed From your screenshot it looks like this only addresses the input button and not the radio button? The goal is for some vertical spacing above both of those elements, so there is also space between the help text and the radio button.

See an equivalent page in a -prod exhibit for an example of what I mean.

mejackreed:
Updated to be more consistent based off of @ggeisler concerns.

PR #2415, opened 2020-02-10
Remove admin tab-pane negative margin

### LINKED ISSUE #2412: Admin page tabs bottom margin ignored on some pages
On Admin pages that contain tabbed pages, there's a `<ul class="nav nav-tabs" role="tablist">` that adds `1.5rem` as bottom margin to the tabbed items, which is good for separating the bottom of the tabs and the top of the first element on the page.

However, on some pages, one or more the tabs show a different behavior, where the first element on the page doesn't respect the tabs' bottom margin and as a result there isn't enough separation:

Screen Shot 2020-02-07 at 6 11 35 PM

It happens on these pages/tabs:

- Curation > General: just Languages, others are okay
- Curation > Appearance: all tabs
- Curation > Search: Facets and Results tabs

Here'e the first tab in the same Configuration > Search menu as in the screenshot above:

Screen Shot 2020-02-07 at 6 11 57 PM

mejackreed:
Fixes #2412 

## After

Screen Shot 2020-02-10 at 11 49 58 AM

### DIFF
--- a/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
@@ -86,9 +86,6 @@ label.radio {
 }
 
 .tab-pane {
-  > p.instructions:first-child, > fieldset:first-child {
-    margin-top: -$padding-base-vertical;
-  }
   fieldset {
     margin-bottom: $padding-large-vertical * 2;
     margin-top: $padding-large-vertical * 1;


app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
```diff
@@ -86,9 +86,6 @@ label.radio {
 }
 
 .tab-pane {
-  > p.instructions:first-child, > fieldset:first-child {
```
mejackreed: I couldn't find a reason to keep this around

PR #2431, opened 2020-02-11
Add margin for zipr button fixes #2430

### LINKED ISSUE #2430: Add vertical spacing above "Show in ZPR viewer" button
The "Show in ZPR viewer" button currently has little or no spacing between it and the element (image or caption) above it:

Screen Shot 2020-02-11 at 1 42 35 PM

I suggest we add some SCSS so that the ZPR button has about `0.75rem` spacing between it and the element above it, which would change the example above to something like this:

Screen Shot 2020-02-11 at 1 46 23 PM

mejackreed:
Adds the requested 0.75rem margin.

Screen Shot 2020-02-11 at 4 05 19 PM

### DIFF
--- a/app/assets/stylesheets/spotlight/_pages.scss
+++ b/app/assets/stylesheets/spotlight/_pages.scss
@@ -267,4 +267,5 @@
 
 .zpr-link {
   @extend .btn-sm;
+  margin-top: $spacer * .75;
 }


app/assets/stylesheets/spotlight/_pages.scss
```diff
@@ -267,4 +267,5 @@
 
 .zpr-link {
   @extend .btn-sm;
+  margin-top: 0.75rem;
```
jkeck: Ah, I guess we can't use a `mt-` thing here because there isn't one that would end up being that.  Would `spacer * .75` be appropriate here?

mejackreed: Sure that makes sense. I've made the change.

PR #2437, opened 2020-02-12
add blacklight-oembed as a direct dependency of spotlight 

dunn:
it needs to be a dependency of spotlight itself to avoid loading errors during
generation (https://gist.github.com/dunn/e29ee0c32b1f1e03be471386da0b6ed3) and
it needs to be a dependency of the generated application so that its assets are
available to sprockets; otherwise we get

```
ActionView::Template::Error (couldn't find file 'blacklight_gallery/default' with type 'application/javascript'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,3 @@ cache:
   ##
   # Caching bundler here in a hope to reduce Travis test time
   bundler: true
-  ##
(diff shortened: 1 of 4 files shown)


dunn:
closing in favor of GH-2439

PR #2439, opened 2020-02-12
Disentangle oembed and blacklight-oembed from the spotlight engine

cbeer:


dunn:
the failure is because travis is caching the old test app; i deleted the cache and set it to re-run

PR #2442, opened 2020-02-12
Add a wrapper around the Title/Subtitle and apply stying from the...

jkeck:
...site-title to that.

The spacing and positioning needs to be on an element that contains both the title and subtitle.

(Before this change, subtitles would not show up when an exhibit had a masthead image applied)

### DIFF
--- a/app/assets/stylesheets/spotlight/_header.scss
+++ b/app/assets/stylesheets/spotlight/_header.scss
@@ -20,8 +20,12 @@
 
 
   .site-title {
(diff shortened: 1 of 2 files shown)


cbeer:
For some reason I thought not showing the subtitle was an intentional choice from way back in the day?

ggeisler:
@cbeer We do intentionally hide the subtitle on mobile viewports, but this was a problem where the subtitle was not visible in any viewport.

PR #2450, opened 2020-02-13
Remove the facet terms displayed as a tooltip on the info icon…

jkeck:
…(Facet config dashboard).

Closes sul-dlss/exhibits#1630

## Before
before

## After
after

### DIFF
--- a/app/views/spotlight/search_configurations/_facet_metadata.html.erb
+++ b/app/views/spotlight/search_configurations/_facet_metadata.html.erb
@@ -6,9 +6,4 @@
   <% else %>
     <%= content_tag :span, t(:'.value_count', count: metadata[:value_count]) %>
   <% end %>
-  <% if metadata[:terms].any? %>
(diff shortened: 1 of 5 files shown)


app/views/spotlight/search_configurations/_facet_metadata.html.erb
```diff
@@ -6,9 +6,4 @@
   <% else %>
     <%= content_tag :span, t(:'.value_count', count: metadata[:value_count]) %>
   <% end %>
-  <% if metadata[:terms].any? %>
```
cbeer: I'm guessing if we're removing this, there's some controller or model-land code that can also be thrown out?

jkeck: Yeah, I assume we'll need it again down the road for the issue that was spun out, but w/o knowing the data that we want to present, it can probably be safely removed and re-added later to fit our needs.

PR #2465, opened 2020-02-14
Move some of our BS form overrides to be applied to forms regardless …

### LINKED ISSUE #2447: Need vertical space on Curation > About, Add Exhibit Contact page
This is super minor but the fix for the problem in #2411 (PR #2414) didn't help the issue on the Curation > About, Add Exhibit Contact page:

Screen Shot 2020-02-12 at 4 29 20 PM

It looks like we have vertical spacing above the radio button, so maybe just adding some to the file upload input, to be similar what was done to add the vertical spacing in places like Configuration > Appearance, Exhibit Masthead, Image Source, upload radio button and file upload input:

Screen Shot 2020-02-12 at 4 41 28 PM

jkeck:
…if they are in a tab-pane or not.

Closes #2447 

## Before
contact-form-before

## After
contact-form-after

### DIFF
--- a/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
@@ -80,21 +80,22 @@ h1,h2,h3,h4,h5,h6 {
   }
 }
 
-.tab-pane {
-  fieldset {
-    margin-bottom: $spacer * 2;
-    margin-top: $spacer * 1;
-  }
-  legend {
-    margin-bottom: $spacer;
-  }
-  small.form-text {
-    margin-bottom: $spacer;
-    margin-top: $spacer * 0.3125;
-  }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
```diff
@@ -80,21 +80,22 @@ h1,h2,h3,h4,h5,h6 {
   }
 }
 
-.tab-pane {
```
jkeck: I _believe_ in the cases of the `<fieldset>`, `<legend>`, etc. we want the styles we've applied to be present regardless of whether it is in a tabbed display or not.

ggeisler: Makes sense to me. The After definitely looks like what we want.

PR #2468, opened 2020-02-14
Update bootstrap display utility classes for exhibit description …

### LINKED ISSUE #2443: Landing page card back description truncated unnecessarily at XL viewport
As part of the BL7/BS4 update, the truncation strategy for the description text on the back of the landing page card changed. Because the landing page cards are larger at the `xl` viewport width than the `lg` and `md` viewport widths, there is more room on the card for the description text at the `xl` viewport width. Previously we adjusted the truncation so that more description text was displayed in the `xl` viewport:

Screen Shot 2020-02-12 at 11 49 27 AM

As the example above shows, there is plenty of room to display the entire description on the card in the `xl` viewport width, but we're truncating it as if it were on the smaller, `lg` and `md` viewport width.

jkeck:
…truncation.

This technically closes #2443 (I think) and is based on some work that I done looking into it (and gets in more in-line w/ the previous behavior I believe), but not sure if this work will ultimately be done away with for other home-page design considerations. 🤷‍♂️

xs
sm
md
lg
xl

### DIFF
--- a/app/views/spotlight/exhibits/_exhibit_card_back.html.erb
+++ b/app/views/spotlight/exhibits/_exhibit_card_back.html.erb
@@ -11,8 +11,8 @@
     <p class="description">
       <%= truncated_description = exhibit.description.truncate(168, omission: '', separator: ' ') %>
       <% if exhibit.description.length > 168 %>
-        <%= content_tag(:span,"&hellip;".html_safe, class: "d-sm-block d-md-block") %>
-        <span class="d-sm-none d-md-none">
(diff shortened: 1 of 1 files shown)


jkeck:
@ggeisler it looks like we can get some wrapping on the `sm` breakpoint (which might just be due to the grid system updates in bootstrap).  Should we truncate at `sm` as well or keep as-is?

ggeisler:
@jkeck I think you can truncate at `sm` as well so we don't have to worry about that case. I think it's better to truncate than to overlap elements.

jkeck:
Okay, just updated to truncate at `sm` too.

PR #2481, opened 2020-02-20
Rework the home page exhibit cards

cbeer:
Part of https://github.com/sul-dlss/exhibits/issues/1712; some liberties were taken with the  design to align with bootstrap variables and sizes (which can be overridden by local implementations).

Screen Recording 2020-02-20 14_07_27


cbeer:
@jvine also note I had to add an extra link to the card image (the sliding feature interferes with the `stretched-link`).

jvine:
If the second link is handled in the same way it is in search results, that's fine with me.

PR #2493, opened 2020-02-26
Ensure that breadcrumbs are available without show search session

mejackreed:
I'm not sure if this is the correct way or not to handle this issue. But it seems previously for show actions, a search session was activated for that. But maybe that isn't happening anymore?

This is a work around, which aims to fix https://github.com/sul-dlss/exhibits/issues/1749 but might not be the correct way to do this. 

Is there somewhere we just are not creating search sessions for show page access anymore?

## Update 

This PR now does two things

1. Uses page referrers (or at least tries to) to determine where the heck we are linking from
1. Adds the page logic to preference the browse category logic since those search sessions can linger

## Update 2

Is this all not working because for some reason this isn't executing? 

https://github.com/projectblacklight/spotlight/blob/8eec0de68b8861c5847a4b3fb31a0a126e142bb7/app/controllers/spotlight/pages_controller.rb#L142-L146

### DIFF
--- a/app/controllers/spotlight/concerns/catalog_search_context.rb
+++ b/app/controllers/spotlight/concerns/catalog_search_context.rb
@@ -7,12 +7,10 @@ module Spotlight
     module CatalogSearchContext
       protected
 
+      ##
+      # Represents the current page referrer or from a search context
       def current_page_context
-        @current_page_context ||= if current_search_session_from_home_page?
-                                    current_exhibit.home_page if can? :read, current_exhibit.home_page
-                                  elsif current_search_session_from_page?
-                                    current_search_session_page_context
-                                  end
+        @current_page_context ||= current_page_from_page_context if current_page_from_page?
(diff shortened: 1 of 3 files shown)


app/controllers/spotlight/concerns/catalog_search_context.rb
```diff
@@ -35,24 +33,21 @@ def current_search_session_from_browse_category?
...
-        current_search_session &&
-          current_search_session.query_params['action'] == 'show' &&
-          current_search_session.query_params['controller'].ends_with?('_pages')
+      def current_page_from_page?
+        page_referrer &&
+          page_referrer[:action] == 'show' &&
+          page_referrer[:controller].ends_with?('_pages')
       end
 
-      def current_search_session_from_home_page?
```
mejackreed: I couldn't figure out a reason why we need this. We guard against this case elsewhere already.

mejackreed:
Closing in favor #2500

PR #2491, opened 2020-02-26
Revise exhibit card focus styles

camillevilla:
Part of sul-dlss/exhibits#1727
@jvine @ggeisler 

**There are two aims in this PR:**
- make focus styles for exhibit cards more consistent within the same browser (i.e. `outline:auto` is not used for `:focus` in WebKit browsers) 
- enhance native focus on exhibit cards where it is currently insufficient (i.e. Firefox) 

**How do we do this?** 
- unify styles where WebKit is available (i.e. Chrome, Safari) 
- default to Bootstrap focus styles where it's not available (i.e. Firefox)

**Out of scope:** 
- unifying focus styles across Spotlight (some work on this sul-dlss/exhibits#1745)

In the screenshots below I've artificially focused on both exhibit cards and tabs to contrast the exhibit card enhancements with the browser focus styles.

### Before (`outline:auto`)
chrome focus doesn't match

### After (`-webkit-focus-ring-color`)
chrome focus matches!

## Safari (unchanged) 
safari focused card
### Before (`outline:auto`)
firefox focus doesn't match

### After (Bootstrap focus color)
firefox focus enhanced

PR #2494, opened 2020-02-26
Enhance visibility of Firefox focus indicator globally

camillevilla:
This PR aims to unify the experience of focus styles within Firefox. We do this by overriding the default Firefox focus style so that it matches(ish) the Bootstrap focus style (see work on exhibit cards in #2491). @jvine and I paired on this so this just needs dev review. 

## Before: Firefox default, gray dotted lines
firefox-default-focus

## After: Bootstrap-esque, solid blue 5px border
firefox-enhanced-focus

### DIFF
--- a/app/assets/stylesheets/spotlight/_accessibility.scss
+++ b/app/assets/stylesheets/spotlight/_accessibility.scss
@@ -1,0 +1,8 @@
+// Global accessibility overrides
+
+// Increase visibility of Firefox focus indicator
+@-moz-document url-prefix() {
+  &:focus {
+    outline: $input-focus-border-color auto 5px;
+  }
+}
--- a/app/assets/stylesheets/spotlight/_spotlight.scss
+++ b/app/assets/stylesheets/spotlight/_spotlight.scss
@@ -31,6 +31,7 @@
 @import "spotlight/collapse_toggle";
 @import "spotlight/translations";
 @import "spotlight/utilities";
+@import "spotlight/accessibility";
 
 
 


cbeer:
I'm  not  sure how I feel about this in spotlight. I guess  it feels a little design-specific?

camillevilla:
Okayyyyyy, I could see the argument for doing this downstream in sul-dlss/exhibits. 

On the one hand...
I could see a case where an implementer could need to define two different focus styles for their application (i.e. one for light areas, one for dark areas) and they might trip over our styles trying to implement that. But we already waded into this by implementing an outline style for exhibit card with `:focus-within`. 

On the other hand... 
Not every institution that implements a Spotlight instance is going to have spare dev resources for implementing accessibility enhancements. Making the enhancement here for others' benefit might be nice. I don't think we're doing anything crazy by enhancing the focus style in the context of a local application.

More reading on focus styles here: 
https://hacks.mozilla.org/2019/06/indicating-focus-to-improve-accessibility/

jvine:
I mean... Bootstrap doesn't use the browser's default style. So the app already gives mixed signals depending on the type of element the user is focused on. We've added a third style to the mix with `:focus-within`, which gets into ransom-note territory. Users shouldn't be made aware of arbitrary distinctions between elements or made to wonder about their meaning. Making the focus styles consistent across the app is an accessibility improvement, and - I agree with Camille - a benefit to institutions who don't have dev resources (or awareness) to do this locally.

PR #2497, opened 2020-02-27
Only render subsections if there are documents to show

### LINKED ISSUE #2496: Omit the nested feature page sidebar list when there are no child pages
Related to #2495, in that in that PR I added bottom margin to the `ol.subsection` used in the feature page sidebar. That list is used only to display lists of child pages. However, we output the list whether or not there are child pages to put in the list:

Screen Shot 2020-02-27 at 3 59 14 PM

As a result, PR #2495 adds a small amount of unwanted vertical margin to the sidebar list of feature pages when there are no child pages.

### Ideal fix
In `feature_pages/sidebar.html.erb` we should only output the `<ol class="subsection">` element  ([line 6](https://github.com/projectblacklight/spotlight/blob/9cd1b139073ae023ffefc7df3a59786fa7438798/app/views/spotlight/feature_pages/_sidebar.html.erb#L6)) when there are child page list items to put in it. If we do that, the CSS selector I added to PR #2495 won't affect the sidebar spacing when there are no child feature pages.

### Alternative fix
We could use a `ol.subsection:empty` CSS selector to remove the unwanted bottom margin when there are no child pages in the list, except the `<ol class="subsection">` element is rendered in a way that it is technically not empty (i.e. has whitespace) so the `:empty` doesn't match it.

So if we can't avoid outputting the `<ol class="subsection">` element, if we could ensure it contains no whitespace we could fix the issue with one more line of CSS.

cbeer:
Fixes #2496

### DIFF
--- a/app/views/spotlight/feature_pages/_sidebar.html.erb
+++ b/app/views/spotlight/feature_pages/_sidebar.html.erb
@@ -3,11 +3,15 @@
     <% @exhibit.feature_pages.for_locale.published.at_top_level.each do |feature_section| %>
       <li class="<%= 'active' if current_page? [spotlight, @exhibit, feature_section] %>">
         <h2 class="h6"><%= link_to_unless_current feature_section.title, [spotlight, @exhibit, feature_section] %></h2>
-        <ol class="subsection">
-          <% feature_section.child_pages.published.each do |page| %>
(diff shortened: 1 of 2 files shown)


ggeisler:
Nice, thanks.

PR #2507, opened 2020-04-01
Replace the representable-based exhibit serializer with our own

cbeer:
This PR rips out the previous import/export routines and replaces it with a slightly simpler, if more verbose, version that just uses `as_json` (an a bit of post-processing) and activerecord magic to do the same thing.

This was motivated by https://github.com/sul-dlss/dlme/issues/893, which may have a simple fix with representable, but a) it's unclear where things aren't working right, and, b)the gem seems to be unmaintained and has some half-baked 3.x documentation.

There's some obvious opportunities for DRYing the code out, but unfortunately not a lot of consistency between our models, but I'd suggest attacking that as part of some additional feature development (like selective importing?).


app/services/spotlight/exhibit_import_export_service.rb
```diff
@@ -0,0 +1,379 @@
...
+                        },
+                        languages: {
+                          except: %i[id exhibit_id]
+                        },
+                        translations: {
+                          only: %i[locale key value interpolations is_proc]
+                        })
+          ).merge(
+            if_include?(:pages,
+                        searches: { # thumbnail
```
jkeck, line 322: Are searches directly associated w/ page exports?  It seems like one might want to be able to export search/browse categories for an exhibit that doesn't really use Feature/About pages and may want to exclude them?  (although perhaps I'm miss-understanding what functionality this bit of code is providing)

cbeer: This matches the previous behavior. My thinking at the time was to avoid weird breakage if you had a page that referenced a browse category, and that browse categories were a type of page (not a `Spotlight::Page`, but an HTML page).

PR #2513, opened 2020-04-07
Re-write Spotlight::Resources::Upload specs to be a bit more manageable.

jkeck:


app/models/spotlight/resources/upload.rb
```diff
@@ -16,11 +16,13 @@ def self.fields(exhibit)
         @fields ||= {}
         @fields[exhibit] ||= begin
           index_title_field = exhibit.blacklight_config.index.title_field
+          if index_title_field.is_a?(Blacklight::Configuration::Field)
+            index_title_field = index_title_field.tap { |f| f.normalize! unless f.field }.field
```
jkeck, line 20: This feel a little hokey. Happy to entertain alternatives.

cbeer: I'd almost say if the index title field isn't a string/symbol, the implementor ought to either ensure there's an explicit field or provide `config.upload_title_field`.

jkeck: I can see that.  In that case should we put a warning in the log if we have something that's not a string/symbol?

I'm not sure how prevalent using config objects for titles are but it was easy enough for us to one w/o realizing that we had broken the ability to add titles to uploaded documents (since if I remember correctly this also ends up w/ junk in the `name` attribute of the input).

cbeer: I'd say they're not prevalent -- we only added them for DLME to support the complex multilingual display rules.

PR #2519, opened 2020-04-10
Add Link URL attribute to the uploaded item row widget

### LINKED ISSUE #2034: Update Uploaded Item Row widget
An exhibit visitor can click on thumbnails of an _Exhibit item_ Item Row widget, because those lead to the exhibit item show page. But there is currently no option for the curator to make thumbnails of the _Standard item_ Uploaded Item Row widget items link to anything. Sometimes that's fine, but I believe there are plenty of use cases where the curator would like the exhibit visitor to actually go to another page/site represented by the uploaded item. Below is a suggestion for updating the Uploaded Item Row widget to enable this (also included is an unrelated minor issue that would probably be easy to do as part of this ticket):

uploaded-item-row-updates

### A - Update widget's intro text

Add sentence to end of existing intro text: "The item caption and link URL fields are also optional." So the full revised intro would read:

> This widget displays uploaded items in a horizontal row. Optionally, you can add a heading and/or text to be displayed adjacent to the items. The item caption and link URL fields are also optional.

### B - Add a new input field for Link URL

This is very similar to the caption field. The current caption input is not a wide as it could be, however, so I'd suggest making the input for the link URL field as wide as it can be in its current container, as shown in the mockup above.

If the curator does provide a link URL value for an item, when the widget is rendered the image thumbnail should be contained within a link tag so that clicking on the thumbnail image leads the user to that URL. 

For example, the items in the mockup above each have a caption and a link URL value. So when this widget is rendered like this:

1__before_you_start___exhibits_documentation_-_spotlight_at_stanford

Clicking on the, say, first thumbnail, would take the user to the site at https://exhibits.stanford.edu/american-enlightenment

(Note that we probably need to make a minor CSS adjustment related to this to ensure the rendered uploaded item thumbnail's still get a 1px border. I can do that later if necessary - would take less effort for me to make the CSS update than explain it here.)

### C - Extend the existing Caption input to be wider

As mentioned above, the caption field has more horizontal space available (first "C" in the mockup), so I suggest it be widened to match the width of the link URL field (second "C" in the mockup).

cbeer:
fixes #2034

Screen Shot 2020-04-10 at 06 56 53
Screen Shot 2020-04-10 at 06 58 29

### DIFF
--- a/app/assets/javascripts/spotlight/admin/blocks/uploaded_items_block.js
+++ b/app/assets/javascripts/spotlight/admin/blocks/uploaded_items_block.js
@@ -85,9 +85,13 @@ SirTrevor.Blocks.UploadedItems = (function(){
                 '</div>',
                 '<div class="main form-horizontal">',
                   '<div class="title card-title">' + dataTitle + '</div>',
-                  '<div class="field">',
+                  '<div class="field row mr-3">',
                     '<label for="' + this.formId('caption_' + dataId) + '" class="col-form-label col-md-3"><%= i18n.t("blocks:uploaded_items:caption") %></label>',
-                    '<input type="text" class="form-control" id="' + this.formId('caption_' + dataId) + '" name="item[' + index + '][caption]" data-field="caption"/>',
+                    '<input type="text" class="form-control col" id="' + this.formId('caption_' + dataId) + '" name="item[' + index + '][caption]" data-field="caption"/>',
+                  '</div>',
+                  '<div class="field row mr-3">',
+                    '<label for="' + this.formId('link_' + dataId) + '" class="col-form-label col-md-3"><%= i18n.t("blocks:uploaded_items:link") %></label>',
+                    '<input type="text" class="form-control col" id="' + this.formId('link_' + dataId) + '" name="item[' + index + '][link]" data-field="link"/>',
                   '</div>',
                 '</div>',
                 '<div class="remove float-right">',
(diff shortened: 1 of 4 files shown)


ggeisler:
Based on the screenshot it looks like the caption is linked, rather than the thumbnail image? It's been a while since I created the ticket (great to see it being addressed now) but I'm pretty sure the goal was to link the thumbnail image not the caption. The ticket included this:

> If the curator does provide a link URL value for an item, when the widget is rendered the image thumbnail should be contained within a link tag so that clicking on the thumbnail image leads the user to that URL.

Someone could have a lengthy caption, which won't look great as a link, and we don't link captions for other widgets, so I believe that was my rationale for linking the image, rather than the thumbnail.

An objection to this might be that the linked image is not as evident to the user as linking the caption but we do use the pattern of linking thumbnails rather than the caption throughout the application, so this would be consistent with that.

cbeer:
👍 Both the caption and image are currently linked. If we remove the caption text from the link, I think we need to provide the caption as an alt tag (and presumably do something to prevent screenreaders from reading the caption twice)?

ggeisler:
Yeah, I agree if we unlink the caption we should add the caption as alt text on the image.

At that point, I think the structure would follow what we currently do for the regular item row widget. So if we go with that as a first step, we're at least not creating a new problem, in terms of the screenreader issue.

But you raise a good question about the duplicate alt tag/caption value. I can talk to JV about that and whatever solution we might need, it seems we would apply to both the uploaded item row and regular item row widgets, as a new issue?

cbeer:
Sounds good; I've made that change.

The item row widget is a little different because it pulls the solr document title no matter what caption is showing.

PR #2520, opened 2020-04-14
Add YML export for locales beside the default language

### LINKED ISSUE #2516: YML import/export for languages by besides the default locale
This is a follow-on task from #2509. Currently, the YML import / export feature in `Curation > Translations` only exports the application's default locale. In the `.internal_test_app`, this is set to English. It would be good to add the ability to export other locales.

camillevilla:
Fixes #2516 

This PR builds on work from #2509 to make curator-provided translations exportable to YML files. This PR:
- Adds metadata field labels to YML export
- Updates UI for exporting YML: a user can export both the current language being translated and the default / source language
- adds `#translated_#{attr_name}` methods for looking up strings. My first pass at implementing this involved a lot of conditional logic in the `yamlbuilder`. 
  - Lookup for these strings can vary depending on several factors, such as:
    - whether the value is in the default locale or a translated language
    - whether a value has been provided in a translated language
    - whether the value uses Spotlight's default or has been customized by curator (e.g. facet field labels) 
  - empty string values have been provided because I18n throws a `MissingTranslationData` error when provided with nil, but ignores empty strings

### Before
translation interface with English export only

### After
translation interface with options to export to Spanish and English

### Sample YAML output - default language
```yaml
---
en:
  test:
    title: test
    subtitle: provided subtitle
    description:
```

### Sample YAML output - translated language
```yaml
---
es:
  test:
    title: titulo de ejemplo
    subtitle: ''
```

### DIFF
--- a/app/models/concerns/spotlight/translatables.rb
+++ b/app/models/concerns/spotlight/translatables.rb
@@ -15,8 +15,24 @@ module Spotlight
       # Set up a reader for the specified attribute that uses the I18n backend,
       # and defaults to the ActiveRecord value
       def define_translated_attr_reader(attr_name)
+        # Define a dynamic method for translating database-backed attributes,
+        # falling back to the database information as needed.
+        #
+        # Note: the empty string is provided as the final fallback to avoid i18n blowing
+        # up on nil attributes.
         define_method(:"#{attr_name}") do
-          I18n.translate(attr_name, scope: slug, default: attr_translation(attr_name))
+          send("translated_#{attr_name}", default: [attr_translation(attr_name), ''])
+        end
+
+        # Define an accessor that gets the value of the attribute in a given locale,
+        # returning `nil` for untranslated values.
+        #
+        # Note: For the default locale, we actually want to dig into the database,
+        # because that is the source of truth for the data.
+        define_method(:"translated_#{attr_name}") do |default: [], **options|
+          default = Array.wrap(default)
+          default.prepend(attr_translation(attr_name)) if I18n.locale == I18n.default_locale
+          I18n.translate(attr_name, scope: slug, default: default, **options).presence
         end
       end
     end
--- a/app/views/spotlight/translations/show.yaml.yamlbuilder
+++ b/app/views/spotlight/translations/show.yaml.yamlbuilder
@@ -1,20 +1,20 @@
-json.set!(I18n.locale) do
+json.set!(locale) do
   json.set!(current_exhibit.slug) do
-    json.title(current_exhibit[:title])
-    json.subtitle(current_exhibit[:subtitle])
-    json.description(current_exhibit[:description])
+    json.title(current_exhibit.translated_title(default: ''))
+    json.subtitle(current_exhibit.translated_subtitle(default: ''))
+    json.description(current_exhibit.translated_description(default: ''))
   end
 
   json.spotlight do
     json.curation do
       json.nav do
-        json.home(t(:'spotlight.curation.nav.home', locale: I18n.default_locale))
+        json.home(t(:'spotlight.curation.nav.home', default: ''))
       end
     end
 
     json.catalog do
       json.breadcrumb do
-        json.index(t(:'spotlight.catalog.breadcrumb.index', locale: I18n.default_locale))
+        json.index(t(:'spotlight.catalog.breadcrumb.index', default: ''))
       end
     end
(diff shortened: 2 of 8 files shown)


app/views/spotlight/translations/show.yaml.yamlbuilder
```diff
@@ -1,20 +1,20 @@
...
+    json.title(current_exhibit[:title]) if Translation.find_by(exhibit: current_exhibit, key: "#{current_exhibit.slug}.title", locale: @language).present?
+    json.subtitle(current_exhibit[:subtitle]) if Translation.find_by(exhibit: current_exhibit, key: "#{current_exhibit.slug}.subtitle", locale: @language).present?
+    json.description(current_exhibit[:description]) if Translation.find_by(exhibit: current_exhibit, key: "#{current_exhibit.slug}.description", locale: @language).present?
   end
 
   json.spotlight do
     json.curation do
       json.nav do
-        json.home(t(:'spotlight.curation.nav.home', locale: I18n.default_locale))
+        json.home(t(:'spotlight.curation.nav.home', locale: @language)) if try(t(:'spotlight.curation.nav.home', locale: @language))
```
camillevilla, line 11: This is also admittedly weird. I try tacking a `.present?` on the key lookup I get a translation missing error interpolated instead.
```
  spotlight:
    catalog:
      breadcrumb:
        index: '<span class="translation_missing" title="translation missing: es.spotlight.catalog.breadcrumb.index,
          locale: es">Index</span>'
```

config/locales/spotlight.en.yml
```diff
@@ -780,7 +780,8 @@ en:
       import:
         description: You can import and export Rails-based YAML files containing the translated label values for the currently selected language for use in internationalization tools such as Transifex and i18n-tools. This feature does not include importing or exporting translations for Pages.
         export_label: Export a YAML file
-        export_submit: Export translations
+        export_current_locale: Export %{language}
+        export_default_locale: Export %{language}
```
camillevilla, line 784: The values are equivalent here, but I created separate keys in case someone wants to override these downstream.

app/models/concerns/spotlight/translatables.rb
```diff
@@ -16,7 +16,13 @@ def translates(*attr_names)
       # and defaults to the ActiveRecord value
       def define_translated_attr_reader(attr_name)
         define_method(:"#{attr_name}") do
-          I18n.translate(attr_name, scope: slug, default: attr_translation(attr_name))
+          send("translated_#{attr_name}", default: [attr_translation(attr_name), ''])
+        end
+
+        define_method(:"translated_#{attr_name}") do |default: [], **options|
```
jkeck, line 22: Would it be possible to add a comment here as to what this dynamically defined method is doing?  I'm a bit miffed at the `default.prepend` here which seems like it would duplicate the value (since `attr_translation(attr_name)` is prepended and is the default) under the case that the current locale is the default locale.

I'm assuming I'm miss-reading/understanding the intent of the code, but perhaps a comment could clear that up.  I would also be happy to have it explained here in the PR and realize that I'm just confusing myself unnecessarily.

cbeer, line 22:
```suggestion
        # Define an accessor that gets the value of the attribute in a given locale,
        # returning `nil` for untranslated values. 
        #
        # Note: For the default locale, we actually want to dig into the database,
        # because that is the source of truth for the data.
        define_method(:"translated_#{attr_name}") do |default: [], **options|
```

app/models/concerns/spotlight/translatables.rb
```diff
@@ -16,7 +16,13 @@ def translates(*attr_names)
       # and defaults to the ActiveRecord value
       def define_translated_attr_reader(attr_name)
         define_method(:"#{attr_name}") do
```
cbeer, line 18:
```suggestion
        # Define a dynamic method for translating database-backed attributes,
        # falling back to the database information as needed.
        #
        # Note: the empty string is provided as the final fallback to avoid i18n blowing
        # up on nil attributes.
        define_method(:"#{attr_name}") do
```

app/models/concerns/spotlight/translatables.rb
cbeer: Yep, you got it. I've suggested a comment about why it appears doubled.

PR #2521, opened 2020-04-20
Update selector to be compatible with sassc 2.3.0

mejackreed:
🤷 https://github.com/sass/sass/issues/1873#issuecomment-152293725

cc @camillevilla 

Breaking spotlight builds using sassc 2.3.0

### DIFF
--- a/app/assets/stylesheets/spotlight/_accessibility.scss
+++ b/app/assets/stylesheets/spotlight/_accessibility.scss
@@ -2,7 +2,7 @@
 
 // Increase visibility of Firefox focus indicator
 @-moz-document url-prefix() {
-  &:focus {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/spotlight/_accessibility.scss
```diff
@@ -2,7 +2,7 @@
 
 // Increase visibility of Firefox focus indicator
 @-moz-document url-prefix() {
-  &:focus {
+  #{if(&, "&", "*")}:focus {
```
cbeer, line 5:
```suggestion
  :focus {
```

Looking at the output CSS, at least, the  `&` doesn't seem to be doing anything for us anyway (and we're not in a mixin context where we need dynamic binding or anything)

PR #2523, opened 2020-04-22
Pass format explicitly through page form/controller when updating.

jkeck:
Fixes sul-dlss/dlme#967

The polymorphic routing seems to choke on allowing us to handle both the has_many feature/about pages relation and the has_one home_page relation the same.

In the case of the has_one relationship we're seeing the "slug" or ID of the page being appended like a format to the URL.

If we explicitly pass a nil format through the form (which should always be HTML and is handled appropriately as the default) and then pass the format param explicitly back through the redirect after update (so updating via API will still work in theory) and things appear to work out both on the home and feature/about pages side of things.

## Before
page-update-format-before

## After
page-update-format-after

### DIFF
--- a/spec/features/home_page_spec.rb
+++ b/spec/features/home_page_spec.rb
@@ -31,6 +31,15 @@ describe 'Home page', type: :feature, versioning: true do
     expect(page).to have_selector '.card-title a', text: 'New Home Page Title'
   end
 
+  it 'redirects back to the correct home page URL after updating (no format param to interfere w/ faceting)' do
+    visit spotlight.exhibit_home_page_path(exhibit.home_page)
+
+    click_link 'Edit'
+    click_button 'Save changes'
+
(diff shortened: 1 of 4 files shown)


spec/features/home_page_spec.rb
```diff
@@ -31,6 +31,15 @@
     expect(page).to have_selector '.card-title a', text: 'New Home Page Title'
   end
 
+  it 'redirects back to the correct home page URL after updating (no format param to interfere w/ faceting)' do
+    visit spotlight.exhibit_home_page_path(exhibit.home_page)
+
+    click_link 'Edit'
+    click_button 'Save changes'
+
+    expect(page.current_url).to eq spotlight.exhibit_home_page_url(exhibit)
```
jkeck, line 40: I can also add an explicit test/step that clicks on a facet to ensure that it doesn't bomb out, but felt a little redundant.

PR #2526, opened 2020-04-28
Blacklist page  slugs that conflict with our routing

### LINKED ISSUE #2525: Routing conflict between new contacts and About pages named "Contacts"
Originally reported by @caaster in sul-dlss/exhibits#1873

To recreate: 
- Create a new About page named "Contacts". Edit and click `Save`. 
- You should get a 400 error for the `POST` request to `/exhibit-slug/about/contacts` because this route is also used when adding a new contact to the exhibit.

cbeer:
Fixes #2525

### DIFF
--- a/app/models/spotlight/page.rb
+++ b/app/models/spotlight/page.rb
@@ -7,7 +7,11 @@ module Spotlight
     MAX_PAGES = Spotlight::Engine.config.max_pages
 
     extend FriendlyId
-    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale]
+    # Note: This configuration also needs to be duplicated on the
+    # STI models ({Spotlight::AboutPage}, {Spotlight::FeaturePage}, {Spotlight::HomePage})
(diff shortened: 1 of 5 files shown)


app/models/spotlight/page.rb
```diff
@@ -7,7 +7,9 @@ class Page < ActiveRecord::Base
     MAX_PAGES = Spotlight::Engine.config.max_pages
 
     extend FriendlyId
-    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale]
+    friendly_id :title, use: %i[slugged scoped finders history], scope: %i[exhibit locale] do |config|
```
jkeck, line 10: We defined friendly_id on the `AboutPage` and `FeautrePage` classes as well so we would need to change it there too.   That means that we may only need to reserve `contacts` for AboutPages (🤷‍♂️).

We called this out in https://github.com/projectblacklight/spotlight/pull/2080#discussion_r199303425 but perhaps now would be a good time to validate that is still necessary.

cbeer: Nope, probably still true. I'll leave a code comment  here as a reminder too.

PR #2530, opened 2020-05-13
Add Transifex translations

jacobthill:
Manually add translations from Transifex. Most labels are curator facing and don't require translation. Transifex will only initiate a PR if all fields are translated and reviewed.


camillevilla:
Closing in favor of #2534

PR #2544, opened 2020-07-29
Define SirTrevor block types in Spotlight's engine config, and …

jkeck:
…consume that config in the sir_trevor_rails initializer.

This is intended to help w/ the upgrade path/backwards compatibility of the breaking change SirTrevor introduced in 0.6: https://github.com/madebymany/sir-trevor-rails#upgrade-guide-to-v060

### DIFF
--- a/lib/generators/spotlight/templates/config/initializers/sir_trevor_rails.rb
+++ b/lib/generators/spotlight/templates/config/initializers/sir_trevor_rails.rb
@@ -4,25 +4,7 @@
 # https://github.com/madebymany/sir-trevor-rails#upgrade-guide-to-v060
 class SirTrevorRails::Block
   def self.custom_block_types
-    %w[
-      Browse
-      FeaturedPages
-      Heading
-      Iframe
-      LinkToSearch
-      List
-      Oembed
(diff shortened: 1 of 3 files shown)


jkeck:
@dunn this is what I had in mind in https://github.com/projectblacklight/spotlight/pull/2371#pullrequestreview-453739154.  Do you see any potential issues w/ this approach?

dunn:
With this approach, unless the user pins sir-trevor to `< 0.6`, won't it fail on account of failing to find the generated initializer, if the user just upgrades in-place without running `install`?

jkeck:
I _think_ that. would be the same issue as we're already in since we're referencing the class that is generated by the initializer? 

It seems like either way, the downstream application needs to have our sir_trevor_rails initializer generated.  With the approach above, if a user had already defined their own custom blocks using `Spotlight::Engine.config.sir_trevor_widgets` (which was the recommendation) then that customization will continue to work instead of requiring them to both run the generator and reconcile their customization in the new class.

dunn:
@jkeck ah, I understand.  This sounds good then.

PR #2552, opened 2020-08-07
Update layout of Feedback (Report a problem) form

### LINKED ISSUE #2543: Update layout of Feedback (Report a problem) form
Below are recommendations for updating the layout of the Feedback/Report a problem form. The basic design of the current form is six years old and when recently evaluating our feedback forms at Stanford we came up with some ideas for modernizing it a bit. These recommendations don't change the actual data that is posted from the form, just rearranges elements, renames labels, and adds a couple of elements.

Screen Shot 2020-07-14 at 4 07 23 PM

---

- [x] **A:** Change the gray background color to be the same as the `<body>` color (by default, `#fff` I believe)

- [x] **B:** Remove the form heading. Since the form opens immediately below the "Feedback" link that triggers it, it is likely very clear to user what the form is for without a heading.

- [x] **C:** Reorder the form fields and rename the "Name" and "Email" fields. So the updated name labeling and order is "Message", "Your name", and "Your email". The "Message" field could be reduced in height, as well.

- [x] **D:** Change the alignment of the action buttons to be left-aligned, "Send" and then "Cancel".

In addition to those changes, there are two recommended additions:

Screen Shot 2020-07-14 at 4 13 46 PM

---

- [x] **E:** Add a Bootstrap `alert-primary` component to the top of the form, which contains a string with the URL of the current page ("Reporting from: _current page URL_"). This helps the user understand that we know which page they're reporting from, so they don't have to include it in their feedback message.

- [ ] **F:** Assuming it's feasible to add a Captcha in Spotlight, versus having to add it in a localized repo (e.g., Stanford Exhibits), for non-logged-in users, add a Captcha. 

In the mockup the text below the Captcha says "(Stanford users can avoid this by logging in.)", which is obviously a localized string. The default string could be "(You can avoid this Captcha by logging in.)" so if an implementer forgets to localize it, it will still make sense to the user.

cbeer:
Fixes #2543 
Screen Shot 2020-08-07 at 11 58 14

### DIFF
--- a/app/views/spotlight/shared/_honeypot_field.html.erb
+++ b/app/views/spotlight/shared/_honeypot_field.html.erb
@@ -1,0 +1,4 @@
+<span style="display:none;visibility:hidden;">
+  <% honeypot_field_name = Spotlight::Engine.config.spambot_honeypot_email_field %>
+  <%= f.email_field honeypot_field_name, label: t(:'spotlight.shared.report_a_problem.honeypot_field_explanation') %>
+</span>
--- a/app/assets/stylesheets/spotlight/_report_a_problem.scss
+++ b/app/assets/stylesheets/spotlight/_report_a_problem.scss
@@ -1,7 +1,10 @@
 #report-problem-form {
   display: none;
-  background-color: #e7e7e7;
-  border-bottom: 1px dotted $gray-600;
+  border-bottom: 1px solid $gray-600;
+
+  h2 {
+    @extend .sr-only;
+  }
 }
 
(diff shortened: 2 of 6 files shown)


app/views/spotlight/shared/_honeypot_field.html.erb
```diff
@@ -0,0 +1,4 @@
+<span style="display:none;visibility:hidden;">
```
mejackreed, line 1: This is nice to be able to override now

PR #2579, opened 2020-09-25
Reserve the exhibit slug "site" so it does not conflict with site…

jkeck:
… wide admin pages.

### DIFF
--- a/app/models/spotlight/exhibit.rb
+++ b/app/models/spotlight/exhibit.rb
@@ -22,10 +22,15 @@ module Spotlight
     paginates_per 48
 
     extend FriendlyId
-    friendly_id :title, use: %i[slugged finders]
+    friendly_id :title, use: %i[slugged finders] do |config|
+      config.reserved_words.concat(%w[site])
+    end
+
     validates :title, presence: true, if: -> { I18n.locale == I18n.default_locale }
     validates :slug, uniqueness: true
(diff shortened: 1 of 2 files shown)


ggeisler:
Nice @jkeck. This could also be handy for implementers who have certain institutional-related slugs they don't want used for their exhibits whatever reason.

Super minor but shouldn't the error message in the test by something like: `Validation failed: slug is reserved` instead of `Validation failed: Friendly is reserved`?

jkeck:
@ggeisler, there are already some hooks that a downstream consumer could use to do this sort of thing in theory (although that's buried in the FriendlyID documentation so might not be obvious).

The error text is a mix of Friendly ID's defaults + i18n, which luckily means we can change it.  The way that this ultimately ends up being displayed to the user is a bit different though.  So, even w/ the change to `slug` in the text, the user will end up seeing something like this: 

Screen Shot 2020-09-25 at 1 05 37 PM


Is that okay, or should we dig into the i18n config and change this from the default for spotlight core? (Downstream consumers will be able to modify the i18n config in theory to make this say whatever they want)

ggeisler:
It makes sense to me to go with what you have for the UI error message. Worst case `is reserved` is pretty clear, and best case is implementers customize the string, as you suggest, to use text that makes sense at their institution. And this is only visible to superadmins, I believe, so presumably only a small number of relatively knowledgeable Spotlight users would ever see it.

mejackreed (approved):

PR #2580, opened 2020-11-05
Change label to Remove from exhibit

elrayle:
Label `helpers.action.spotlight/role.destroy` is set to `Remove from site` but is used on the Exhibit Dashboard -> Users list when the action icon is clicked to allow a user to be removed from having a role in the exhibit.  I could not find it used anywhere else to remove a user from the entire site.

The action that occurs when the button is clicked is the user is removed from the current exhibit.  It has no impact on the user’s roles in other exhibits or their access to the site.  The current label is confusing.

### DIFF
--- a/config/locales/spotlight.en.yml
+++ b/config/locales/spotlight.en.yml
@@ -46,7 +46,7 @@ en:
         create: Add new field
       spotlight/role:
         create: Add a new user
-        destroy: Remove from site
+        destroy: Remove from exhibit
       spotlight/search:
         destroy: Delete
         edit_long: Edit this saved search


ggeisler:
👍 I suspect this is mostly confusing only to superadmins/developers, rather than the actual exhibit admins/curators who are unlikely to know about the distinction between `site` and `exhibit` (at least in Stanford's implementation, exhibit users are unlikely to think they have power to affect anything beyond the current exhibit), but "Remove from exhibit" seems like a better label for everyone regardless.

elrayle:
We have a local customization to address this.  Feel free to close if you don't want a similar update to Spotlight.

ggeisler (commented):
This looks good to me, but the build is failing so I don't want to merge just because I don't understand enough about the build issues to know if that's a good idea or not (obviously this is a simple commit that shouldn't affect cause the build to fail). Maybe you @elrayle or someone else can get the build to pass and then we can merge this.

PR #2586, opened 2020-12-09
add docker-compose environment

dunn:
Following @no-reply's work on Hyrax dockerization, this adds a generated test app in `.testapp` and configures a Docker environment so that development can be done using docker-compose.

The test app gets up an running, but currently I'm running into an issue with uploading files:
```
web_1         | F, [2020-12-09T00:24:12.982526 #1] FATAL -- :
web_1         | Riiif::ImageNotFoundError (Riiif::ImageNotFoundError):
web_1         |
web_1         | riiif (bdfa71037571) app/resolvers/riiif/abstract_file_system_resolver.rb:18:in `path'
web_1         | riiif (bdfa71037571) app/resolvers/riiif/abstract_file_system_resolver.rb:11:in `find'
web_1         | riiif (bdfa71037571) app/models/riiif/image.rb:33:in `file'
web_1         | riiif (bdfa71037571) app/models/riiif/image.rb:53:in `info'
web_1         | /spotlight/engine/app/services/spotlight/upload_solr_document_builder.rb:28:in `add_image_dimensions'
web_1         | /spotlight/engine/app/services/spotlight/upload_solr_document_builder.rb:14:in `block in to_solr'
web_1         | /spotlight/engine/app/services/spotlight/upload_solr_document_builder.rb:9:in `tap'
web_1         | /spotlight/engine/app/services/spotlight/upload_solr_document_builder.rb:9:in `to_solr'
web_1         | /spotlight/engine/app/services/spotlight/solr_document_builder.rb:17:in `documents_to_index'
web_1         | /spotlight/engine/app/models/spotlight/resource.rb:53:in `block (2 levels) in reindex'
```
I'm having trouble diagnosing this one because it should be using the CarrierwaveFileResolver, but apparently isn't.


mejackreed (commented):
Thanks for pushing up a WIP of this.

I think it is interesting to see where this goes in other projects that are Rails engines, maybe they can pave the way. But I would be concerned about modifying the approach to using a local rails application for testing rather than our engine cart usage and the new skill learning that will require for Spotlight and Spotlight developers.

While I can see the benefit of some of the dependencies can easily be installed using this type of setup, I personally would not be comfortable moving Spotlight towards a docker-based development and testing workflow. Maybe others are?

How was that decision made in Hyrax? Are there other rails engines that do similar things?

no-reply:
>  But I would be concerned about modifying the approach to using a local rails application for testing rather than our engine cart usage and the new skill learning that will require for Spotlight and Spotlight developers.

it might be worth decoupling the test app hosting from the docker question. there's no theoretical reason this compose environment couldn't use a generated `engine_cart` application. we chose to do these things together in Hyrax because they both seemed attractive and doing them both saved work on each.  @dunn it's not obvious to me how much of what's circumscribed by "add test app" is just `rails new`, and how much was manual action. if the changes to the generated app are pretty isolated, it wouldn't be a huge burden to add them to the engine cart generation.

the Hyrax strategy is to use both. we have `engine_cart` baked into the CI process, which tests app generation. but also provide a more stable test app for development purposes, which helps us take impact on long-term applications (not just ephemeral ones) into account when making engine changes. normal development cycles run the test suite on both. so far, this seems to be working well.

> While I can see the benefit of some of the dependencies can easily be installed using this type of setup, I personally would not be comfortable moving Spotlight towards a docker-based development and testing workflow. Maybe others are?

can you say more about your concerns? are you against having a `Dockerfile` and `docker-compose` in the repository, or just worried about the pre-generated application? other?

> How was that decision made in Hyrax? Are there other rails engines that do similar things?

how much do you want to know? :grinning:

mejackreed:
Thanks for providing that context about the Hyrax approach especially "which helps us take impact on long-term applications (not just ephemeral ones) into account when making engine changes".  Is that https://github.com/samvera/hyrax/tree/master/.dassie ?

> can you say more about your concerns?

Sure thing, but this is just my opinion and I would interested in hearing other viewpoints in the community.

> are you against having a Dockerfile and docker-compose in the repository

Not necessarily, but I'm concerned about having the resourcing to maintain these additional development options. There was some work in Blacklight to support a Docker based Solr, based on community input. But that took a while to get right, and its unclear to me about the long term viability of that. We have a smaller developer community in Spotlight and the team I'm on largely does not do Docker based development. It would introducing an alternative that some of the core maintainers do not use regularly. With a potential upside of increasing maintainers/developers by providing Docker?

> or just worried about the pre-generated application

Yes some concerns here too just about the extra maintenance footprint of this code. 

Both of these things I think could improve the developer experience, but I don't think either are yet fully socialized as where we may see these community improvements. Because of this, I'm hesitant, and really am curious about others thoughts and opinions here.

The team I'm on also helps to maintain a lot of other Blacklight engines (RangeLimit, Advanced Search, Gallery, oEmbed, Heatmaps, etc). A really useful benefit is having consistency across these projects in how they are setup and run. The ripple effect of divergence here trickles into Spotlight eventually.

Moving this forward, it seems like it would be useful to gain consensus on the benefits and risks of these changes separately so we can determine if there is a path forward.

> how much do you want to know? 😀


🙉 maybe I don't :) But I realize I don't know everything here and if we can benefit the project by gaining from other's experience.. I'm for that for sure.

dunn:
i think i fixed the config upload, but now i'm getting the (same?) riiif error again:
```
web_1         | W, [2021-03-10T01:52:50.972058 #7]  WARN -- : [ActiveJob] [Spotlight::ReindexJob] [e0d97a16-ecdc-4fbb-9721-20c59767af98] [Spotlight::Etl::Pipeline] [#<Spotlight::Resources::Upload id=1>] [#<Proc:0x0000565176e05e98 /spotlight/engine/app/services/spotlight/etl/transforms.rb:14 (lambda)>] Invalid characters in id `1`
web_1         | E, [2021-03-10T01:52:50.972565 #7] ERROR -- : [ActiveJob] [Spotlight::ReindexJob] [e0d97a16-ecdc-4fbb-9721-20c59767af98] [Spotlight::Etl::Pipeline] [#<Spotlight::Resources::Upload id=1>] [#<Proc:0x0000565176e05e98 /spotlight/engine/app/services/spotlight/etl/transforms.rb:14 (lambda)>] Caught exception Riiif::ImageNotFoundError
```

dunn:
i forgot i ran into this error last year and fixed it in riiif, but forgot to open the pr: https://github.com/curationexperts/riiif/pull/138

dunn:
got image uploads working, but it required a few upstream patches that have yet to be merged, so this is blocked by

- https://github.com/bitnami/bitnami-docker-solr/pull/25
- https://github.com/curationexperts/riiif/pull/138
- https://github.com/curationexperts/riiif/pull/134

dunn:
got the riiif PR merged, so i think this is good to go? @no-reply @mejackreed

lib/generators/spotlight/templates/config/initializers/translation.rb
```diff
@@ -5,19 +5,20 @@
...
 Translation = I18n::Backend::ActiveRecord::Translation
 
-ActiveSupport::Reloader.to_prepare do
-  if Translation.table_exists?
-    ##
-    # Sets up the new Spotlight Translation backend, backed by ActiveRecord. To
-    # turn on the ActiveRecord backend, uncomment the following lines.
+# Don't allow initializer to break if DB doesn't exist yet
+# see: https://github.com/projectblacklight/spotlight/issues/2133
+if ENV['SKIP_TRANSLATION'].blank?
```
dunn, line 10: this is a solution that @mcritchlow figured out for the Starlight fork: https://gitlab.com/surfliner/surfliner/-/commit/aa0a8183246326e3317997c7f19fe2285087b321

dunn:
not sure what to make of this failure: https://github.com/projectblacklight/spotlight/runs/6041994611?check_suite_focus=true#step:5:5007
```
  1) I18n does not have unused keys
     Failure/Error:
       expect(unused_keys).to be_empty,
                              "#{unused_keys.leaves.count} unused i18n keys, run `i18n-tasks unused' to show them"

       5 unused i18n keys, run `i18n-tasks unused' to show them
```
the allegedly unused keys are
```
Unused keys (5) | i18n-tasks v1.0.9
+--------+-------------------------------------------------------+------------------+
| Locale | Key                                                   | Value            |
+--------+-------------------------------------------------------+------------------+
|   en   | spotlight.job_trackers.show.messages.status.completed | job completed    |
|   en   | spotlight.job_trackers.show.messages.status.enqueued  | job enqueued     |
|   en   | spotlight.job_trackers.show.messages.status.failed    | job failed       |
|   en   | spotlight.job_trackers.show.messages.status.missing   | (missing status) |
|   en   | spotlight.job_trackers.show.messages.status.pending   | job pending      |
+--------+-------------------------------------------------------+------------------+
```
but when i remove them i get missing key errors

mcritchlow:
> not sure what to make of this failure: https://github.com/projectblacklight/spotlight/runs/6041994611?check_suite_focus=true#step:5:5007

@dunn - Seeing that in both PRs I put up as well last week

dunn:
i'm just going to go ahead and mark it pending like the other i18n spec

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+  yarn \
+  zip
+
+RUN addgroup --gid 10001 --system spotlight && \
+  adduser --uid 10000 --system \
+  --ingroup spotlight --home /spotlight spotlight
+
+USER spotlight
+RUN gem update bundler
+RUN gem install --no-document rails -v '< 6.1'
```
mcritchlow, line 33: Should we make the rails version an `ARG`?

dunn: oh yeah good idea

mcritchlow (approved):

PR #2588, opened 2020-12-10
Replace deprecated update_attributes with update

mejackreed:
Removed in Rails 6.1


camillevilla (approved):
https://edgeguides.rubyonrails.org/6_1_release_notes.html#active-record-removals
Other Rails 6.1 work in #2589

PR #2590, opened 2021-01-12
Try to fix routing issue in locale picker triggered by bots[WIP]

camillevilla:
https://github.com/sul-dlss/dlme/issues/932


jkeck:
Ugh, after looking at this a little more closely in DLME, I think that it can generate some incorrect URLs in some instances. 😢

cbeer:
Replaced by #2637

Issue #2567, opened 2020-08-19
Display published browse groups as filters on Browse landing page


ggeisler:
Related to #2562.

If an exhibit has no published browse groups, nothing (in terms of what the exhibit visitor sees) visibly changes from how the browse categories work today. The browse landing page looks and works the same way.

If an exhibit curator has added browse groups and published any of them, however, the browse landing page is affected.

By _affected_, I mean that filter buttons will be displayed at the top the browse landing page and selecting a filter button will change the browse categories displayed on the page. This is the same interaction as on the Spotlight _site_ landing page, where filter buttons can be used to display exhibits that match exhibit tags, and I suggest we use the same layout and styling (the styling in my browse group mockups looks different from the screenshot below because they reflect Stanford's Spotlight theme styling):

#### Spotlight site landing page

<img width="1152" alt="Screen Shot 2020-08-19 at 10 56 45 AM" src="https://user-images.githubusercontent.com/101482/90672286-be640e00-e20a-11ea-948f-e20a08c5e8f4.png">

---

#### Proposed browse landing page with published browse groups

<img width="1024" alt="Screen Shot 2020-08-19 at 10 45 59 AM" src="https://user-images.githubusercontent.com/101482/90674619-37b13000-e20e-11ea-8b80-2aee73825fde.png">

---

## Strategy for displaying browse group filters

- If at least one browse group is published and has at least one browse category assigned to it, display the filter buttons. (We could add the additional requirement that a published browse group must have at least one published browse category assigned to it, to prevent an empty result when that group filter is selected, but I'd suggest we keep it simpler and rely on the exhibit curator to know not to publish browse groups without assigned categories.)

- When displaying the filter buttons, always display an "All" filter first. The "All" filter is the default selection when navigating to the browse landing page without any query parameters that contain a specific filter value.

- The "All" filter displays all published browse categories in order, just as the browse landing page currently does.

- For each published browse group, display a filter button. The order of the filter buttons should follow the order of the browse group cards on the Curation > Browse page.

- When a browse group's filter button is selected, display only the published browse categories that are currently assigned to that browse group.

- The implementation should ensure that we can link to the browse landing page with a specific filter selected, as a query parameter (or whatever makes the most sense technically). This is probably necessary for a browse group widget, and the View action on the browse group card on the Curation > Browse page. We currently do this for the site exhibit tags, e.g.: `http://localhost:3000/spotlight/?tag=Installations`

## Example
[...]

Issue #2562, opened 2020-08-18
Add new feature: Browse groups


ggeisler:
This is a story ticket for a new feature. These are the specific work tickets:

- [x] #2594 - Add tabs to Curation > Browse admin page
- [x] #2596 - Enable browse group creation and management
- [x] #2597 - Enable associating a browse category with a browse group 
- [x] #2567 - Display published browse groups as filters on Browse landing page

Not required to complete this new feature, but would take advantage of it:

- #2568 - Add a browse group widget to Exhibit item widgets

As an overview, the problem this proposed new feature addresses is that at Stanford we've found that some exhibits contain many browse categories and that subsets of those categories can relate to each other by theme. For example, in this exhibit: https://exhibits.stanford.edu/bacon/browse the exhibit curator prefixed the titles of browse categories "Farm workers:" and "Immigrants:" as a way to show which browse categories are related by theme. But since we show a flat list of all browse categories on the browse landing page, it is not as easy of it might be for the exhibit visitor to pick out themes of interest from a page with many browse categories.

The proposed browse group feature would enable the exhibit curator to **optionally** group browse categories into themes, or groups. The browse group feature can be ignored by exhibit curators who don't need it and the existing browse categories feature shouldn't be affected at all if browse groups are not used in an exhibit.

If an exhibit curator does want to use the browse groups feature, they would organize their browse categories into groups by first creating browse groups (ignore the callouts on the mockups below; they are referenced in the specific work tickets listed at the top of this comment):

<img width="766" alt="Browse groups tab - active groups" src="https://user-images.githubusercontent.com/101482/104525047-98c21180-55bc-11eb-9bb1-541b5e670c36.png">

---

And then associating a browse category with one or more browse groups:

<img width="788" alt="Group tab - groups checked" src="https://user-images.githubusercontent.com/101482/104525101-b68f7680-55bc-11eb-91c7-55972754101c.png">

---

When a browse group is published and contains child browse categories, the browse landing page shows a row of browse group filter buttons at the top. When the exhibit visitor selects a browse group filter, the browse landing page displays only the browse categories that belong to that browse group. For example, below the selected browse group is "Medium", which contains five browse categories:

<img width="991" alt="Screen Shot 2020-08-18 at 10 08 14 AM" src="https://user-images.githubusercontent.com/101482/90546671-3c5ae300-e13f-11ea-8e8c-7a0f7ab033f6.png">

mejackreed:
[...]

PR #2598, opened 2021-01-14
Add Spotlight::Group and Spotlight::GroupMember models

cbeer:
Part of #2562 

This sets up models and associations to "group" browse categories. The group modeling sets us up for potential polymorphic membership (or, more likely, model re-use for grouping other types of things), but Rails doesn't care for polymorphic `has_many` relations, so we're stuck with the database model being ready for it but with Rails accessors that need to be targeted at specific models 🤷‍♂️ 

For future compatibility, the join table for group members is made explicit in anticipation of emerging or future requirements  (perhaps ordering group members or other features that require some metadata on the membership relation).

TODO:
- [x] model tests
- [x] adding the user-facing piece #2567 to validate the modeling

### DIFF
--- a/app/controllers/spotlight/browse_controller.rb
+++ b/app/controllers/spotlight/browse_controller.rb
@@ -9,7 +9,8 @@ module Spotlight
     include Spotlight::Catalog
     include Blacklight::Facet
 
-    load_and_authorize_resource :search, except: :index, through: :exhibit, parent: false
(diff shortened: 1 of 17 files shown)


mejackreed:
Updated now and fixes #2567 

Kapture 2021-01-20 at 15 53 22

jkeck (approved):
👍  This looks good to me

PR #2609, opened 2021-01-22
Fixes #2597 by associating BrowseCategories with Groups

### LINKED ISSUE #2597: Enable associating a browse category with a browse group
Part of #2562. Supersedes #2565.

After a curator creates browse groups, they need to be able to associate browse categories with those browse groups. This tickets proposes that is done by editing a browse category and selecting the group(s) it should be associated with.

## Add new tab to Curation > Browse, Edit browse category page.
Because the actions involved in associating a browse category with a browse group are distinct from the category description fields, and because many curators will not be interested in using the browse groups feature, I'm suggesting this feature belongs on a new, dedicated tab (Group) on the Edit browse category page:

Group tab - no groups defined

---

## State of tab content when no browse groups have been created
- **Intro text:** If the curator hasn't yet created any browse groups, there isn't anything they can do on the new Group page, but we should indicate what is possible (which will help to educate curators that this feature exists). So if no browse groups exist, the intro text is:

> You cannot add this browse category to a browse group because no browse groups have been created. To create a browse group, use the Browse groups tab on the Curation > Browse page.

- **Save/cancel buttons:** These aren't useful on this page when there are no browse groups, but I believe these buttons apply to changes made on any tab within the Edit browse category pages, so we need to display them even in this state.

## State of tab content when there are existing browse groups
When there are existing browse groups in the exhibit, the intro text is different and there are checkboxes below the intro text:

Group tab - groups checked

---

- **A - Intro text:**  

> You can add this browse category to any number of browse groups. When you add a browse category to a group it will be displayed when an exhibit visitor selects that browse group’s filter button on the Browse landing page. 

- **B - Checkboxes:** For each browse group defined in the exhibit, display a checkbox and associated label.

- Checkbox labels are the browse group title
- Checkboxes are displayed in the same order as the curator has ordered them on the Curation > Browse groups page
- Default state of each checkbox is Off
- If the curator changes the checkbox for a group to On, the browse category should be associated with that browse group, where associated means it will be included on the browse landing page when filtered by that browse group (as detailed in #2567), if the browse category is published.
- A published browse category is always displayed on the browse landing page when the All filter is active; its inclusion in any associated browse groups is additional to, not instead of, inclusion in the All filter (I'm sure this is obvious, just being clear that the browse group filters and category inclusion behavior is exactly the same as the site exhibits landing page filters and exhibit inclusion behavior)
- The user must select the Save changes button to save state changes to the checkboxes

mejackreed:
## No groups present
Screen Shot 2021-01-22 at 7 34 32 AM

## Groups are present
Kapture 2021-01-22 at 07 35 32

### DIFF
--- a/app/views/spotlight/searches/_form.html.erb
+++ b/app/views/spotlight/searches/_form.html.erb
@@ -17,6 +17,10 @@
       </li>
 
       <li role="presentation" class="nav-item">
+        <a href="#search-group" aria-controls="search-group" role="tab" data-toggle="tab" class="nav-link"><%= t(:'.search_group') %></a>
+      </li>
+
+      <li role="presentation" class="nav-item">
         <a href="#search-masthead" aria-controls="search-masthead" role="tab" data-toggle="tab" class="nav-link"><%= t(:'.search_masthead') %></a>
       </li>
 
@@ -48,6 +52,14 @@
               </div>
             </div>
           </div>
+        <% end %>
+      </div>
+      <div role="tabpanel" class="tab-pane" id="search-group">
+        <% if @groups.present? %>
+          <p class="instructions"><%= t(:'.group.help') %></p>
+          <%= f.collection_check_boxes(:group_ids, @groups, :id, :title, hide_label: true) %>
+        <% else %>
+          <p class="instructions"><%= t(:'.group.help_no_groups') %></p>
         <% end %>
       </div>
       <div role="tabpanel" class="tab-pane" id="search-masthead">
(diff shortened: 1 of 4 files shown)


app/views/spotlight/searches/_form.html.erb
```diff
@@ -50,6 +54,14 @@
           </div>
         <% end %>
       </div>
+      <div role="tabpanel" class="tab-pane" id="search-group">
+        <% if @groups.present? %>
+          <p class="instructions"><%= t(:'.group.help') %></p>
+          <%= f.collection_check_boxes(:group_ids, @groups, :id, :title, hide_label: true) %>
```
mejackreed, line 60: 4uu4x1

PR #2611, opened 2021-01-23
Creates a BrowseGroupCategories SirTrevor widget

mejackreed:
Fixes https://github.com/sul-dlss/dlme/issues/1145

Paired w/ @jkeck 

**NOTE**

This differs from the designs in a few ways:
 1. We cannot have an individual `Include category item counts?` checkbox for each individual page, only per widget box
 2. We don't yet have a way to restrict the number of autocompletable  items in a widget
 3. We advise not to store category counts in the SirTrevorBlocks on the admin page, as these will not update as a Browse group is modified.

Follow on work in https://github.com/sul-dlss/dlme/issues/1142 will address the display on a `Spotlight::Page`

Kapture 2021-01-22 at 17 12 32


ggeisler:
The deviations from the design sound fine, although not being able to show the number of categories in a group is a bummer. That could help the curator understand how that group's row is going to be experienced by the user (4 categories versus 20 categories will feel a lot different to the user). But they can always look at the Browse filter page for the group or go to the admin to figure it out.

A couple of visual design things: 

- Can we right-align the Remove link with the right-edge of the card container (with 20px or so horizontal space between)? It just feels a bit weird that the Remove link is kind of hanging out 2/3 of the way across the container. See the mockups for the preferred alignment.
- The prompt for the autocomplete box should be `Enter a browse group title` (trying to be more specific here than usual because of the likely confusion between "browse categories" and "browse groups"). That wording is trying to reinforce to the user that what this widget is based on is browse groups and not categories.

mejackreed:
Thanks @ggeisler I've updated to incorporate this.

Screen Shot 2021-01-22 at 6 11 12 PM

ggeisler:
Great @mejackreed. Except that change seems to have affected the available space for the browse title ("very very good" now wraps)?

We definitely want to keep the title on one line as much as possible.

mejackreed:
Nice catch @ggeisler .. repushed now adding a flex-grow to fill the space better

Screen Shot 2021-01-25 at 6 22 25 AM

PR #2617, opened 2021-01-27
Adds the ability to provide curator translations to Browse Groups

mejackreed:
Not specified yet in a ticket I could find, but provides a similar interface to Browse Categories.

Screen Shot 2021-01-27 at 6 21 43 AM

Kapture 2021-01-27 at 06 13 18

### DIFF
--- a/app/views/spotlight/translations/_groups.html.erb
+++ b/app/views/spotlight/translations/_groups.html.erb
@@ -1,0 +1,34 @@
+<div role="tabpanel" class="tab-pane <%= 'active' if @tab == 'groups' %>" id="groups">
+  <%= bootstrap_form_for current_exhibit, url: spotlight.exhibit_translations_path(current_exhibit), layout: :horizontal do |f| %>
+    <% # Add hidden fields for the language and tab so the redirect knows how to come back here %>
+    <%= hidden_field_tag :language, @language %>
+    <%= hidden_field_tag :tab, 'groups' %>
+
+    <% current_exhibit.groups.each do |group| %>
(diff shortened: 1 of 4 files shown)


cbeer (approved):

ggeisler (commented):
Thanks for working on this without a ticket, though I think it might have been more efficient to create a design needed ticket before working on this. This approach seems based on how the Browse categories translation page is set up, but since there is only one string per group label that needs to be translated (versus three related fields with browse categories), I think a better model would've been the General, Metadata, or Search field pages. In any case, here are some suggestions:

- Move the Browse groups tab to come after the Browse categories tab
- Remove the "English language title" and "[current language] translation" headings (that row is not really necessary since we know we're in a given language based on the button at the top of the page; it was used for the Browse categories tab because that is a more complicated field arrangement (1 label with 3 potential associated translations) and the headings helped visually distinguish the two parts)
- Right-align the English language labels (similar to the General tab, for example, but we don't need the same translated field hint text part of the input; the input is fine as it is now)
- Make sure the English language labels are bold (default formatting for labels I think); maybe they are already, it's just not quite clear to me from the screenshot

mejackreed:
Thanks for the feedback @ggeisler . I think I've made the specified changes except for the align-right and font-weight changes. The reason I didn't make these, was for consistency with other form elements here. This has been overriden downstream in some of our other applications.

Screen Shot 2021-01-27 at 10 56 43 AM

ggeisler:
👍 on the align-right and font-weight as we discussed (I didn't realize they are styled differently in Spotlight than the Exhibits example I was looking at for reference).

PR #2620, opened 2021-01-28
Update the i18n for the Spotlight::Search model name to …

### LINKED ISSUE #2615: Update flash message text of Curation > Browse categories page
This is not directly related to the recent browse groups feature, but I noticed it when testing that feature.

When you select Save on the Curation > Browse, Browse categories page/tab, the flash message uses the term "Searches":

Screen Shot 2021-01-26 at 5 56 33 PM

---

I know that technically these are _saved searches_ but once the search is saved we try to get the user to think about these as _browse categories_ and that's the term we use elsewhere on this page. So ideally the flash message should use matching terminology and use "Browse categories" instead of "Searches":

> Browse categories were successfully updated.

That change would also match the terminology used for the browse groups page/tab flash message, as noted in #2613.

jkeck:
…"Browse category"

Closes #2615 

Since I'm overriding how the model name is translated, this gets update in a few other places

Screen Shot 2021-01-27 at 5 31 30 PM
Screen Shot 2021-01-27 at 5 31 43 PM
Screen Shot 2021-01-27 at 5 32 18 PM
Screen Shot 2021-01-27 at 5 32 25 PM

### DIFF
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -39,6 +39,7 @@ ignore_unused:
   - activerecord.attributes.spotlight/masthead.display # app/views/spotlight/appearances/edit.html.erb
   - activerecord.attributes.spotlight/custom_field.is_multiple # app/views/spotlight/custom_fields/_form.html.erb
   - activerecord.attributes.spotlight/custom_search_field.field # app/views/spotlight/custom_search_fields/_form.html.erb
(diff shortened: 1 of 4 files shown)


ggeisler:
I guess this is already merged, but in the dialog ("Are you sure you want to delete this Browse category?" the word browse should not be capitalized.

jkeck:
Yeah, I'm not 100% sure why that was the case (I'm guessing `Search` was also incorrectly capitalized)

PR #2623, opened 2021-01-28
Handle potential Arrays as heading values in autocomplete title response

mejackreed:
Fixes https://github.com/sul-dlss/dlme/issues/1160

Note: I've looked into this from the other side of things and really wasn't able to make headway on it.

### DIFF
--- a/app/controllers/concerns/spotlight/base.rb
+++ b/app/controllers/concerns/spotlight/base.rb
@@ -32,7 +32,7 @@ module Spotlight
     def autocomplete_json_response_for_document(doc)
       {
         id: doc.id,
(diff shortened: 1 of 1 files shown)


cbeer:
Hm. Do you know why `heading` is an array? Is it the DLME multilingual stuff, or??

mejackreed:
Yes it seems to be the multilingual stuff. And I've found a different behavior by default in that lambda, and when I insert a debugger in there everything works fine. Its really odd.

mejackreed:
Closing this in favor upstream work in DLME. https://github.com/sul-dlss/dlme/pull/1162

PR #2624, opened 2021-01-28
Updates to resolve issues with import/export exhibit lifecycle

mejackreed:
Fixes https://github.com/sul-dlss/dlme/issues/1158

This pull request does two things:

1. Move all of the iiif crop image params to a single helper so that if we change/modify them they won't be different. A previous PR that added a param missed one of these.
2. Unset the iiif_tilesource for a Spotlight::FeaturedImage when importing an exhibit. The rationale here is that this field can be regenerated if unset and could contain some incorrect data coming from a remote image being reimported.

### DIFF
--- a/app/services/spotlight/exhibit_import_export_service.rb
+++ b/app/services/spotlight/exhibit_import_export_service.rb
@@ -216,6 +216,8 @@ module Spotlight
                                                      filename: file[:filename],
                                                      content_type: file[:content_type]
       end
+      # Unset the iiif_tilesource field as the new image should be different
+      image.iiif_tilesource = nil
       image.save!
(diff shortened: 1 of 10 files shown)


app/services/spotlight/exhibit_import_export_service.rb
```diff
@@ -216,6 +216,8 @@ def deserialize_featured_image(obj, method, data)
                                                      filename: file[:filename],
                                                      content_type: file[:content_type]
       end
+      # Unset the iiif_tilesource field
```
mejackreed, line 219: I'm not sure if this is the right place to do this. Any other options?

PR #2625, opened 2021-01-28
Refactor indexing + status reporting to support better parallelization

cbeer:
This PR does 3 major things (and several minor fixes, hardening, and throwing away unused code discovered along the way):

## Replace ReindexingLogEntries with more generic JobTracker + Event reporting && use activejob-status to provide real-time progress information (🤞 )

The existing job tracking was somewhat fragile. ActiveJob::Status helps by taking the real-time progress information and storing it in the Rails cache instead. We wrap that with a `Spotlight::JobTracker` to persist (and expose) job information longer than the cache lifetime (so, 🤞 , we get real-time information when the jobs are in flight and stored information after they are completed.

Events are not currently displayed to the end user from within Spotlight itself, but are available for downstream implementations to use. Ideally, we'll add some kind of administrator "log" view to help debug indexing issues.

## Add a Spotlight::ReindexExhibitJob

For exhibits with lots of resources (e.g. DLME), the single-threaded exhibit indexer is not performant. This new job takes the exhibit reindex request and shards it by paginating the jobs list into N chunks or M documents (by default, the number of chunks grows with the number of resources available; 10 resources => 3 chunks, 1000 resources => 8 chunks, etc).

## Implement an ETL-like pattern for indexing documents

The SolrDocumentBuilder "pattern" worked, but led to some weird interactions between the reindex job, the resource, and the document builder (ping-ponging back and forth, hard to monitor, and hard to debug when things broke).

Instead, this PR replaces the pattern (without deprecation or migration path -- sorry, they're just two totally different paradigms.. although converting existing code to use the new system is fairly painless; at a cursory glance of available code on github, Stanford's the only ones who made significant use of the older pattern, and the other projects are still pinned to 2.x version of spotlight, so 🤞  we can get away with this with the 3.0 bump?)  to use an ETL-like approach where each Spotlight resource defines an ETL pipeline that can index the documents. The ETL pipeline is defined in 3 main stages:

- `sources`, which receive the ETL context (providing access to the `Spotlight::Resource` instance, among other things) and return an enumerable (meaning a single Spotlight::Resource may produce 1 or more solr documents) of sources
- `transform`, which receives the current data output and the context (again, providing access to the resource, but also the current source) and returns a new version of the data.
- `load`, out-of-the-box, currently just a SolrLoader that buffers documents into batches, does error handling, and fires a commit request after the indexing is complete.

More information is available in the YARD docs for the `Spotlight::Etl::Pipeline` and other components.

### Moving to the new system

A quick-and-dirty conversion to the new system might look something like:

```diff
 class DlmeJson < Spotlight::Resource
-  self.document_builder_class = DlmeJsonResourceBuilder

+  def self.indexing_pipeline
+    @indexing_pipeline ||= super.dup.tap do |pipeline|
+      # if, say, you wanted to feed the transform with multiple source documents (here, by calling the `#iiif_manifest` method on the DlmeJson instance); previously, the #to_solr method of the document builder would have done this extraction
+      # pipeline.sources = [Spotlight::Etl::Sources::SourceMethodSource(:iiif_manifests)]
+      pipeline.transforms = [
+        ->(data, p) { data.merge(DlmeJsonResourceBuilder.new(p.context.resoure).to_solr) }
+      ]
+    end
+  end
```

```diff
 # transforms a DLME JSON intermediate representation into solr documents
-class DlmeJsonResourceBuilder < Spotlight::SolrDocumentBuilder
+class DlmeJsonResourceBuilder
 
+  attr_reader :resource
+
+  def initialize(resource)
+    @resource = resource
+  end
+
   def to_solr
     source = resource.json
+  # do whatever munging is necessary. this logic might be unchanged from what you had previously... except that it only deals with a single ETL source at a time instead of doing the source iteration internally
```


mejackreed (approved):
Reviewed in a pairing session

PR #2632, opened 2021-02-01
Setup a destroy method so that turbolinks navigation still renders th…

mejackreed:
…e sliders appropriately

Fixes an issue that when navigating between pages, tns needs to be reset so that it can render appropriately.

### DIFF
--- a/app/assets/javascripts/spotlight/user/browse_group_categories.js
+++ b/app/assets/javascripts/spotlight/user/browse_group_categories.js
@@ -28,6 +28,15 @@
       });
     }
 
+    // Destroy the slider instance, as tns will change the dom elements, causing some issues with turbolinks
+    function setupDestroy() {
+      document.addEventListener('turbolinks:before-cache', function() {
+        if (slider && slider.destroy) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/spotlight/user/browse_group_categories.js
```diff
@@ -25,6 +25,17 @@
       });
     }
 
+    // Destroy the slider instance, as tns will change the dom elements, causing some issues with turbolinks
+    function setupDestroy() {
+      if (Turbolinks.supported) {
```
cbeer, line 30: Is it.. possible Turbolinks will be null? Would you be safe just always adding the listener?

mejackreed: Thanks for the review/suggestions. I've removed the check, and changed the event. That seems to be working just fine.

cbeer: 🎉

cbeer (approved):

PR #2655, opened 2021-02-08
Change the per_page setting so we get the default_per_page set …

jkeck:
…explicitly in the configuration instead re-ordering per_page

Closes sul-dlss/exhibits#1941

In the example below the exhibit has set a default per-page of 50, and the user has not selected a specific per-page option (so the default is being acted on).

## Before
per-page-before

## After
per-page-after

### DIFF
--- a/app/models/spotlight/blacklight_configuration.rb
+++ b/app/models/spotlight/blacklight_configuration.rb
@@ -84,6 +84,8 @@ module Spotlight
 
         config.default_solr_params = config.default_solr_params.merge(default_solr_params)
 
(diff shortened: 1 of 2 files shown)


cbeer (approved):
I assume... we just didn't have default_per_page 7 years ago?

jkeck:
I've been trying to figure that out, but yeah, I  _think_ we just used to go w/ the first one back then, otherwise this bit of code was a bit silly (and I'm surprised nobody has noticed / said anything in 7 years)

jkeck:
Build os broken by github/view_component#642

PR #2656, opened 2021-02-11
Fix accessibility issue where a link with the same name goes to different locations

elrayle:
_NOTE: This was identified as an accessibility error by SiteImprove_

### Description

On item show pages (e.g. http://localhost:3000/test-exhibit/catalog/3-9), the "Home" link in the exhibit navigation menu and the "Home" link in the breadcrumbs are going to two different locations. 

| Menu | URL |
| -- | -- |
| "Home" navigation menu | https://localhost/test-exhibit |
| "Home" breadcrumb | https://localhost/test-exhibit?q= |

### Solution

Change the breadcrumb "Home" link to match the one in the navigation menu.

### Potential similar fix

Line 131 of the same file uses the same `q: ''` parameter which causes the URL to end with `?q=`.  I haven't identified a specific error associated with this, but it might make sense to make the same change on that line.

### Request to maintainers

Please backport this to the 2.x series.

### DIFF
--- a/spec/controllers/spotlight/catalog_controller_spec.rb
+++ b/spec/controllers/spotlight/catalog_controller_spec.rb
@@ -29,7 +29,7 @@ describe Spotlight::CatalogController, type: :controller do
       let(:search) { FactoryBot.create(:search, exhibit: exhibit) }
 
       it 'shows the item' do
-        expect(controller).to receive(:add_breadcrumb).with('Home', exhibit_path(exhibit, q: ''))
+        expect(controller).to receive(:add_breadcrumb).with('Home', exhibit_path(exhibit))
         expect(controller).to receive(:add_breadcrumb).with('L&#39;AMERIQUE', exhibit_solr_document_path(exhibit, document))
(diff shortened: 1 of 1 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -131,7 +131,7 @@ def attach_breadcrumbs
         add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: @exhibit.title), exhibit_root_path(@exhibit, q: '')
       else
         # When not on the admin page, get the translated value for the "Home" breadcrumb
-        add_breadcrumb t(:'spotlight.curation.nav.home', title: @exhibit.title), exhibit_root_path(@exhibit, q: '')
```
cbeer, line 134: The comment above this code suggested it was a necessary workaround, at least at some point in time. We should check if it's still an issue, or if it's fixed, also fix this on line 131, remove the comment, and pin the dependencies as appropriate.

elrayle:
Replaced by PR #2658.

PR #2658, opened 2021-02-15
Accessibility: Link text used for multiple different destinations - Home Links on Exhibit

elrayle:
**_NOTE: This replaces PR #2656.  I had problems with the fork associated with that PR because it was out of date.  I had to create a new fork and new PR.  Comments are copied forward from the original PR._**

----
_NOTE: This was identified as an accessibility error by SiteImprove: Link text used for multiple different destinations_

### Description

On item show pages (e.g. http://localhost:3000/test-exhibit/catalog/3-9), the "Home" link in the exhibit navigation menu and the "Home" link in the breadcrumbs are going to two different locations. 

| Menu | URL |
| -- | -- |
| "Home" navigation menu | https://localhost/test-exhibit |
| "Home" breadcrumb | https://localhost/test-exhibit?q= |

### Solution

Change the breadcrumb "Home" link to match the one in the navigation menu.

### Potential similar fix

Fixed a similar issue on line 131 of the same file that also used `q: ''` parameter which causes the URL to end with `?q=`.

### Request to maintainers

Please backport this to the 2.x series.

### DIFF
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -126,15 +126,11 @@ module Spotlight
     protected
 
     def attach_breadcrumbs
-      # The "q: ''" is necessary so that the breadcrumb builder recognizes that a path like this:
-      # /exhibits/1?f%5Bgenre_sim%5D%5B%5D=map&q= is not the same as /exhibits/1
-      # Otherwise the exhibit breadcrumb won't be a link.
-      # see http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-current_page-3F
       if view_context.current_page?({ action: :admin })
-        add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: @exhibit.title), exhibit_root_path(@exhibit, q: '')
+        add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: @exhibit.title), exhibit_root_path(@exhibit)
       else
         # When not on the admin page, get the translated value for the "Home" breadcrumb
-        add_breadcrumb t(:'spotlight.curation.nav.home', title: @exhibit.title), exhibit_root_path(@exhibit, q: '')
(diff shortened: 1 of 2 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -126,15 +126,11 @@ def manifest
     protected
 
     def attach_breadcrumbs
-      # The "q: ''" is necessary so that the breadcrumb builder recognizes that a path like this:
-      # /exhibits/1?f%5Bgenre_sim%5D%5B%5D=map&q= is not the same as /exhibits/1
-      # Otherwise the exhibit breadcrumb won't be a link.
-      # see http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-current_page-3F
       if view_context.current_page?({ action: :admin })
-        add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: @exhibit.title), exhibit_root_path(@exhibit, q: '')
+        add_breadcrumb t(:'spotlight.exhibits.breadcrumb', title: @exhibit.title), exhibit_root_path(@exhibit)
```
elrayle, line 130: Comment by @cbeer from original PR #2656.
> The comment above this code suggested it was a necessary workaround, at least at some point in time. We should check if it's still an issue, or if it's fixed, also fix this on line 131, remove the comment, and pin the dependencies as appropriate.

elrayle: The code and comment were added when Rails 4 was required.  I was not able to reproduce the issue described in the comment since Rails 4 is no longer allowed and other dependencies also prevent using Rails 4.  I tested with Rails 5.2.  Both URLs were generated correctly.

RE: Pin - current Spotlight pins to `'rails', '>= 5.2', '< 6.2'`.  I believe this pin is sufficient based on testing with Rails 5.2.

elrayle:
The only test failure is the same one that is failing on a number of open PRs.

elrayle:
The failing test will be fixed when PR #2662 is merged.

elrayle:
@jkeck This would pass if it was the jobs are restarted.  I don't have the ability to do that.

PR #2592, opened 2021-01-12
generated initializer list of sir treavor widgets should match configured list

elrayle:
Updates list of sir treavor widgets in the generator initializer to match those defined in [lib/spotlight/engine.rb](https://github.com/projectblacklight/spotlight/blob/master/lib/spotlight/engine.rb#L270-L274)

PR #2667, opened 2021-02-23
generated initializer list of sir treavor widgets should match configured list

elrayle:
Updates list of sir treavor widgets in the generator initializer to match those defined in [lib/spotlight/engine.rb](https://github.com/projectblacklight/spotlight/blob/master/lib/spotlight/engine.rb#L270-L274)


elrayle:
This was originally submitted in PR #2592.  The fork it was based on was deleted, so I had to close that and add a new PR.

@mejackreed You reviewed PR #2592.  This is the same change after rebasing.

PR #2668, opened 2021-02-24
Add bulk action functionality for updating item visibility

### LINKED ISSUE #2644: Modify curator view of search results page to offer new bulk actions button
Part of #2643.

This ticket just sets up the larger tickets for adding the new bulk actions. There are two main pieces to this one:

- Adjusting the layout of the search results page to support two curator-only buttons
- Adding a new “Bulk actions” button

Screen Shot 2021-02-03 at 3 47 04 PM

---

 - [Blacklight ticket](https://github.com/projectblacklight/blacklight/issues/2429) - Add a new row (`<div class="row"><div class="col-md-12">`) directly after the row used for flash messages. (The “Save this search” button already wraps awkwardly at the `LG` breakpoint and below so a new row would fix that plus support the new button. This entire row could then be rendered only for curators, I believe.)
- [x] Add a new button `Bulk actions` next to the existing “Save this search” button
- [x] The new button should use the same styling as the “Save this search” button
 - [Blacklight ticket](https://github.com/projectblacklight/blacklight/issues/2429) - The two curator-only buttons should be right-aligned in the row
 - [Blacklight ticket](https://github.com/projectblacklight/blacklight/issues/2429) - The two curator-only buttons should use similar horizontal spacing between as the search result controls in the row below
- [x] The “Bulk actions” button is a select that should eventually include the following options (although we should only add them as we add support for the action):
- `Change item visibility`
- `Add item tags`
- `Remove item tags`
- `Update custom metadata field`

---

Screen Shot 2021-02-03 at 3 47 39 PM

### LINKED ISSUE #2645: Add bulk action: Change item visibility
Part of #2643. Depends on #2644.

Item visibility can be either `Public` or `Private`. This action should enable the curator to set the visibility of all items in the current result set to either `Public` or `Private`. This means that if the results set contains some items that are currently `Public` and some items that are currently `Private`, only some items in the result set are actually updated.

As with all bulk actions, the form the curator uses to update the items should be presented in a modal. The curator can use a “Cancel” button (or close icon, or click on the scrim outside the modal) to close the modal and return to the results page with no items updated. Or the curator can select the submit button, in which case the update should be performed, the modal closed, and the results page updated (using the current URL parameters).

## Bulk actions button option

vis-button

---

- Option label is `Change item visibility`

## Modal window

vis-modal

---

The modal positioning and styling details can be based on the current Save this search modal.

- [x] A: Modal heading: `Change item visibility`
- [x] B: Intro text: `There are N items in the current search result set. This change will be applied to all items.` (where N is the number of items in the result set)
- [x] Ideally, pluralize the `are N items` part of the text above so it works when there is only one item in the result set (`is one item`), as well as when there are multiple items.
- [x] C: Radio button group label `Item visibility`
- [x] D: Radio button labels `Public` and `Private`. Ideally, if all items in the result set have the same visibility value, the default value of the button is that value (e.g., if all items are currently `public`, the default button is `Public`; that way if the user selects “Change” unintentionally or without thinking, no change is made). If we can’t do that, or if the set has items of both values, the default radio button should be `Private`.
- [x] E: `Cancel` closes the modal without making any update. `Change` closes the modal, updates all items in the current result set as necessary to set their visibility value to the select radio option.

## Flash message
If the user does choose the Change action, display a flash message. Below are two options.

### Option A - Minimal approach

flash-basic

---

If we don’t have access to the number of items that actually needed to be updated, and we can’t rollback the change, then the message can be pretty basic:

- [x] `Visibility of N items updated.` where N is the total number of items in the result set.
- [x] Pluralize "item" so it makes grammatical sense when there is either a single item or multiple items in the result set

### Option B - Preferred approach

We can be more informative and helpful to the user (and presumably eliminate some questions for the service team) if we can probably a more detailed message and offer to rollback the changes. So this is the preferred option if it is technically feasible.

flash-adv

---

- [ ] A: `The visibility setting was updated for N items in the result set.` where N is the actual number of items that needed to be updated, rather than the total number of items in the result set (i.e., some items might have already have the desired value and didn’t require the update).
- [ ] Pluralize "item" so it makes grammatical sense when either a single item or multiple items were updated
- [ ]  B: `Undo this action.` Assuming we can rollback the changes, this action could help the user who didn’t quite understand how the feature worked and didn’t expect the update to affect so many items.

If we can’t provide the more informative update message but can do the rollback, that should be part of Option A above.
    
 I’m not certain what the potential failures would look like so am not providing any error versions of the flash message, but I can help with that later if necessary.

mejackreed:
Fixes #2645 and fixes #2644
Screen Shot 2021-02-24 at 7 56 54 AM
Screen Shot 2021-02-24 at 7 57 43 AM

Kapture 2021-02-24 at 07 56 31

Todos:
 - [x] Use cursorMark to paginate solr response in job
 - [x] commit optimization?
 - [x] job tracking? (maybe can be split out?)

### DIFF
--- a/app/assets/stylesheets/spotlight/_catalog.scss
+++ b/app/assets/stylesheets/spotlight/_catalog.scss
@@ -117,7 +117,7 @@ form.edit_solr_document {
   @extend .clearfix;
 }
 
-#save-this-search {
+#save-this-search, .bulk-actions-dropdown {
   float:left;
   margin-right: $spacer * 0.1;
 }
--- a/app/controllers/concerns/spotlight/catalog.rb
+++ b/app/controllers/concerns/spotlight/catalog.rb
@@ -31,5 +31,10 @@ module Spotlight
       (current_exhibit && can?(:curate, current_exhibit)) &&
         !(params[:controller] == 'spotlight/catalog' && params[:action] == 'admin')
     end
+
+    def render_bulk_actions?
+      (current_exhibit && can?(:curate, current_exhibit)) &&
+        !(params[:controller] == 'spotlight/catalog' && params[:action] == 'admin')
+    end
   end
 end
--- a/app/controllers/spotlight/bulk_actions_controller.rb
+++ b/app/controllers/spotlight/bulk_actions_controller.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Spotlight
+  ##
+  # Controller enabling bulk functionality for search results
+  class BulkActionsController < Spotlight::CatalogController
+    before_action :authenticate_user!
+    before_action :check_authorization
+
+    # rubocop:disable Metrics/MethodLength
+    def visibility
+      solr_params = nil
+      # Get the total number of results
+      (response,) = search_service.search_results do |builder|
+        builder.merge(fl: 'id', rows: 1)
+        solr_params = builder.to_h
+      end
+
+      Spotlight::ChangeVisibilityJob.perform_later(
+        solr_params: solr_params,
+        exhibit: current_exhibit,
+        visibility: visibility_param,
+        user: current_user
+      )
+
+      redirect_back fallback_location: fallback_url,
+                    notice: t(:'spotlight.bulk_actions.change_visibility.changed', count: response.total)
+    end
+    # rubocop:enable Metrics/MethodLength
+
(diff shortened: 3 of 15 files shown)


jkeck (approved):
LGTM 👍

PR #2677, opened 2021-02-25
Adds bulk action to add tags

### LINKED ISSUE #2647: Add bulk action: Add item tags
Part of #2643. Depends on #2644.

An exhibit item can have zero or more tags. This bulk action should enable the curator to add one or more tags to all of the items in the current result set at once. (There is a separate action for removing tags from items, to prevent the form from being too complicated and making it more challenging for the user to understand what is going on.)

As with all bulk actions, the form the curator uses to update the items should be presented in a modal. The curator can use a “Cancel” button (or close icon, or click on the scrim outside the modal) to close the modal and return to the results page with no items updated. Or the curator can select the submit button, in which case the update should be performed, the modal closed, and the results page updated (using the current URL parameters).

## Bulk actions button option

add-button-option

---

- [ ] Option label is `Add tags`

## Modal window

add-tags-modal-initial

---

The modal positioning and styling details can be based on the current Save this search modal.

- [x] A: Modal heading: `Add tags`
- [x] B: Intro text: `There are N items in the current search result set. Tags entered below will be added to all items in the current result set. This won’t affect any tags already assigned to items.` (where N is the number of items in the result set)
- [x] Ideally, pluralize the `are N items` part of the text above so it works when there is only one item in the result set (`is one item`), as well as when there are multiple items.
- [x] C: Input label `Tags`
- [x] D: Input. The design assumes this works exactly the same as the similar input on the item show page in edit mode (see example of tags entered below). We could look at other approaches (this is not necessarily the most intuitive for a first encounter, in my opinion) but if this is technically feasible it has the advantage of consistency with the item edit page, enables adding both a new tag and selecting an existing one, and once you know how to use it, it’s pretty straightforward.
- [x] E: `Cancel` closes the modal without making any update. `Add` closes the modal and updates all items in the current result set with the tag or tags specified (some items could already have the added tags associated with them).

add-tags-enter-tags

---

## Flash message

If the user does choose the Add action, display a flash message. Below are two options.

### Option A - Minimal approach

flash-msg

If we don’t have access to the number of items that were actually updated (since some items in the result set might already have all of the tags being added), and we can’t rollback the change, then the message can be pretty basic:

- [x] `Tags added to N items.` Where N is the number of items in the result set.
- [x] Pluralize "Tag" so it makes grammatical sense when both a single tag is added and when more than one tag is added
- [x] Pluralize "item" so it makes grammatical sense when there is either a single item or multiple items in the result set

### Option B - Preferred approach
We can be more informative and helpful to the user (and presumably eliminate some questions for the service team) if we can provide a more detailed message and offer to rollback the changes. So this is the preferred option if it is technically feasible.

flash-adv

---

- [ ] `Tags added to N items in the result set. Undo this action.` Where N is the number of items that were actually updated with one or more tags.
- [ ] Pluralize "Tag" so it makes grammatical sense when both a single tag is added and when more than one tag is added
- [ ] Pluralize "item" so it makes grammatical sense when both a single item or multiple items are updated

If we can’t provide the more informative update message but can do the rollback, that should be part of Option A above.

I’m not certain what the potential failures would look like so am not providing any error versions of the flash message, but I can help with that later if necessary.

## Flash message - alternative approach
A potential alternative the the flash message options above might be to use a confirmation dialog instead. It feels a bit awkward to present a dialog window above a modal window, but this approach could be simpler than the flash message approach:

1. When the user selects the "Add" action to add a tag(s), we either perform the update or, more ideally, execute a query to determine how many items would be affected by the update without actually making the update (I'm not sure if the latter is possible).
2. Display a confirmation dialog with the count of items affected:

Screen Shot 2021-02-11 at 5 25 16 PM

3. Cancel - Update is not made (or if we've already made it, rollback the update), dismiss the dialog, and leave the modal in place.
4. OK - Update is executed (if it hasn't been already), dismiss the dialog, cancel the modal, and refresh the results page. No flash message is necessary because we've already informed the user of how many items have been affected by the update.

mejackreed:
Fixes #2647 

Screen Shot 2021-02-25 at 4 41 12 PM
Screen Shot 2021-02-25 at 4 41 23 PM

Kapture 2021-02-25 at 16 40 48

### DIFF
--- a/app/assets/javascripts/spotlight/admin/catalog_edit.js
+++ b/app/assets/javascripts/spotlight/admin/catalog_edit.js
@@ -1,50 +1,16 @@
 Spotlight.onLoad(function() {
-  if($('#solr_document_exhibit_tag_list').length > 0) {
-    // By default tags input binds on page ready to [data-role=tagsinput],
-    // however, that doesn't work with Turbolinks. So we init manually:
-    $('#solr_document_exhibit_tag_list').tagsinput();
-
-    var tags = new Bloodhound({
-      datumTokenizer: function(d) { return Bloodhound.tokenizers.whitespace(d.name); },
-      queryTokenizer: Bloodhound.tokenizers.whitespace,
-      limit: 100,
-      prefetch: {
-        url: $('#solr_document_exhibit_tag_list').data('autocomplete_url'),
-        ttl: 1,
-        filter: function(list) {
-          return $.map(list, function(tag) { return { name: tag }; });
-        }
+  $(".visiblity_toggle").blCheckboxSubmit({
+    //css_class is added to elements added, plus used for id base
+    cssClass: "toggle_visibility",
+    //success is called at the end of the ajax success callback
+    success: function (isPublic){
+      // We store the selector of the label to toggle in a data attribute in the form
+      var docTarget = $($(this).data("label-toggle-target"));
+      if ( isPublic ) {
+        docTarget.removeClass("blacklight-private");
+      } else {
+        docTarget.addClass("blacklight-private");
       }
-    });
-
-    tags.initialize();
-
-    $('#solr_document_exhibit_tag_list').tagsinput('input').typeahead({highlight: true, hint: false}, {
-      name: 'tags',
-      displayKey: 'name',
-      source: tags.ttAdapter()
-    }).bind('typeahead:selected', $.proxy(function (obj, datum) {
-      $('#solr_document_exhibit_tag_list').tagsinput('add', datum.name);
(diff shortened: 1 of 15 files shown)


ggeisler:
Thanks for the screenshots and gif/video. Submit button should be labelled "Add" rather than "Change".

mejackreed:
Nice catch @ggeisler . I've updated it

Screen Shot 2021-02-25 at 5 03 58 PM

jkeck (approved):
LGTM 👍  (I paired on some parts of this so will give the opportunity for other 👀)

PR #2678, opened 2021-02-26
Add the initial bulk updates admin pages

jkeck:
Part of sul-dlss/exhibits#1995
Closes sul-dlss/exhibits#1996
Closes sul-dlss/exhibits#1997
Closes sul-dlss/exhibits#1998
Closes sul-dlss/exhibits#1999

This will be a bit slow with a large amount of records.  I'm not sure there is much we could do about that unless we put this in the background and email the resulting CSV / CSV download link (which would be a pretty big departure from this)

Screen Shot 2021-02-25 at 4 36 00 PM

Screen Shot 2021-03-03 at 1 11 58 PM

### DIFF
--- a/app/views/spotlight/shared/_curation_sidebar.html.erb
+++ b/app/views/spotlight/shared/_curation_sidebar.html.erb
@@ -17,7 +17,7 @@
   <% if (can? :manage, current_exhibit.translations.first_or_initialize) && current_exhibit.languages.any? %>
     <%= nav_link t(:'spotlight.curation.sidebar.translations'), spotlight.edit_exhibit_translations_path(current_exhibit), 'data-no-turbolink' => true %>
   <% end %>
-  <% if can? :curate, current_exhibit %>
+  <% if can? :bulk_update, current_exhibit %>
     <%= nav_link t(:'spotlight.curation.sidebar.bulk_updates'), spotlight.edit_exhibit_bulk_updates_path(current_exhibit), 'data-no-turbolink' => true %>
   <% end %>
 </ul>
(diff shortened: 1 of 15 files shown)


jkeck:
@ggeisler the CSV Guidelines in my screenshot above don't quite match what you have in your design.  How faithfully should I be matching spacing/indentation/etc?

ggeisler:
@jkeck This looks fine. I might make some minor spacing tweaks at some point but I don't see anything that should hold up you going forward with this now.

app/views/spotlight/shared/_curation_sidebar.html.erb
```diff
@@ -17,4 +17,7 @@
   <% if (can? :manage, current_exhibit.translations.first_or_initialize) && current_exhibit.languages.any? %>
     <%= nav_link t(:'spotlight.curation.sidebar.translations'), spotlight.edit_exhibit_translations_path(current_exhibit), 'data-no-turbolink' => true %>
   <% end %>
+  <% if can? :curate, current_exhibit %>
```
cbeer, line 20: I wonder if we want to use a different ability to let downstream consumers limit who can bulk update separate from other curator actions? (And, we could limit it to superadmins in exhibits until we've got all the pieces in place?)

jkeck: That was certainly brought up in our initial spike, but given that bulk actions only check if the user could curate the exhibit we followed the same pattern.  I'm happy to go down that route though.  If we do, a few questions come to mind.

What should that ability be?  Should we also be checking the same/similar ability for bulk actions?

jkeck: Went with `bulk_updates` and only am checking it for the bulk update admin UI.

Issue #2676, opened 2021-02-25
Split out SolrDocumentSidecar tags to not have the Exhibit as a tagger


mejackreed:

cbeer:
This is going to be a little tricky because we're using the tagger scope to get all the tags that are in an exhibit (used for autocomplete, on the tags admin page.). IT might be possible to use some distasteful SQL to get the result we want, though:

```
ActsAsTaggableOn::Tag.joins(:taggings).joins('INNER JOIN spotlight_solr_document_sidecars ON taggings.taggable_type="Spotlight::SolrDocumentSidecar" AND taggings.taggable_id=spotlight_solr_document_sidecars.id').where('spotlight_solr_docume
t_sidecars.exhibit_id = ?', 15).select(:name).distinct.pluck(:name)
```

PR #2683, opened 2021-03-01
Spike on processing CSV uploads

cbeer:
TODO:

- [ ] can we avoid querying solr after each and every document? At the very least, we should batch queries and only retrieve the document ids instead of the full record
- [ ] is it worth trying to optimize only to records we know are changing?
- [ ] ugh, tags. #2676 , but 🤷‍♂️ 
- [ ] move column names into a configuration


jkeck:
Cherry-picked to #2700

PR #2684, opened 2021-03-02
Do not render the bulk_actions UI on Browse groups

mejackreed:
I'm not sure what the intention should be, my guess would be to not show it.. but lets disable it until we determine how to get the search injected.

### DIFF
--- a/app/controllers/spotlight/browse_controller.rb
+++ b/app/controllers/spotlight/browse_controller.rb
@@ -134,5 +134,9 @@ module Spotlight
     def render_save_this_search?
       false
     end
+
+    def render_bulk_actions?
+      false
+    end
   end
 end


cbeer (approved):

ggeisler:
Thanks for the quick fix on this! 

Longer-term I do think it would be ideal if we were able to support bulk actions on a browse category results page. A browse category could represent the exact search result set an curator wants to act on (e.g., tag everything in a given browse category with a specific tag) so forcing them to recreate the search result manually feels less than ideal.

If supporting browse category results turns out to be challenging for a technical reason, I do think we could live without it (we'd have to add some documentation to make the limitation clear to curators), but it seems to me it is worth spending some time looking into to see if we can support it.

Issue #2685, opened 2021-03-02
Add support for bulk actions on browse category pages


mejackreed:

ggeisler:
For context: I do think it would be ideal if we were able to support bulk actions on a browse category results page. A browse category could represent the exact search result set an curator wants to act on (e.g., tag everything in a given browse category with a specific tag) so forcing them to recreate the search result manually feels less than ideal.

If supporting browse category results turns out to be challenging for a technical reason, I do think we could live without it (we'd have to add some documentation to make the limitation clear to curators), but it seems to me it is worth spending some time looking into to see if we can support it.

corylown:
This issue is currently stale and is being closed. If you would like to discuss this work further, please reopen or create a new issue. Thank you!

PR #2688, opened 2021-03-03
Use a shared 'tool' for all curator actions and use blacklight config for tools + actions

cbeer:
Screen Shot 2021-03-03 at 10 11 02

### DIFF
--- a/app/controllers/concerns/spotlight/catalog.rb
+++ b/app/controllers/concerns/spotlight/catalog.rb
@@ -27,12 +27,7 @@ module Spotlight
                                         }
     end
 
-    def render_save_this_search?
-      (current_exhibit && can?(:curate, current_exhibit)) &&
-        !(params[:controller] == 'spotlight/catalog' && params[:action] == 'admin')
-    end
-
-    def render_bulk_actions?
+    def render_curator_actions?
       (current_exhibit && can?(:curate, current_exhibit)) &&
         !(params[:controller] == 'spotlight/catalog' && params[:action] == 'admin')
     end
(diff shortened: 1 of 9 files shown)


mejackreed:
I think we do want this, but the ticket was created upstream in Blacklight

ggeisler:
Yep, that is definitely the end result we want, but whether it makes sense to do it here or in Blacklight I'll leave up to you all.

One thing I see from the screenshot @cbeer is that the existing buttons on the bottom row seem to have lost their horizontal spacing (between Sort and Per page and between Per page and the result view buttons). There wasn't a lot of horizontal space before but there was some, and from the screenshot it appears there is none.

cbeer:
thanks; I'm not sure where that spacing magically comes from, but i've added some back.

app/controllers/concerns/spotlight/catalog.rb
```diff
@@ -27,12 +27,7 @@ def add_facet_visibility_field
                                         }
     end
 
-    def render_save_this_search?
-      (current_exhibit && can?(:curate, current_exhibit)) &&
-        !(params[:controller] == 'spotlight/catalog' && params[:action] == 'admin')
-    end
-
-    def render_bulk_actions?
+    def render_curator_actions?
```
mejackreed, line 30: I'm good with this, but @ggeisler is there a reason on when we would want to show "Save search" and "Bulk actions" separately? Does this make #2685 impossible?

mejackreed (commented):
This looks great to me, just had one question about us enabling bulk actions in the "Browse" context.

app/controllers/concerns/spotlight/catalog.rb
cbeer: We should be able to put `if`/`unless` conditions on the individual tools.

mejackreed: WFM

mejackreed (approved):

PR #2690, opened 2021-03-03
Rename extension_whitelist to extension_allowlist

jkeck:
extension_whitelist has been deprecated upstream:
https://github.com/carrierwaveuploader/carrierwave/blob/13330a70323cde83b319520f5c1874227de79cdf/lib/carrierwave/uploader/extension_whitelist.rb#L35-L36


cbeer:
Should we update the dependency specification to `s.add_dependency 'carrierwave', '~> 2.2'`, or make this backwards compatible with older versions?

jkeck:
Good catch, I think since this is coming out in spotlight 3 (technically) making sure folks are on an up-to-date version of carrierwave seems reasonable (plus we were giving them a different configuration path if they were trying to update the extensions, 🤞)

PR #2691, opened 2021-03-04
Update text and styling of bulk actions modals

ggeisler:
There are kind of a lot of text changes but it is pretty straightforward. I just customized the singular case strings and the plural case strings bit more, and updated a few words here and there. 

I also added a sentence to the confirmation dialog, but used the same wording for all confirmation cases to keep it simple.

Also made a couple of very small styling updates. I created a new SCSS partial for the modals because I can foresee downstream implementers wanting to do a bit more styling customization on them.

### Examples with one result item
Screen Shot 2021-03-03 at 5 59 50 PM

Screen Shot 2021-03-03 at 5 59 35 PM

Screen Shot 2021-03-03 at 6 00 07 PM

### Examples with more than one result item
Screen Shot 2021-03-03 at 6 00 57 PM
Screen Shot 2021-03-03 at 6 00 46 PM
Screen Shot 2021-03-03 at 6 00 30 PM

### Confirmation dialog update
Screen Shot 2021-03-03 at 6 01 11 PM

### DIFF
--- a/app/assets/stylesheets/spotlight/_modals.scss
+++ b/app/assets/stylesheets/spotlight/_modals.scss
@@ -1,0 +1,3 @@
+.modal-body .form-check-label {
+  margin-bottom: 0.5rem;
+}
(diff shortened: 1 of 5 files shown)


jkeck (approved):
LGTM 👍

PR #2699, opened 2021-03-04
Fix issues with flaky bulk actions/update specs

### LINKED ISSUE #2693: Fix the flakey tagging spec
We have intermittent failures when the full test suite is run. My guess is due to an ordering issue and data hanging around.

mejackreed:
Paired w/ @jkeck on this one.

Fixes #2693 

Have rerun CI `3` times with all 🍏

### DIFF
--- a/spec/controllers/spotlight/bulk_actions_controller_spec.rb
+++ b/spec/controllers/spotlight/bulk_actions_controller_spec.rb
@@ -2,7 +2,7 @@
 
 describe Spotlight::BulkActionsController, type: :controller do
   routes { Spotlight::Engine.routes }
-  let(:exhibit) { FactoryBot.create(:exhibit) }
+  let(:exhibit) { FactoryBot.create(:skinny_exhibit) }
 
   before do
     allow(Spotlight::ChangeVisibilityJob).to receive(:perform_later)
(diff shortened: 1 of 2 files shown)


spec/controllers/spotlight/bulk_actions_controller_spec.rb
```diff
@@ -2,12 +2,18 @@
...
 describe Spotlight::BulkActionsController, type: :controller do
   routes { Spotlight::Engine.routes }
-  let(:exhibit) { FactoryBot.create(:exhibit) }
+  let(:exhibit) { FactoryBot.create(:skinny_exhibit) }
 
-  before do
-    allow(Spotlight::ChangeVisibilityJob).to receive(:perform_later)
+  # rubocop:disable RSpec/BeforeAfterAll
+  before(:all) do
+    ActiveJob::Base.queue_adapter = :test
```
cbeer, line 9: fwiw, i couldn't find any reason we're using `:inline` at all: https://github.com/projectblacklight/spotlight/pull/2698

mejackreed: Got it thanks.. I'll cherry pick that in and then remove the before/after blocks.

mejackreed: Scratch that.. I still have to set `ActiveJob::Base.queue_adapter = :test` as the ` Internal::Application.config.active_job.queue_adapter` change doesn't seem to have the same impact.

PR #2700, opened 2021-03-04
Add an CSV upload form for bulk updates

jkeck:
Part of sul-dlss/exhibits#1995
Closes sul-dlss/exhibits#1200

## TODO
- [ ] Add tests
- [ ] Look into any possible error handling
- [ ] Hook up to job

Screen Shot 2021-03-04 at 12 11 01 PM


cbeer (approved):
I'm giving this a 👍 and happy for someone else to give it a merge.

I saw you spun off #2708 for performance improvements. I think we could also spin off some stronger validation the provided CSV at least cursorily resembles our CSV format before trying to process it.

And reporting row-level errors is on our todo list somewhere...

And (idle musing..) now that we needed an actual model, should we go full-blown cancancan on the auth permissions? And, when should we switch carrierwave out for activestorage?

jkeck:
Yeah, I thought about the auth given the new model, but figured the seam that we've added is reasonable (at least reads well in its current form: `can? :bulk_update, current_exhibit`), but happy to switch that over to the checking the ability on the model instead.

WRT swapping out Carrierwave for ActiveStorage, that felt a little larger than the scope of this issue, but agreed that is where Spotlight should be going.

mejackreed:
I'm good with this then and just following up on the spin off items.

PR #2711, opened 2021-03-09
Rename ZPR Viewer/Link to "View larger"

### LINKED ISSUE #2709: Rename the "Show in ZPR viewer" button label (Item row widget)
The item row widget includes an option to display a button beneath each image in the widget so that visitors can display a larger version of the image in a modal window. The current label for that button is "Show in ZPR viewer" . Most end-users probably don't know what "ZPR" stands for, and in fact the default viewer currently doesn't support rotate (at least there is no visible button for rotate), so this is not a great label for the action. I suggest:

- [x] Change the "Show in ZPR viewer" button label to "View larger"

Screen Shot 2021-03-05 at 2 59 47 PM

---

If we do that, we should also change the corresponding widget configuration option.

- [x] Rename the Item Row widget option label from "Display ZPR link" to `Offer “View larger” option`

Screen Shot 2021-03-05 at 2 57 54 PM

jkeck:
Closes #2709

## Widget
Screen Shot 2021-03-09 at 11 27 54 AM

## Page
Screen Shot 2021-03-09 at 11 28 06 AM

### DIFF
--- a/app/views/spotlight/sir_trevor/blocks/_solr_documents_block.html.erb
+++ b/app/views/spotlight/sir_trevor/blocks/_solr_documents_block.html.erb
@@ -26,7 +26,7 @@
               </div>
             <% end %>
             <% if solr_documents_block.zpr_link? && block_options[:iiif_tilesource].present? %>
-              <button class="btn btn-secondary zpr-link" data-iiif-tilesource="<%= block_options[:iiif_tilesource] %>">Show in ZPR viewer</button>
+              <button class="btn btn-secondary zpr-link" data-iiif-tilesource="<%= block_options[:iiif_tilesource] %>"><%= t('.zpr_link') %></button>
             <% end %>
           </div>
         </div>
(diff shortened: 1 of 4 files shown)


jkeck:
One thing that jumps out to me from an accessibility perspective (and since we didn't really have this in an i18n label previously anyway), is we may want to consider giving these buttons some context (e.g. what is it viewing larger, to differentiate the buttons w/ the same text from each other).  Happy to work that in as an `.sr-only` addition to the i18n key (or go another route)

app/views/spotlight/sir_trevor/blocks/_solr_documents_block.html.erb
```diff
@@ -26,7 +26,7 @@
               </div>
             <% end %>
             <% if solr_documents_block.zpr_link? && block_options[:iiif_tilesource].present? %>
-              <button class="btn btn-secondary zpr-link" data-iiif-tilesource="<%= block_options[:iiif_tilesource] %>">Show in ZPR viewer</button>
+              <button class="btn btn-secondary zpr-link" data-iiif-tilesource="<%= block_options[:iiif_tilesource] %>"><%= t('.zpr_link') %></button>
```
cbeer, line 29: 😆

ggeisler:
Great point about the a11y aspect of this @jkeck. I was just writing this when I saw your update.

That seems good to me. I was going to suggest using the `alt-text` value but maybe that is always the same thing?

jkeck:
@ggeisler the value that will be in the button will be the same as the alt text on the thumbnail (which should help users connect the two nicely 🤞)

PR #2715, opened 2021-03-10
Let browse#show handle all the other formats that catalog#index norma…

cbeer:
…lly does.

It's a little funky that the browse#show action renders the catalog#index partial, and could lead to some weird behavior.  https://github.com/projectblacklight/blacklight/pull/2432 makes it a little more flexible, but... still weird.


cbeer:
This will probably be closed in favor of #2716. It's a viable approach, but deferring entirely to the catalog controller is more predicable and probably better overall (alternatively... split the BrowseController into a BrowseIndexController and a BrowseSearchController that subclasses Spotlight::CatalogController entirely?).

PR #2716, opened 2021-03-10
Have catalog#index handle all non-HTML search responses for browse categories

cbeer:
By using catalog#index in the first place, things like rss/atom/json feeds, heatmaps, range limit, facet modals, etc will just work.

Upstream needs:

- [x] https://github.com/projectblacklight/blacklight/pull/2434
- [x] https://github.com/projectblacklight/blacklight/pull/2435
- [x] follow-on task to update Blacklight::Solr::Request to automagically use boolean JSON DSL queries as needed?


app/models/concerns/spotlight/browse_category_search_builder.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    # Adds a filter that excludes resources that have been marked as not-visible
+    def restrict_to_browse_category(solr_params)
+      return unless current_exhibit && blacklight_params[:browse_category_id].present?
+
+      search = current_exhibit.searches.find(blacklight_params[:browse_category_id])
+      search_builder = self.class.new(@processor_chain, @scope)
+      browse_category_search_params = search_builder.with(search.query_params).to_hash
+
+      # something clever with q
```
cbeer, line 20: For our sanity, I think the "something clever" needs to be an extension on https://github.com/projectblacklight/blacklight/pull/2434.

Previously, it looks like we just concatenated the two query strings together and called it a day, which probably caused some weird behavior wrt Solr's `mm` making some of the browse category tokens optional. 🤷‍♂️

camillevilla (approved):

PR #2718, opened 2021-03-11
Make BulkUpdatesController 'live'

cbeer:
🤷‍♂️

### DIFF
--- a/app/controllers/spotlight/bulk_updates_controller.rb
+++ b/app/controllers/spotlight/bulk_updates_controller.rb
@@ -6,6 +6,8 @@ module Spotlight
   ##
   # Controller enabling bulk functionality for items defined in a spreadsheet.
   class BulkUpdatesController < Spotlight::ApplicationController
+    include ActionController::Live
+
     before_action :authenticate_user!
     before_action :check_authorization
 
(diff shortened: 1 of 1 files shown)


app/controllers/spotlight/bulk_updates_controller.rb
```diff
@@ -6,6 +6,8 @@ module Spotlight
   ##
   # Controller enabling bulk functionality for items defined in a spreadsheet.
   class BulkUpdatesController < Spotlight::ApplicationController
+    include ActionController::Live
```
jkeck, line 9: Perhaps I'm a bit naive on `ActionController::Live` (just reading up on it now) but would this require us to do some addl. things in order for this to work out (like setting a different content-type and writing to the `response.stream`) or is there some benefit we get just from including?

cbeer: I'm hoping so; there's some suggestion without it, Rails buffers stuff... whether that applies when we set `response_body`, I'm not sure (but, `Live` does override `response_body=`, so 🤷 )

jkeck: 👌

mejackreed (approved):

PR #2720, opened 2021-03-11
Add vertical space below browse group filter buttons

ggeisler:
It isn't super obvious in the default Spotlight implementation, but when you style the browse group filter links to look like buttons, there isn't any vertical spacing between rows when you have enough group filters to wrap to a second row:

Screen Shot 2021-03-11 at 11 16 13 AM

---

Screen Shot 2021-03-11 at 11 15 25 AM

---

This PR adds some bottom margin to the browse group filter links to provide spacing between rows when the links wrap:

Screen Shot 2021-03-11 at 11 12 53 AM

---

Screen Shot 2021-03-11 at 11 12 10 AM

---

(I had to be more specific with the selector here than ideal but it was either that or use `!important` to override the default styling for `.main-container .nav`)

### DIFF
--- a/app/assets/stylesheets/spotlight/_browse.scss
+++ b/app/assets/stylesheets/spotlight/_browse.scss
@@ -38,6 +38,14 @@ $image-overlay-max-height: 300px;
   }
 }
 
+#main-container .browse-group-navigation.nav {
+  margin-bottom: $spacer;
+
+  .nav-link {
+    margin-bottom: $spacer * .75;
+  }
+}
+
 .browse-landing {
   text-align: center;
   // Placeholder for vertically alignment - might already be available from use in another feature


app/assets/stylesheets/spotlight/_browse.scss
```diff
@@ -38,6 +38,14 @@ $image-overlay-max-height: 300px;
   }
 }
 
+#main-container .browse-group-navigation.nav {
+  margin-bottom: 1rem;
```
mejackreed, line 42: Can we use `$spacer` here? (also below? `($spacer * .75)` or similar?)

mejackreed (changes requested):

app/assets/stylesheets/spotlight/_browse.scss
ggeisler: Sure (got out of the habit of using that, I guess).

PR #2726, opened 2021-03-24
Try a work-around for streaming that may work for passenger + puma at…

cbeer:
… the same time?

It seems like `ActionController::Live` just doesn't work with puma at all (even if you copy their example code into a brand new rails app...). The `Last-Modified` header trick seems to be the recommended work-around for some sort of Rack "feature" that breaks streaming. 🤷‍♂️

### DIFF
--- a/app/controllers/spotlight/bulk_updates_controller.rb
+++ b/app/controllers/spotlight/bulk_updates_controller.rb
@@ -12,6 +12,7 @@ module Spotlight
     def edit; end
 
     def download_template
+      # Set Last-Modified as a work-around for https://github.com/rack/rack/issues/1619
       headers['Last-Modified'] = ''
       headers['Cache-Control'] = 'no-cache'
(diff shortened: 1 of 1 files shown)


cbeer:
I've deployed this to exhibits-stage and can confirm streaming responses still work there.

app/controllers/spotlight/bulk_updates_controller.rb
```diff
@@ -6,14 +6,13 @@ module Spotlight
...
   class BulkUpdatesController < Spotlight::ApplicationController
-    include ActionController::Live unless ENV['CI']
-
     before_action :authenticate_user!
     before_action :check_authorization
 
     def edit; end
 
     def download_template
+      headers['Last-Modified'] = ''
```
cbeer, line 15:
```suggestion
      # Set Last-Modified as a work-around for https://github.com/rack/rack/issues/1619
      headers['Last-Modified'] = ''
```

PR #2730, opened 2021-03-24
Add a job to rename/delete tags

### LINKED ISSUE #2664: Add “Rename” action to the Tags admin page
Because we expect that when the proposed bulk actions feature is implemented we’ll see an increase in the use of exhibit item tags, we should make the current Curation > Tags admin page more useful. This ticket suggests adding a “Rename” action to the page so that the curator can easily rename an existing tag.

## Add a “Rename” button to the current tags table

rename-button

---

- [ ] In each row of the table, add a `Rename` button to the left of the Delete button
- [ ] Use `btn btn-secondary btn-sm` for styling the button

## Enable user to rename button

tag-edit

---

When the Rename button is selected in a given row of the tags table:

- [ ] A: Replace the tag link with an input with the current tag label as the input text
- [ ] B: Expand the row height to accommodate Cancel and Save changes buttons
- [ ] B: The Cancel and Save changes buttons should be right-aligned in the same `<td>` as the tag input (or appear that way)
- [ ] The Cancel action reverts the row to its initial state, hiding the input, the action buttons, and displaying the tag link
- [ ] The Save action also reverts the row to its initial state, hiding the input and the action buttons, but displays a tag link with the new value of the input

For example, if the curator renamed the “oranges” tag label above to “tangerines” and selected the Save changes button, the page updates to look like this:

after-edit

cbeer:
Part of #2664 

https://user-images.githubusercontent.com/111218/113211394-44791600-922a-11eb-9147-7beb56e56a1e.mp4

### DIFF
--- a/app/assets/stylesheets/spotlight/_nestable.scss
+++ b/app/assets/stylesheets/spotlight/_nestable.scss
@@ -114,3 +114,11 @@ tr.dd-item {
 .dd3-content .card-header {
   padding: .5rem 1rem
 }
+
+.dd3-content.tag {
+  .dd3-handle {
+    display: none;
+  }
+
+  padding-left: 0;
+}
--- a/app/controllers/spotlight/tags_controller.rb
+++ b/app/controllers/spotlight/tags_controller.rb
@@ -35,5 +35,31 @@ module Spotlight
 
       redirect_to exhibit_tags_path(@exhibit)
     end
+
+    def update_all
+      tags_to_rename = batch_update_params['owned_tags_attributes'].values.select do |tag|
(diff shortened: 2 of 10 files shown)


ggeisler:
@cbeer @mejackreed This looks fine to me. As I discussed with @cbeer, I don't love that the rename action is hidden behind the fairly unobvious edit-in-place feature, but of course we do use that feature elsewhere and we can document that that is how you rename tags. And I think we're better off giving curators a way to rename tags this way than not being able to rename at all because the more ideal way suggested in https://github.com/projectblacklight/spotlight/issues/2664 isn't currently feasible.

camillevilla (approved):
🚢

PR #2739, opened 2021-03-29
Remove carrierwave dependency from Spotlight, relying on ActiveStorage

mejackreed:
- [ ] determine if we need/want to update existing SirTrevor blocks with uploaded item urls
 - [ ]  if we do ^^ , allow for the migration task to cleanup old carrierwave files


app/services/spotlight/carrierwave_file_resolver.rb
```diff
@@ -9,10 +9,10 @@ def initialize
     end
 
     def pattern(id)
-      uploaded_file = Spotlight::FeaturedImage.find(id).image.file
+      uploaded_file = Spotlight::FeaturedImage.find(id).image.blob
       raise Riiif::ImageNotFoundError, "unable to find file for #{id}" if uploaded_file.nil?
 
-      uploaded_file.file
+      ActiveStorage::Blob.service.path_for(uploaded_file.key)
```
mejackreed, line 15: I'm not sure how this will work with other types of ActiveStorage::Service, but should those people just write their own file resolver?

lib/tasks/spotlight_tasks.rake
```diff
@@ -77,6 +77,43 @@ namespace :spotlight do
...
+    puts 'Mirating Spotlight::Attachment'
+    Spotlight::Attachment.find_each do |attachment|
+      next if attachment.file.blob
+
+      attachment.file.attach(
+        io: File.open(Rails.root.join("public/uploads/spotlight/attachment/file/#{attachment.attributes['file']}")),
+        filename: attachment.attributes['file'],
+        content_type: Mime::Type.lookup_by_extension(File.extname(attachment.attributes['file'])[1..])
+      )
+    end
```
mejackreed, line 91: Should we clean up the old files?

app/models/spotlight/attachment.rb
```diff
@@ -4,11 +4,18 @@ module Spotlight
...
+    # Open to alternatives on how to do this, but this works
+    include Rails.application.routes.url_helpers
+
     belongs_to :exhibit
-    mount_uploader :file, Spotlight::AttachmentUploader
+    has_one_attached :file
 
-    def as_json(options = nil)
-      file.as_json(options).merge(name: name, uid: uid, attachment: to_global_id)
+    def as_json(_options = nil)
```
mejackreed, line 13: This is super weird, but seems like we need to provide the accessible url at this point for our uploader forms to work correctly.

app/services/spotlight/carrierwave_file_resolver.rb
cbeer: Yeah..  I think we might have to adapt RIIIF to handle some of this better. I think  ` ActiveStorage::Blob.service.path_for` won't work with some of the remote storage options, right?

(but... the whole download-with-a-block API won't play nice with RIIIF)

mejackreed:
Putting this on ice as it seems like it may cause some additional complexities.

jcoyne:
It looks like this work was wrapped in #3051 . Thank you.

PR #2740, opened 2021-03-31
Add an exhibit viewer role; fixes #2585

### LINKED ISSUE #2585: Add a "Viewer" role for read-only access to unpublished exhibit
We sometimes have cases where an exhibit curator needs to get stakeholder feedback, or for other reasons needs to share an unpublished exhibit with users that the curator does not want to have any capability of editing the exhibit directly.

A less-than-ideal workaround is to temporarily publish the exhibit and then un-publish it again after the stakeholder has seen what they need to see.

A more effective approach might be:

- Add a new role to Configuration > Users: `Viewer`
- Enable a user logged in as "Viewer" to view an unpublished exhibit, but not to have access to the Edit action of pages, and not to have access to the exhibit's administration section at all.

Screen Shot 2020-11-18 at 12 32 00 PM

(In at least the Stanford context, this is still a limited solution, in that we restrict adding users to people who have a SUNet ID, but it would still solve some cases that have come up.)

cbeer:


lib/spotlight/engine.rb
```diff
@@ -305,6 +305,6 @@ def self.call(template, source = nil)
 
     config.assign_default_roles_to_first_user = true
 
-    config.exhibit_roles = %w[admin curator]
+    config.exhibit_roles = %w[admin curator viewer]
```
mejackreed, line 308: There is a little confusion here by me about the relationship between this configuration `exhibit_roles` and the method `exhibit_roles`. Just confirming those are not the same?

cbeer: The method on User?

mejackreed: Yes.. so `User#exhibit_roles` are roles where I can do stuff on the Exhibit. And the configuration.. does??

cbeer: Yeah; the engine configuration is the list of possible roles... and the User#exhibit_roles is the list of (curator-like) roles a user has on an exhibit. Confusing? yes... but 🤷‍♂️ ?

mejackreed: Got it.. 🆗

mejackreed (approved):

PR #2749, opened 2021-05-03
Remove funky parent search count from search title partial

mjgiarlo:
@cbeer says it is best.


camillevilla (approved):

camillevilla:
Getting out of the business of passing around `browse_q` _is_ good.

PR #2752, opened 2021-05-05
Check that we rendered the right view in search results block

cbeer:
Replaces #2748, now that an upstream fix is available


mjgiarlo (approved):

mjgiarlo:
@camillevilla oops, I just noticed you were tagged for a review on this one. Should I revert?

PR #2796, opened 2022-02-16
Truncate titles in breadcrumbs and limit them to 1 line

### LINKED ISSUE #2787: Truncate long breadcrumbs
It appears that our strategy to truncate long breadcrumbs regressed at some point. The examples below are from DLME but I see similar behavior in Stanford exhibits so I'm assuming it's a core Spotlight issue.

It appears that we fixed this (or thought we did) last January [in this PR](https://github.com/projectblacklight/spotlight/pull/2593).

Here's a [sample record with a long title](https://dlmenetwork.org/library/catalog/81055%2Fvdc_100024174406.0x000009_dlme).

And [this is the record](https://dlmenetwork.org/library/catalog/themaghribpodcast.podbean.com%2Faux-origines-de-l%25e2%2580%2599imprimerie-en-alge%25cc%2581rie-les-mythes-les-me%25cc%2581dias-et-les-masses-b2b721c28edd82a13e82e631f08acb5b) used in PR #2593 to show that it was apparently working then.

## Current behavior
Screen Shot 2021-11-19 at 10 21 57 AM

---

Screen Shot 2021-11-19 at 10 46 40 AM

## Desired behavior

Screen Shot 2021-11-19 at 10 49 35 AM

---

Screen Shot 2021-11-19 at 10 48 50 AM

marlo-longley:
Fixes #2787

This fix works by applying the existing truncation CSS to the block-level `.breadcrumb-item` element rather than a `span`. Because `span`'s don't have width, the truncation wasn't working.

I also addressed another issue I came across, which was that titles with multiple metadata sections were being displaying as multiple truncated lines. I thought it looked best to display only the first line. Here's what I mean by multiple metadata sections in the title:
Screen Shot 2022-02-16 at 10 51 40
Screen Shot 2022-02-16 at 10 51 24

The result looks like this:
Screen Shot 2022-02-16 at 10 58 42


app/assets/stylesheets/spotlight/_breadcrumbs.scss
```diff
@@ -1,8 +1,10 @@
...
-  .truncated-value {
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
-    width: 30ex;
-  }
+    max-width: 30ex;
+}
+
+.breadcrumb-item > span > :not(:first-child) {
```
cbeer, line 8: I feel like this rule is a little too specific to DLME to go upstream.

app/assets/stylesheets/spotlight/_breadcrumbs.scss
```diff
@@ -1,8 +1,10 @@
 .breadcrumb-item {
```
cbeer, line 1: I believe `breadcrumb-item` comes from bootstrap, so I think we should avoid overriding the upstream styling.

marlo-longley: @cbeer Ok makes sense, thanks for reviewing! I added the specific class `truncated-value` back into the code. It is added at the same level as `breadcrumb-item` however, rather than to the child span like before, so that the styling will work.

marlo-longley: I will submit a DLME PR for that other issue since I already wrote the code

PR #2805, opened 2022-02-28
Cut our dependency on the sir_trevor_rails gem 

cbeer:
...by bringing in the last pieces we haven't overridden ourselves.


app/models/sir_trevor_rails/block.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'ostruct'
+
+module SirTrevorRails
+  # Forked from (former) upstream sir-trevor-rails 0.6.2 gem:
+  # https://github.com/madebymany/sir-trevor-rails/blob/931b9554f5268158b4da8817477cdc82e4e2e69c/lib/sir_trevor_rails/block.rb
+  # Copyright (c) 2013-2014 by ITV plc - http://www.itv.com
+  class Block < OpenStruct
+    DEFAULT_FORMAT = :markdown
```
thatbudakguy, line 10: Doable to not use `OpenStruct` here? or not worth the effort currently

thatbudakguy (approved):
fine with porting the rest of this, since there's so little to speak of. hopefully it just makes it easier to continue modifying.

app/models/sir_trevor_rails/block.rb
cbeer: Nope. Each type of block can have its own data.

lib/generators/spotlight/templates/config/initializers/sir_trevor_rails.rb
```diff
@@ -1,10 +0,0 @@
-# frozen_string_literal: true
-
-# Required for 0.6 and up:
-# https://github.com/madebymany/sir-trevor-rails#upgrade-guide-to-v060
-class SirTrevorRails::Block
```
dunn, line 5: this initializer has to be manually removed on existing apps; otherwise you get this error during `db:migrate` https://gist.github.com/dunn/d20dedc996fd7fee7c9261ae72a513dc

are there upgrade notes mentioning this?

PR #2807, opened 2022-02-28
Allow paper_trail v12

cbeer:


thatbudakguy:
what's this for?

cbeer:
We track model changes (e.g. https://github.com/projectblacklight/spotlight/blob/master/app/models/spotlight/page.rb#L38)  so we can revert stuff if needed. I don't think we've ever built out friendly tools, but being able to get previous revisions from the rails console is handy 🤷‍♂️

thatbudakguy:
were we pinning it because of a different version constraint that's now resolved? just wondering why we now can use v12

cbeer:
Ah. I think we were just pessimistically pinning the version in case there was a breaking change going from 11 -> 12, but there wasn't.

thatbudakguy (approved):

PR #2813, opened 2022-04-14
Render Analytics link in sidebar when enabled

mcritchlow:
If Analytics is not enabled, we should not show an Analytics link in the
sidebar for an Exhibit.


mcritchlow:
Just noting I'm still working on getting a working dev environment locally so I can make sure these tests are setup properly. Having something like #2586 in would make this a lot simpler, at least for those of us accustomed to working with containerized environments.

dunn (approved):

Issue #2814, opened 2022-05-05
clicking image in exhibit goes to catalog show page and includes search buttons


elrayle:
### Description

In Spotlight 3.3.0, when clicking on an image in an exhibit, the catalog show page includes search buttons.  It does not include the pagination links.

----
#### From exhibit image

When clicking an image in an exhibit, it should navigate to the catalog show page.  
* The page should NOT include the Start Over and Back to Search buttons.  
* The page should NOT include the pagination navigation area.

![image](https://user-images.githubusercontent.com/6855473/167022899-1d3a96eb-0bfd-4a32-beb3-4f9006697924.png)

#### From search

When searching within an exhibit and clicking one of the search results, it should navigate to the catalog show page.
* The page should include the Start Over and Back to Search buttons.  
* The page should include the pagination navigation area.

![image](https://user-images.githubusercontent.com/6855473/167022347-bd4f7a04-6fc0-41fa-9051-96da2fca5420.png)

----
#### From exhibit image

When clicking an image in an exhibit, it should navigate to the catalog show page.  
* The page includes the Start Over and Back to Search buttons.  (wrong behavior)
* The page does not include the pagination navigation area. (correct behavior)

_NOTE: This image shows the Start Over and Back to Search buttons included when they shouldn't be._
_NOTE: The size difference is due to the images being captured at different times._

![image](https://user-images.githubusercontent.com/6855473/167021759-548c2c3c-ac95-4a6e-a165-4600dfed9e78.png)

#### From search

When searching within an exhibit and clicking one of the search results, it should navigate to the catalog show page.
* The page includes the Start Over and Back to Search buttons.  (correct behavior)
* The page includes the pagination navigation area. (correct behavior)

_NOTE: This image is the same as that shown in expected._

----
### Steps to reproduce

From an exhibit image:
1) navigate to an exhibit that has an image that can be clicked to see its show page
2) click the image from the exhibit

Page should not have Start Over and Back to Search buttons.  Page should not have pagination area.

From exhibit search:
1) navigate to an exhibit that supports search
2) type a query in the search box (preferably one that will have at least 2 results)
3) click on one of the results

Page should have Start Over and Back to Search buttons.  Page should have pagination area.

corylown:
We think this is Blacklight issue. In specific implementations these controls have been turned off in this view. Please re-open if you think this is still an issue.

PR #2815, opened 2022-05-05
do not show search buttons when catalog show comes from an image click in an exhibit

elrayle:
Fixes Issue #2814 

In Spotlight 2.x & Blacklight 6.x, the search buttons and the pagination links were included in the same area of the page.  They were not rendered when `current_search_session_from_page?` returns `true`.

In Blacklight 7.x, the decision to show the search buttons and the pagination links are handles separately and displayed in separate areas of the page.  Spotlight 3.x overrides the code that determines whether or not to display the pagination in the [catalog controller](https://github.com/projectblacklight/spotlight/blob/v3.3.0/app/controllers/spotlight/catalog_controller.rb#L135-L146).  _NOTE: This same code is used in Spotlight 2.x._

In Blacklight 7.x determines whether to show the search buttons based on whether or not there is a search session.  This check is performed in the [catalog/show view](https://github.com/projectblacklight/blacklight/blob/v7.25.0/app/views/catalog/show.html.erb#L1-L6).

This PR overrides the catalog/show view and only displays the search buttons if `current_search_session_from_page?` returns `false`.

### DIFF
--- a/app/controllers/spotlight/concerns/catalog_search_context.rb
+++ b/app/controllers/spotlight/concerns/catalog_search_context.rb
@@ -5,6 +5,15 @@ module Spotlight
     ##
     # Search context helpers
     module CatalogSearchContext
+      extend ActiveSupport::Concern
+
+      # The following code is executed when someone includes Spotlight::Concerns::CatalogSearchContext in their
+      # own controller.
+      included do
+        # Provides access to the method in views as a helper method.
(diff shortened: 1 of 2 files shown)


elrayle:
I removed all my changes and tests still failed.  I don't believe the test failures are related to the changes in this PR.

PR #2819, opened 2022-05-23
Fix sort order in sir trevor displayable mixin

jzimpel:
Solves #2820 

Several blocks of sir-trevor-rails display their items in a wrong order
(other than set in the backend) if there are more than ten items.
This bug concerns all blocks including the mixin Displayable:
- FeaturedPagesBlock
- BrowseBlock
- LinkToSearchBlock
- SearchResultsBlock
- SolrDocumentsBlock

The items are sorted in a non-natural order and thereby the item with a
weight of 10 is displayed between the ones with a weight of 1 and that
with a weight of 2.
As ruby does not provide a method for a natural sort order, fix this by
padding the weight with leading zeroes in Displayable#ordered_items

### DIFF
--- a/app/models/sir_trevor_rails/blocks/displayable.rb
+++ b/app/models/sir_trevor_rails/blocks/displayable.rb
@@ -14,7 +14,7 @@ module SirTrevorRails
       end
 
       def ordered_items
-        items.sort_by { |x| x[:weight] }.map { |x| x[:id] }
+        items.sort_by { |x| x[:weight].to_s.rjust(5, '0') }.map { |x| x[:id] }
       end
 
       private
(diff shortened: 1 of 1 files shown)


app/models/sir_trevor_rails/blocks/displayable.rb
```diff
@@ -14,7 +14,7 @@ def item_ids
       end
 
       def ordered_items
-        items.sort_by { |x| x[:weight] }.map { |x| x[:id] }
+        items.sort_by { |x| x[:weight].to_s.rjust(5, '0') }.map { |x| x[:id] }
```
cbeer, line 17:
```suggestion
        items.sort_by { |x| x[:weight].to_i) }.map { |x| x[:id] }
```

Could we just make sure to parse the weight as an integer?

jzimpel: Yes, I like that solution. In fact I thought about a cast to integer as well, but I'm relatively new to ruby and so I was not sure if there might be cases when this can fail or raise an exception.

thatbudakguy (approved):

PR #2830, opened 2022-06-28
Remove hard dependency on postgres

cbeer:


corylown (approved):
I tried verifying that this doesn't break the docker setup, but it appears to already be broken in the `master` branch.

cbeer:
Thanks. Let's ship this to unbreak regular deployments and deal with the docker dev environment separately.

PR #2845, opened 2022-09-21
Revert "Remove hard gem dependencies on application dependencies"

dunn:
This reverts commit 83a6c47b959b9d05dbd3a437736a077d5630355f.

Including the DB setup during `rails new` causes ActiveRecord to attempt to initialize a database connection, which fails because there's no DB running during the image build: https://gist.github.com/dunn/8abda1b63399c41226cea28662765fc0


jcoyne (changes requested):
This seems like a pretty extensive change.  Would it be possible to remove the `-d postgresql` and add `RUN bundle add pg` ?

blacklight-spotlight.gemspec
```diff
@@ -43,23 +43,27 @@ these collections.)
...
   s.add_dependency 'leaflet-rails'
   s.add_dependency 'legato'
+  s.add_dependency 'mimemagic', '0.3.8'
   s.add_dependency 'mini_magick'
   s.add_dependency 'nokogiri'
   s.add_dependency 'oauth2'
   s.add_dependency 'openseadragon'
   s.add_dependency 'ostruct', '!= 0.3.0', '!= 0.3.1', '!= 0.3.2'
   s.add_dependency 'paper_trail', '>= 11.0', '< 13'
+  s.add_dependency 'pg'
```
cbeer, line 53: I think it is inappropriate to make `pg` a hard dependency of spotlight. It works with other databases too (same for sidekiq, webpacker, etc)

cbeer (changes requested):

dunn:
@jcoyne yes, it was a pretty extensive change, and broke docker builds, which is why i'm proposing we revert it.  if we want to remove the hard dependencies on postgres gems, that can be a follow-up PR, tested against the docker builds.

dunn:
to that end, added a workflow step that just checks to make sure the docker image builds

dunn:
@jcoyne's fix works, so I've updated the PR.  Though for the record, I would have appreciated reverting the breaking commit and then providing a working replacement.

jcoyne (approved):

PR #2846, opened 2022-09-30
pin view_component to < 2.74

dunn:
until there's a tagged blacklight release including https://github.com/projectblacklight/blacklight/pull/2818

PR #2847, opened 2022-10-04
add iiif_service config option to allow external IIIF server

dunn:
This is an attempt to extract references to RIIIF into a new `RiiifService` so that in the future applications can swap it out with a service module that connects to an external IIIF server (like Cantaloupe) instead.

There are still references I haven't removed yet, and this includes ~~https://github.com/projectblacklight/spotlight/pull/2845~~ and https://github.com/projectblacklight/spotlight/pull/2846 for ease of development.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -216,8 +216,6 @@ Rails/WhereExists: # (new in 2.7)
   Enabled: true
 Rails/WhereNot: # (new in 2.8)
   Enabled: true
-Gemspec/DeprecatedAttributeAssignment: # (new in 1.10)
-  Enabled: true
 Lint/NumberedParameterAssignment: # (new in 1.9)
   Enabled: true
 Lint/OrAssignmentToConstant: # (new in 1.9)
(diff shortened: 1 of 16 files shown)


dunn:
bizarrely, only seeing this error on Ruby 3: https://github.com/projectblacklight/spotlight/actions/runs/3201034405/jobs/5228805376#step:5:4829

dunn:
i'm not sure what's happening here but it seems like the object isn't being saved? https://github.com/projectblacklight/spotlight/actions/runs/3221407448/jobs/5269301341#step:5:4853

dunn:
can confirm this allows spotlight to use an external IIIF server (in this case, cantaloupe): https://gitlab.com/surfliner/surfliner/-/merge_requests/2767

hackartisan (commented):
Great idea. Thanks for doing this! I think it looks good but it would be nice for someone else to review as well.

tpendragon (approved):

PR #2854, opened 2023-01-10
Allow view component to use the latest releases.

jcoyne:


dunn:
is there a blacklight release that includes https://github.com/projectblacklight/blacklight/pull/2818 and is used by spotlight?

dunn:
oh i see, 7.30.0: https://github.com/projectblacklight/blacklight/compare/v7.29.0...v7.30.0

PR #2871, opened 2023-02-13
Revert adding shakapacker

jcoyne:
Spotlight has lots of turbolinks code, so let's test it.

Reverts part of https://github.com/projectblacklight/spotlight/pull/2865/commits/ccf26faf6d9f89f556d50ea726d97be780088ee6

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,6 @@ RUN apk --no-cache upgrade && \
   libxml2-dev \
   libxslt-dev \
   nodejs \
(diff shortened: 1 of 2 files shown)


dunn:
The removal of turbolinks was probably unnecessary; I added it when I was debugging errors that probably would have been resolved with the `require_tree .` that I later restored.  So I'd first try just dropping the line that removes turbolinks.

dunn (approved):

PR #2876, opened 2023-02-14
Test on latest version of solr

jcoyne:


docker-compose.yml
```diff
@@ -48,14 +48,21 @@ services:
...
+      - $PWD/lib/generators/spotlight/templates/solr/conf:/opt/solr/conf
+    ports:
+      - "${SOLR_PORT}:8983"
+    entrypoint:
+      - docker-entrypoint.sh
+      - solr-precreate
+      - blacklight-core
+      - /opt/solr/conf
+      - "-Xms256m"
+      - "-Xmx512m"
```
thatbudakguy, line 65: why the JVM memory settings? I usually don't bother in docker unless something is breaking

thatbudakguy (approved):

docker-compose.yml
jcoyne: I just copied from blacklight: https://github.com/projectblacklight/blacklight/blob/main/docker-compose.yml#L33-L39  Want me to remove this?

thatbudakguy: yeah, I think unless we need it better not to set anything.

jcoyne: Okay, they've been removed.

thatbudakguy: thank you!

PR #2884, opened 2023-02-14
Remove unused chrome-selenium container

jcoyne:


thatbudakguy:
@dunn do you know if this is/was used for anything?

dunn:
not sure; we use chromium for our spotlight app's integration tests, so i may have just assumed it was needed here.

thatbudakguy:
i think integration tests can be run using chromium from outside the containers, so i'm gonna go ahead and merge this since there doesn't seem to be any code related to using chromium from within the container

thatbudakguy (approved):

PR #2886, opened 2023-02-15
Stop including unnecessary polyfill library

jcoyne:


thatbudakguy (commented):
It looks like the change that brought this in (https://github.com/projectblacklight/spotlight/commit/a9fcdaf9c29a21dff67fbe9e308cb3d2592496d2) was focused on pulling IIIF images into editor blocks. I think the babel polyfill was included to ensure support for things like `symbol.iterator` and `Array.from` that are used in our [`Iiif` class](https://github.com/projectblacklight/spotlight/blob/main/app/assets/javascripts/spotlight/admin/iiif.es6). At the time (2017), `symbol.iterator` was [supported in all but safari](https://caniuse.com/?search=symbol.iterator) and `Array.from` was [supported everywhere](https://caniuse.com/?search=Array.from), but they were both still fairly new.

5 years later (and hopefully not caring about IE), I don't think we need this — or at least we don't need to vendorize it; downstream can feel free to roll their own polyfills or whatever. Not sure why tests are failing though.

corylown (approved):

PR #2887, opened 2023-02-15
Use the preconfigured javascript driver

jcoyne:


spec/features/exhibits/translation_editing_spec.rb
```diff
@@ -295,6 +295,7 @@
 
         visit spotlight.search_exhibit_catalog_path(exhibit, q: '*', locale: 'fr')
 
+        find('button[aria-label="plus »"]').click # Hamburger icon
```
thatbudakguy, line 298: was a test failing without this?

thatbudakguy (approved):

spec/features/exhibits/translation_editing_spec.rb
jcoyne: Yes, because at the default width, the facets are folded up.

PR #2913, opened 2023-02-22
Upgrade sir-trevor to 0.7.5

jcoyne:


thatbudakguy (approved):
are we blocked by the gem from going further? it looks like the js portion goes up to v0.9.0: https://github.com/madebymany/sir-trevor-js/tags

jcoyne:
@thatbudakguy I was just trying to take it a step at a time. That allows us to back up if there is a problem.

PR #2918, opened 2023-02-22
Export Spotlight so that it's configuration may be modified

jcoyne:


thatbudakguy (approved):
needs rebase because of rebuilt js

PR #2929, opened 2023-02-28
Move default blacklight config to after_initialize

### LINKED ISSUE #2928: Blacklight defaults are not being set
I'm seeing some cases in a development server (puma) where these values are not "remembered" by the code that needs them.

https://github.com/projectblacklight/spotlight/blob/main/lib/spotlight/engine.rb#L251-L262

This is manifesting as:
```
Blacklight::Exceptions::InvalidRequest in Spotlight::CatalogController#admin
RSolr::Error::Http - 400 Bad Request
Error: Can't determine a Sort Order (asc or desc) in sort spec ' desc', pos=5
```
because `blacklight_config.index.timestamp_field` is `nil` when we try to use it here: https://github.com/projectblacklight/spotlight/blob/76a2b437f1e34854cf4bde43625089d4c2a14e96/app/controllers/spotlight/dashboards_controller.rb#L54

jcoyne:
Fixes #2928

### DIFF
--- a/lib/spotlight/engine.rb
+++ b/lib/spotlight/engine.rb
@@ -248,7 +248,7 @@ module Spotlight
 
     config.spambot_honeypot_email_field = :email_address
 
-    config.to_prepare do
+    config.after_initialize do
       Blacklight::Configuration.try(:initialize_default_configuration) unless Blacklight::Configuration.try(:initialized_default_configuration?)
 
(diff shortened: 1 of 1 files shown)


jcoyne:
The fix is in https://github.com/projectblacklight/blacklight/pull/3028

PR #2930, opened 2023-03-01
Use Blacklight::Configuration.default_configuration to configure blacklight configuration defaults

cbeer:
Fixes https://github.com/projectblacklight/spotlight/issues/2928


jcoyne:
Reworked in #2944

PR #2931, opened 2023-03-01
Make sprockets optional to enable propshaft

jcoyne:


jcoyne:
@thatbudakguy By default rails 7 give you sprockets but you can pass a flag to get propshaft instead.  What doesn't work well is both at the same time.

thatbudakguy (approved):

PR #2932, opened 2023-03-01
Drop unused dependency redcarpet

jcoyne:


cbeer:
Isn't this required for markdown support in github/markup?
https://github.com/github/markup#markups

jcoyne:
@cbeer apparently yes. I thought the github gem would take care of this itself.

PR #2944, opened 2023-03-02
Use Blacklight::Configuration.default_configuration for Blacklight 8

jcoyne:
Blacklight 8 refactored how configuration defaults are set, so Spotlight needed to be adjusted to use the appropriate method depending on the version of Blacklight.

### DIFF
--- a/lib/spotlight/engine.rb
+++ b/lib/spotlight/engine.rb
@@ -248,17 +248,30 @@ module Spotlight
 
     config.spambot_honeypot_email_field = :email_address
 
-    config.to_prepare do
-      Blacklight::Configuration.try(:initialize_default_configuration) unless Blacklight::Configuration.try(:initialized_default_configuration?)
+    if Blacklight::VERSION > '8'
+      Blacklight::Configuration.default_configuration do
+        # Field containing the last modified date for a Solr document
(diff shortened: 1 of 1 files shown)


lib/spotlight/engine.rb
```diff
@@ -248,17 +248,30 @@ def self.call(template, source = nil)
...
-      Blacklight::Configuration.default_values[:browse] ||= Blacklight::OpenStructWithHashAccess.new(document_actions: [])
+        Blacklight::Configuration.default_values[:search_state_fields] ||= []
+        Blacklight::Configuration.default_values[:search_state_fields] += %i[id exhibit_id]
+      end
+    else
+      config.to_prepare do
+        Blacklight::Configuration.try(:initialize_default_configuration) unless Blacklight::Configuration.try(:initialized_default_configuration?)
+
+        # Field containing the last modified date for a Solr document
+        Blacklight::Configuration.default_values[:index].timestamp_field ||= 'timestamp'
```
thatbudakguy, line 267: this seems the same between the two branches? i keep staring at it but can't see what's different

thatbudakguy: oh it's inside `config.to_prepare`, nvm. i see

PR #2970, opened 2023-03-10
Poision the exhibit cache value when a document is updated

### LINKED ISSUE #2968: Browse categories not updating
I’m getting some unexpected behavior using exhibit tags to make browse categories and want to check to see if this is a bug or if its how browse categories are supposed to work.

    I tagged 55 items with the “qalaat-seman” tag
    then built a browse category from all records with the “qalaat-seman” tag
    then I tagged more records with the “qalaat-seman” tag

I was expecting the browse category to include these new records but it doesn’t. I then manually selected the linked tag from one of the records which led back to a search result of all records with that tag. I tried to "update the existing category" from those results but the category was not updated.

This seems to resolve after clearing the rails cache, so there seems to be some place where we aren't poisoning the cache.

Moved from https://github.com/sul-dlss/dlme/issues/1753

jcoyne:
This may fix #2968

### DIFF
--- a/app/models/spotlight/solr_document_sidecar.rb
+++ b/app/models/spotlight/solr_document_sidecar.rb
@@ -8,7 +8,8 @@ module Spotlight
 
     acts_as_taggable
 
-    belongs_to :exhibit, optional: false
+    # The "touch: true" ensures the exhibit's cache is invalidated when the document is updated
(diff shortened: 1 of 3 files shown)


hackartisan:
The referenced issue was closed as not replicable

PR #2971, opened 2023-03-14
testing if value is array and joining to make string

hudajkhan:


app/controllers/concerns/spotlight/base.rb
```diff
@@ -43,6 +42,16 @@ def autocomplete_json_response_for_document(doc)
...
     end
-    # rubocop:enable Metrics/AbcSize
+
+    # Some pipeline steps may return an array instead of a string
+    # to_str throws an error if the object is not a string
+    # This workaround does a join on the array to allow values to be returned
+
+    private
+
+    def autocomplete_title(heading)
```
corylown, line 52: I think it would be clearer to keep the comment with the `autocomplete_title` method rather than the `private` keyword in case other methods are added later. What do you think?
```suggestion
    private

    # Some pipeline steps may return an array instead of a string
    # to_str throws an error if the object is not a string
    # This workaround does a join on the array to allow values to be returned
    def autocomplete_title(heading)
```

corylown (changes requested):
Is there a way to add a test for this case?: pipeline returns an array and we can still create a title for autocomplete

app/controllers/concerns/spotlight/base.rb
hudajkhan: I had it that way and then moved it around. The suggestion sounds good! I'll look at setting up a test. thanks!

app/controllers/concerns/spotlight/base.rb
```diff
@@ -43,6 +42,16 @@ def autocomplete_json_response_for_document(doc)
         iiif_manifest: doc[Spotlight::Engine.config.iiif_manifest_field]
       }
     end
-    # rubocop:enable Metrics/AbcSize
+
+    # Some pipeline steps may return an array instead of a string
```
jcoyne, line 46: All pipeline steps are supposed to return an array except for the last join step. I believe it is the pipeline in DLME is not behaving.  I don't think we need to make this change in Spotlight.

jcoyne (changes requested):

app/controllers/concerns/spotlight/base.rb
hudajkhan: The join step in DLME is here https://github.com/sul-dlss/dlme/blob/main/app/processors/join.rb#L8 . It explicitly disallows the joining of the array into a string if it's a JSON API request (i.e. CatalogController request, JSON format). I added comments to the related DLME GitHub issue here: https://github.com/sul-dlss/dlme/issues/1705#issuecomment-1464717444 .  

If all pipeline steps are supposed to have a last join step that transforms the returned value into a string, is there another approach within the DLME Join we could take? I didn't want to override the original logic which seemed to prevent returned a join result if it's a JSON API request.

jcoyne: I would consider having a separate pipeline for JSON display than for HTML display. 

If you didn't want to go that route, could you consider moving these changes from this PR into DLME instead? That way we don't have to add additional complexity in Spotlight that nobody else needs.

hudajkhan: I can follow up on slack with folks regarding the pipeline.  I'll close this PR for now.

PR #2975, opened 2023-04-07
Add some documentation to solr_documents_base_block.js

tpendragon:
@hackartisan and I had a very, very hard time parsing out how it was possible autocomplete wasn't returning IIIF information but somehow carousels were storing iiif_tilesource. I've renamed a couple variables and added some documentation that I wish we'd had while we were digging.

### DIFF
--- a/app/javascript/spotlight/admin/blocks/solr_documents_base_block.js
+++ b/app/javascript/spotlight/admin/blocks/solr_documents_base_block.js
@@ -48,19 +48,24 @@ SirTrevor.Blocks.SolrDocumentsBase = (function(){
       </div>
     `},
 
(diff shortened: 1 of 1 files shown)


hudajkhan:
We were just reviewing this part of the code base and had to track where iiif_tilesource was even generated, so I agree this documentation would be very helpful. I'll review this shortly.

hudajkhan (approved):
The documentation appears correct. The reference to iiif.js is important as that code adds properties like iiif_sourcetitle.  I also tried out the code in a local Spotlight instance and tried out the variable name changes in another application which uses Spotlight.  These changes seem to work, and the documentation is useful.

PR #2979, opened 2023-08-23
Remove unnecessary webdrivers dependency

jcoyne:


hudajkhan (approved):
Some of the rubocop tests are failing but other rubocop changes are on the way, so approving this.

PR #2988, opened 2023-11-19
Correct the property attribute for og:image meta tags

mephillips-durham:
The opengraph metadata for images was being encoded with a property attribute of og:og:image instead of og:image.


jcoyne (approved):

corylown (approved):

corylown:
Tests are passing locally so I'm going to merge this since I can't see a way to re-run CI here.

PR #2999, opened 2024-01-08
Add Rails 7.1 to the test matrix

jcoyne:


jcoyne:
This seems to be dependent on #2998

.github/workflows/ruby.yml
```diff
@@ -37,11 +37,15 @@ jobs:
             blacklight_version: 'latest'
             bootstrap_version: ~> 4.0
             additional_engine_cart_rails_options: -a propshaft
+          - rails_version: '7.1.2'
```
taylor-steve, line 40: Curious if there's a reason to test on 7.1.2 or if we can jump to 7.1.3.4 now?

corylown: I think we can.

taylor-steve:
The SKIP_TRANSLATION change is going to be nice to have, thanks.

taylor-steve (approved):

PR #3004, opened 2024-01-13
Use lucene query parser for search within user-created browse categories

marlo-longley:
Fixes the downstream issue in https://github.com/sul-dlss/exhibits/issues/2377, and restores this functionality which was broken.


jcoyne (changes requested):
Is it possible to add a feature spec for this?

marlo-longley:
@jcoyne I added more to the search builder test to specify that without a search query, we should not override the default parser. This isn't exactly a feature test, but I think it enhances the coverage and captures the changes. I looked into a feature test but it felt complicated to do the setup necessary to execute a real search. Lmk what you think, thanks for the review.

jcoyne:
@marlo-longley yes, but that's just a unit test.  It says "we're going to use this parser", but it doesn't demonstrate why we have to use that parser.  If you could create a feature test that would be better.

marlo-longley:
@jcoyne thanks for reviewing -- added feature tests.

corylown (approved):

jcoyne (approved):

PR #3010, opened 2024-01-30
Pass the class argument to serialize as a keyword

jcoyne:
The positional argument is deprecated and will be removed in rails 7.2

### DIFF
--- a/app/models/spotlight/blacklight_configuration.rb
+++ b/app/models/spotlight/blacklight_configuration.rb
@@ -10,7 +10,7 @@ module Spotlight
     has_paper_trail
 
     belongs_to :exhibit, touch: true, optional: true
-    if Rails.version > '7'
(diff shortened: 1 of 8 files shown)


corylown:
It looks like the method signature for `ActiveRecord::AttributeMethods::Serialization::ClassMethods#serialize` changed to accept keyword arguments with Rails 7.1 not 7.0 so I'm updating this PR to reflect that.

[Rails 7.0](https://api.rubyonrails.org/v7.0/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize)
[Rails 7.1](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html)

taylor-steve (approved):

PR #3019, opened 2024-06-21
Remove obsolete engine_cart config

jcoyne:


edsu (approved):
I have no idea why this is being removed, but I'm assuming you do :-)

jcoyne:
@edsu because it's for Rails 6.0 and 5.2, which are past EOL.

PR #3021, opened 2024-06-21
Test on ruby 3.3

jcoyne:


.github/workflows/ruby.yml
```diff
@@ -36,6 +36,11 @@ jobs:
             bootstrap_version: ~> 4.0
             additional_engine_cart_rails_options: -a propshaft
             additional_name: '/ Propshaft'
+          - rails_version: '7.0.8'
```
edsu, line 39: Is 7.1.3 too high?

edsu (approved):

.github/workflows/ruby.yml
jcoyne: Yes https://github.com/projectblacklight/spotlight/blob/main/blacklight-spotlight.gemspec#L51

corylown (approved):

Issue #3014, opened 2024-03-04
spotlight starts with an error on ec2 startup


cxu0421:
we followed the depolyment on aws ec2 instance running ruby 3.2.2 and Rails 7.1.3.2, the app does not start correctly, we are getting error similar to https://github.com/projectblacklight/blacklight/issues/3031, after following their solution to adjust the manifest.js, we now get the following error on start up, 

ActionView::Template::Error (Multiple files with the same output path cannot be linked ("application.js")
In "/opt/collections/spotlight/app/assets/config/manifest.js" these files were linked:
  - /opt/collections/spotlight/app/assets/javascripts/application.js
  - /opt/collections/spotlight/app/javascript/application.js
):

jcoyne:
As far as I know, Spotlight does not support Rails 7.1.

mephillips-durham:
Using Rails 7.1 is not the cause of this problem. I [opened an issue](https://github.com/projectblacklight/spotlight/issues/2995) not long ago, and we are using 7.0.x

The generated manifest.js is broken. We got past this error by hand-editing the file to remove the line for application.js and add the missing newline. I don't know whether some of the faults we are seeing in our application might be because of this file being incorrrect though. The fact the line for blacklight_gallery was wrong makes me wonder whether it's connected with the [lack of results page type choices](https://github.com/projectblacklight/spotlight/issues/2993) in our application.

taylor-steve:
A new version of blacklight-gallery was released that fixes the missing newline issue.

PR #3034, opened 2024-06-27
Remove application.js after bundle in template

taylor-steve:
Fixes part of #3014. I believe the `manifest.js` issue is fixed, but unreleased, in blacklight-gallery so that would still require manually fixing that newline if testing this for real.

Moves the removal of `app/javascript/application.js` to the end of the end of the process. Steps that run after the install_generator, like stimulus-rails installing, recreate the file.

https://github.com/hotwired/stimulus-rails/blob/ae4b675473b71fdf01530c8a6a3bb277d3388ee2/lib/install/stimulus_with_importmap.rb#L11

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -34,3 +34,10 @@ insert_into_file 'config/application.rb', after: "< Rails::Application\n" do
   config.active_job.queue_adapter = ENV["RAILS_QUEUE"]&.to_sym || :sidekiq
   CONFIG
 end
+
+after_bundle do
+  # Rails installed importmap by default, but we don't have importmap + Blacklight 7 working yet.
+  # Similar to step in Spotlight::Install#add_manifest, but this cleans up the file stimulus-rails
+  # creates afterward.
+  remove_file 'app/javascript/application.js'
+end


lib/generators/spotlight/install_generator.rb
```diff
@@ -21,9 +21,6 @@ def add_js_rails6
     def add_manifest
       append_to_file 'app/assets/javascripts/application.js', "\n//= require_tree .\n"
       append_to_file 'app/assets/config/manifest.js', "\n//= link spotlight/manifest.js"
-
-      # Rails installed importmap by default, but we don't have importmap + Blacklight 7 working yet.
-      remove_file 'app/javascript/application.js'
```
jcoyne, line 26: Don't we need this even if we're not using the template?  The use of template.rb is not required.

taylor-steve: I was assuming that for fresh Rails apps it still wouldn't do much because stimulus-rails would install after. For existing apps, it seemed slightly questionable to delete their application.js (even if it results in Spotlight not working). I haven't tested those assumptions yet, so I'm happy to leave the generator untouched and work toward getting the template instructions working as written.

taylor-steve:
Closing this, the issue will be addressed in other work.

PR #3055, opened 2024-07-16
[WIP] Replace Nestable with Sortable

taylor-steve:
WIP. Experimenting with replacing the unmaintained Nestable library with Sortable. This is intended as a drop-in replacement, without modifying the existing template markup.

Currently Sortable is in vendor, matching our current behavior until we reorganize the JavaScript.

Sortable needs a little space between the last item of a first level nested list and the end of the containing parent list in order to match the behavior of Nestable (e.g., dragging the last nested item left into the parent list). That's why the padding has been added.

Might be overkill but this works for `data-max-depth` of 1-n, the old implementation was 1-2 (which is all Spotlight currently uses).


app/assets/javascripts/spotlight/application.js
```diff
@@ -3,6 +3,7 @@
 //= require sir-trevor
 //= require clipboard
 //= require tiny-slider
+//= require Sortable.min.js
```
corylown, line 6: Is this only used by the admin interfaces? Is that why the nestable import was in [app/javascript/spotlight/admin/index.js](https://github.com/projectblacklight/spotlight/pull/3055/files#diff-45150080ff3f9be0628913e9f2fc161c5542df2eda3af97bafb429e235180621)?

corylown (commented):
This seems to work well. I tested it on the Feature/About pages admin widget and the sir trevor blocks.

app/assets/javascripts/spotlight/application.js
taylor-steve: Yes, I believe that's the case. I'll shuffle this back.

taylor-steve:
Great, thanks for testing. I think in terms of functionality it's nearly 1:1. There are styling differences on the ghost and the target destination, mostly around the current target being an outline only. Do you think that's a concern or are we willing to use Sortable's default styling there? I can get Alan's take on this if we need more input.

Otherwise, I have a bit more testing specifically in the Sir Trevor blocks I'd like to do and maybe some clean-up. Then I'll propose this for real.

corylown:
@taylor-steve I'm not concerned about the styling changes, but does seem worth running by Alan.

taylor-steve:
Noting that nobody had concerns about the difference in styling. I do think it may be prudent to have one or two folks evaluate the ergonomics of nesting (e.g., feature pages). It has a slightly different feel than nestable, but I think it's quite close.

Once we're ready to move to import maps we could pin Sortable and drop the local file if we wished.

cbeer (approved):

corylown (approved):

PR #3058, opened 2024-07-17
Configure the show document component so that upstream can override it

corylown:
This might help us fix https://github.com/sul-dlss/exhibits/issues/2479

### DIFF
--- a/app/controllers/spotlight/catalog_controller.rb
+++ b/app/controllers/spotlight/catalog_controller.rb
@@ -22,10 +22,6 @@ module Spotlight
 
     before_action :load_document, only: %i[edit update make_private make_public manifest]
 
-    before_action only: :show do
-      blacklight_config.show.document_component = Spotlight::DocumentComponent
-    end
-
     before_action only: :admin do
       blacklight_config.view.select! { |k, _v| k == :admin_table }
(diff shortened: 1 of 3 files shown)


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -22,10 +22,6 @@ class CatalogController < ::CatalogController
 
     before_action :load_document, only: %i[edit update make_private make_public manifest]
 
-    before_action only: :show do
```
cbeer, line 25: I suspect we did this so previously generated applications also got the right document component. I don't know what the right backwards-compatible pattern might be 🤷‍♂️

corylown: The current implementation breaks any apps (like our exhibits app) that have additional show partials configured. Is there a reason why a previously generated app needs to use the document component? It appeared to me the if the document component wasn't configured that it would fall back to using the default show partial. My thought was it might be better to let previously generated apps opt-in, but it's possible I'm not understanding something.

cbeer: Nope, I think you've probably got it right. I do think this _was_ a nod at backwards compatibility but only works with Blacklight 8 (with https://github.com/projectblacklight/blacklight/commit/6467c5d3631dfd4d127c73e2e5cdf85e1ad6e03f).

Issue #2992, opened 2023-11-21
No images shown in full record display by default


mephillips-durham:
In a freshly-bootstrapped rails application based on the Spotlight engine, any images, whether static or IIIF, are only shown in the brief results screens and do not appear in the full record display. This is not helpful to new users, who may struggle to work out how to add the images. Most users will want images displayed on the full record view.

By contrast, when you Edit a record, clicking the Edit button from the full record display, the images are shown. It appears this is because the edit screens use an older mechanism. I wonder whether the images disappeared from the default full record display when Blacklight moved to using View Controllers?

I don't know whether there is a very simple fix that I have not found, and I don't know whether it needs addressing in Spotlight or in Blacklight.

### Background investigations
If I turn on diagnostics in the rendering so that the HTML is marked up with the template filenames using
```
config.action_view.annotate_rendered_view_with_filenames = true
```
then in the normal full record display we have these templates in use:

```
/home/vagrant/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/blacklight-7.34.0/app/components/blacklight/document_component.rb
/home/vagrant/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/blacklight-7.34.0/app/components/blacklight/document_metadata_component.rb
/home/vagrant/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/blacklight-7.34.0/app/components/blacklight/metadata_field_component.rb
```

The metadata_field_component would be repeated for each metadata field in the record, displaying each in a new row with the field name on the left. I would expect the document_component.rb to generate or include something to make the image appear. It does make a couple of lines appear with "og" metadata for use by Facebook etc.:

```
<meta property="og:title" content="Example title here" />
<meta property="og:og:image" content="/images/535-8c9b943d3fccdc21ebd2930481a0f3c6/full/!400,400/0/default.jpg" />
```

But there is no <iframe> or <img> tag to make the image appear.

By contrast, if we look at the templates used in the Edit view of the record, we get these:

```
/home/vagrant/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/spotlight-0f5a47eefe43/app/views/spotlight/catalog/edit.html.erb
/home/vagrant/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/blacklight-7.34.0/app/views/catalog/_show_header.html.erb
[...]

PR #3062, opened 2024-07-19
Add test to check that image viewer is rendered on the exhibits show page

corylown:
Prevent a future regression of the issue reported in https://github.com/projectblacklight/spotlight/issues/2992 and fixed in https://github.com/projectblacklight/spotlight/pull/3060

### DIFF
--- a/spec/features/catalog_spec.rb
+++ b/spec/features/catalog_spec.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-describe 'Catalog', type: :feature do
+RSpec.describe 'Catalog', type: :feature do
   let(:exhibit) { FactoryBot.create(:exhibit) }
 
   describe 'admin' do
@@ -25,6 +25,11 @@ describe 'Catalog', type: :feature do
       visit spotlight.search_exhibit_catalog_path(exhibit)
       expect(page).to have_selector '.card-header', text: 'Item visibility'
     end
+
+    it 'has an edit link' do
+      visit spotlight.exhibit_solr_document_path(exhibit, 'dq287tq6352')
+      expect(page).to have_link('Edit')
+    end
   end
 
   it 'has breadcrumbs' do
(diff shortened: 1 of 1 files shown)


spec/features/exhibits_show_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+describe 'Exhibits show page', type: :feature do
```
jcoyne, line 3:
```suggestion
RSpec.describe 'Exhibits show page', type: :feature do
```
Avoid the need for monkeypatching.

jcoyne (changes requested):
It seems to me this is the same scenario as is set up in https://github.com/projectblacklight/spotlight/blob/ec99378f29b42ba4ed742982cc77100054ec8190/spec/features/catalog_spec.rb#L50.  Can they be merged?

corylown:
> . Can they be merged?

Yes, specs are combined now. Thank you.

spec/features/catalog_spec.rb
```diff
@@ -25,6 +25,11 @@
       visit spotlight.search_exhibit_catalog_path(exhibit)
       expect(page).to have_selector '.card-header', text: 'Item visibility'
     end
+
+    it 'has an edit link' do
```
jcoyne, line 29: Feature tests are relatively slow (especially in the setup).  Can we include the edit link assertion in the test above?  I think the feature is "it draws the page" and there are several assertions that support "the page is drawn correctly".

corylown: The two admin tests are hitting different routes so I didn't think it made sense to combine them. The test above is for `search_exhibit_catalog_path` and the Edit link test I added is for `exhibit_solr_document_path`

jcoyne: Oh, okay!

jcoyne (approved):

PR #3074, opened 2024-08-01
[WIP] Replace bootstrap-tagsinput with Tom Select

taylor-steve:
Chasing seemingly unrelated test ghosts locally, let's see what CI says...

bootstrap-tagsinput is unsupported and out of date. Tom Select supports using a text input, has a similar method of loading remote data, and has both Bootstrap 4 and 5 assets. Vendoring the assets for now, but if we change the asset pipeline for Spotlight v4 perhaps we can remove that.


jcoyne:
This seems like a huge addition for a small feature.  Have you also considered adapting https://github.com/sul-dlss/argo/blob/main/app/javascript/controllers/text_extraction_controller.js ?

taylor-steve:
It certainly looks huge. It's ~70 lines, 30 of which are config. The rest are vendor files because we lack a clean way to incorporate those in a different fashion (AFAIK).

I wasn't aware of that code, thanks.

PR #3075, opened 2024-08-02
Add support for Bootstrap 5

corylown:


app/assets/stylesheets/spotlight/_utilities.scss
```diff
@@ -1,7 +1,3 @@
 .inline-block {
   display: inline-block;
 }
-
-.btn-sizing {
-  @include button-size($btn-padding-y, $btn-padding-x, $btn-font-size, $btn-line-height, $btn-border-radius);
-}
```
corylown, line 7: Converted to standard bootstrap `.btn-link` to avoid changes to button-size arguments.

app/assets/stylesheets/spotlight/_blacklight_configuration.scss
```diff
@@ -30,14 +30,6 @@
...
     padding-bottom: 0;
 
-    label {
-      margin-left: $spacer;
-      margin-right: $spacer;
-      padding-left: 0px;
-      padding-bottom: $table-cell-padding;
-      padding-top: 1 + $table-cell-padding;
-      display: block;
-    }
```
corylown, line 40: `$table-cell-padding` is not available in bootstrap 5. This particular label is `visually-hidden`/`sr-only` and didn't seem to impact the display at all so I removed the styling.

taylor-steve:
@corylown I think this is looking great. Here are a few visual differences I found, should you want to address them in this work. I didn't verify that all of them are BS5 related. I'd be happy to move this along if some or all should be addressed individually.

Double X and underline:
Screenshot 2024-08-07 at 8 23 43 AM

Link and checkbox order and alignment differs:
Screenshot 2024-08-07 at 8 32 00 AM

Screenshot 2024-08-07 at 8 27 36 AM

Checkbox alignment differs:
Screenshot 2024-08-07 at 8 29 22 AM

Underlines on tags in item view are new:
Screenshot 2024-08-07 at 8 31 00 AM

corylown:
@taylor-steve thanks for taking a close look at this.

- The double X to dismiss the flash message is fixed in BL 8 so I'd rather wait for us to get to BL8 and get the fix from there: https://github.com/projectblacklight/blacklight/commit/8a9e5507e98e07124b296a15d3dc84e0a0155de4
- I've fixed the checkbox alignment by adding `multiple: true` as an argument to the form -- a totally undocumented feature. I checked and it looks fine in BS4 too.
Screenshot 2024-08-07 at 4 52 05 PM
Screenshot 2024-08-07 at 4 51 56 PM
Screenshot 2024-08-07 at 4 51 47 PM
- I'd like to hold off changing the underlining behavior until we know whether we want to keep the current behavior or let BS5 do what it wants.

jcoyne (approved):

PR #3078, opened 2024-08-06
Add a view component for tagging, replacing bootstrap-tagsinput

### LINKED ISSUE #2890: replace bootstrap-tagsinput dependency
https://github.com/mdbootstrap/bootstrap-templates/blob/2661784c2c281d3a69b93897ff3f39e4ffa5cbd1/package.json

The vendored version is 0.3.9.  We could upgrade https://www.npmjs.com/package/bootstrap-tagsinput

### LINKED ISSUE #2962: Replace dependency on twitter-typeahead rails
Blacklight no longer provides this: https://github.com/projectblacklight/blacklight/pull/2576

taylor-steve:
Closes #2890
Closes #2962

A variation on the code mentioned in #3074, adding the ability to add new tags on the fly in the same fashion as bootstrap-tagsinput.

This would be the last use of bootstrap-tagsinput and twitter typeahead (typeahead.bundle) in Spotlight.


taylor-steve:
Looking for feedback on if my changes (the ability to add a new tag on the fly) to the original, good widget that I adapted ruin it. Sometimes I think it makes it a bit unwieldly, but it is functional.

app/components/spotlight/tag_selector_component.rb
```diff
@@ -0,0 +1,61 @@
...
+        <svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" aria-hidden="true" width="24" height="24" viewBox="0 0 24 24">
+          <path fill="none" d="M0 0h24v24H0V0z"/><path d="M15.5 14h-.79l-.28-.27C15.41 12.59 16 11.11 16 9.5 16 5.91 13.09 3 9.5 3S3 5.91 3 9.5 5.91 16 9.5 16c1.61 0 3.09-.59 4.23-1.57l.27.28v.79l5 4.99L20.49 19l-4.99-5zm-6 0C7.01 14 5 11.99 5 9.5S7.01 5 9.5 5 14 7.01 14 9.5 11.99 14 9.5 14z"/>
+        </svg>
+      SVG
+    end
+
+    private
+
+    def bootstrap_version
+      bootstrap_gem = Gem.loaded_specs['bootstrap']
```
corylown, line 38: In other places in Spotlight we've added the `visually-hidden` class to the close button so it will appear for BS4 and not BS5. Would that be possible here too so we don't have to check the version of the installed gem?

https://github.com/projectblacklight/spotlight/blob/3fd53ca6ddf8bf32f670ef318420a7eb8777b78a/app/views/catalog/_add_tags.html.erb#L8

corylown (commented):
This seems like an okay solution to me if Alan approves of the UI. The UI is...unusual...but in some ways it's clearer than the current implementation. See comment about checking for BS4

jcoyne:
Would it be better to use the tag UI that @dnoneill developed for Argo?

corylown (approved):

Issue #3089, opened 2024-08-19
Bug: Edit page Thumbnail form layout in Bootstrap 5


corylown:
### Description
Spacing between radio button form elements needs adjustment
<img width="1035" alt="edit-page-thumbnail-bs5" src="https://github.com/user-attachments/assets/276e5bde-d6ab-41c6-ba3a-f1e69a8e92d8">

### Steps to reproduce
Go to /spotlight/EXHIBIT/home/edit
Select the "thumbnail" tab

### Expected behavior
<img width="1045" alt="edit-page-thumbnail-bs4" src="https://github.com/user-attachments/assets/e6bb646b-29ec-4be1-939a-d1845cbc24ee">

Issue #3086, opened 2024-08-19
Bug: Exhibit Thumbnail form layout in Bootstrap 5


corylown:
### Description
Not enough space between radio buttons

<img width="795" alt="appearance-exhibit-thumbnail-bs5" src="https://github.com/user-attachments/assets/87733f04-9eb8-49f3-8dae-fbcb7975ac57">

### Steps to reproduce
Go to spotlight/EXHIBIT/appearance/edit
Select the "Exhibit thumbnail" tab

### Expected behavior
<img width="779" alt="appearance-exhibit-thumbnail-bs4" src="https://github.com/user-attachments/assets/81fd4672-716a-482a-b92f-c2a1a0b87b68">

Issue #3085, opened 2024-08-19
Bug: Exhibit Masthead form layout in Bootstrap 5


corylown:
### Description
Not enough space between the radio buttons
Check box should be left aligned with the rest of the form content
<img width="790" alt="appearance-exhibit-masthead-bs5" src="https://github.com/user-attachments/assets/54182b8b-fa32-4a50-8f41-583500ba71e0">

### Steps to reproduce
Go to spotlight/EXHIBIT/appearance/edit
Select the "Exhibit masthead" tab

### Expected behavior
<img width="763" alt="appearance-exhibit-masthead-bs4" src="https://github.com/user-attachments/assets/1e237f4d-80e7-44d9-b89c-98fba440358a">

PR #3093, opened 2024-08-19
Use div instead of small for form-text

corylown:
Fixes: #3089
Fixes: #3086
Fixes: #3085

Fix does not change the BS4 display.

Before:
Screenshot 2024-08-19 at 5 42 06 PM

After:
Screenshot 2024-08-19 at 5 41 47 PM

### DIFF
--- a/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
+++ b/app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
@@ -89,11 +89,6 @@ legend {
   margin-bottom: $spacer;
 }
 
-small.form-text {
-  margin-bottom: $spacer;
-  margin-top: $spacer * 0.3125;
-}
-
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/spotlight/_bootstrap_overrides.scss
```diff
@@ -90,6 +90,7 @@ legend {
 }
 
 small.form-text {
+  display: block;
```
jcoyne, line 93: Wouldn't it be more clear and maintainable  to add the `.d-block` class to the html?

jcoyne (approved):

jcoyne:
@corylown BS5 suggests a `div` for `.form-text`, (rather than `<small>`), in which case the `display: block` would be implicit.

jcoyne (approved):

PR #3094, opened 2024-08-20
Fix translation export button layout for BS5

corylown:
Before:
Screenshot 2024-08-20 at 7 59 15 AM

After:
Screenshot 2024-08-20 at 9 05 09 AM

No impact to Bootstrap 4 layout

### DIFF
--- a/app/views/spotlight/translations/_import.html.erb
+++ b/app/views/spotlight/translations/_import.html.erb
@@ -1,24 +1,37 @@
 <% if can? :import_translations, Spotlight::Exhibit %>
-  <div class="row col-md-12 mb-5">
-    <h2><%= t(:'.header') %></h2>
-    <p class="instructions"><%= t(:'.description') %></p>
-
-    <%= bootstrap_form_for current_exhibit, url: spotlight.import_exhibit_translations_path(current_exhibit), html: { class: 'row col-md-12 align-items-center clearfix mb-3', multipart: true } do |f| %>
-      <%= f.label :value, t(".import_label"), class: 'col-form-label col-12 col-sm-3 pl-0 translation-import-export-label' %>
-
-      <div class="input-group col px-0">
-        <%= file_field_tag :file, class: 'form-control' %>
-        <%= hidden_field_tag :tab, 'import', id: nil %>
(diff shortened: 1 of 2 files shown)


app/views/spotlight/translations/_import.html.erb
```diff
@@ -17,8 +17,8 @@
     <div class="row col-md-12 align-items-center">
```
jcoyne, line 17: Is this because we have the `col-md-12` and the `row` on the same node?  BS styles `.row > *` with `width: 100%`.

jcoyne (changes requested):

app/views/spotlight/translations/_import.html.erb
corylown: The button layout issue is definitely due to `width: 100%` coming from BS5 `.row`. Other solutions to keep the "Export a YAML file" row of buttons and the form above it aligned in both BS4 and BS5 don't seem much better to me. Do you have a suggestion? This works, but I'm not sure it's better than adding `w-auto` to the buttons:

```
    <div class="container-fluid px-0">
      <div class="row">
        <div class="col-form-label col-12 col-md-3 text-md-right text-md-end">
          <%= content_tag :span, t(".export_label"), class: 'px-0' %>
        </div>
        <div class="col-12 col-md-9 px-md-0">
          <%= link_to t(:'.export_current_locale', language: t(:"locales.#{@language}")), spotlight.exhibit_translations_path(current_exhibit, format: "yaml", locale: @language), class: 'btn btn-primary mr-3 me-3' %>
          <%= link_to t(:'.export_default_locale', language: t(:"locales.#{I18n.default_locale}")), spotlight.exhibit_translations_path(current_exhibit, format: "yaml", locale: I18n.default_locale), class: 'btn btn-primary' %>
        </div>
      </div>
    </div>

```

jcoyne: I think that's much more like what the bootstrap developers expect.  https://getbootstrap.com/docs/5.3/forms/layout/#horizontal-form.

jcoyne (approved):

PR #3114, opened 2024-09-11
Export URLify function

taylor-steve:


corylown (changes requested):
@taylor-steve could we add a test for this so we can catch any regressions in the future? I think something like this in `spec/features/create_exhibit_spec.rb` should work:
```
  it 'suggests a slug based on the title', js: true do
    visit spotlight.new_exhibit_path
    fill_in 'Title', with: 'My exhibit title'

    expect(find_field('URL slug')[:placeholder]).to eq 'my-exhibit-title'
  end
```

taylor-steve:
Done. It did work, thank you. I did check a breaking case, which it caught.

PR #3117, opened 2024-09-11
Pass the tile_source_field to blacklight-gallery

marlo-longley:
Blocked by https://github.com/projectblacklight/blacklight-gallery/pull/151

Closes ##3111

We now see the viewer: 
Screenshot 2024-09-11 at 13 47 48

### DIFF
--- a/app/models/spotlight/blacklight_configuration.rb
+++ b/app/models/spotlight/blacklight_configuration.rb
@@ -121,8 +121,12 @@ module Spotlight
         config.default_per_page = default_per_page if default_per_page
 
         config.view.embed!
+        # This is blacklight-gallery's openseadragon partial
         config.view.embed.partials ||= ['openseadragon']
         config.view.embed.if = false
+
+        # blacklight-gallery requires tile_source_field
+        config.view.embed.tile_source_field ||= config.show.tile_source_field
         config.view.embed.locals ||= { osd_container_class: '' }
 
         # Add any custom fields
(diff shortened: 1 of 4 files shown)


taylor-steve:
@marlo-longley I tried this out and I'm not seeing the image on my end. Looking at your screenshot it looks like we're doing the same procedure.

Would it make sense to add a test for this image display? It looks like `spec/features/javascript/blocks/solr_documents_block_spec.rb` covers the majority of the functionality already with `app/javascript/spotlight/admin/blocks/solr_documents_embed_block.js` having only a minor difference. I think the image display would be the only thing we'd need to test?

marlo-longley:
@taylor-steve I added a test. This is ready for review. Thanks.

cbeer:
Do we also need to bump the required blacklight-gallery version?

marlo-longley:
Thanks @cbeer  -- I think all these statements will make sure `4.5.4` gets into the app or test apps --

https://github.com/projectblacklight/spotlight/blob/e546b11a79e1639d3f66f197fe6185c837d31f2f/blacklight-spotlight.gemspec#L27

https://github.com/projectblacklight/spotlight/blob/e546b11a79e1639d3f66f197fe6185c837d31f2f/lib/generators/spotlight/install_generator.rb#L119

https://github.com/projectblacklight/spotlight/blob/e546b11a79e1639d3f66f197fe6185c837d31f2f/spec/test_app_templates/lib/generators/test_app_generator.rb#L14

PR #3123, opened 2024-09-16
update analytics to work with ga-4

### LINKED ISSUE #3050: Determine what is needed to support GA4 (Google Analytics 4)
The previous version of Google Analytics (UA -- Universal Analytics) is no longer supported by Google. We should check whether the existing analytics code in Spotlight needs updating to work with GA4.

https://github.com/projectblacklight/spotlight/tree/main/app/models/spotlight/analytics

dnoneill:
closes #3050


app/models/spotlight/analytics/ga.rb
```diff
@@ -1,80 +1,132 @@
...
       end
 
-      delegate :metrics, to: :model
+      def client
+        Google::Analytics::Data.analytics_data do |config|
+          config.credentials = Spotlight::Engine.config.ga_pkcs12_key_path
+        end
+      rescue StandardError
+        nil
+      end
```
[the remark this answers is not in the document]
jcoyne: I'd also suggest `Honeybadger.notify` as well

app/components/spotlight/analytics/aggregation_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<table class="table analytics">
+  <tr>
+  <% display_fields.each_pair do |key, value| %>
```
corylown, line 3: When there's no data available from GA for an exhibit the app errors here with `undefined method `each_pair' for []:Array`. I think we need to handle the case where there's no data yet for a particular exhibit. I assume the same thing would happen on line 9.

app/components/spotlight/analytics/aggregation_component.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def format_field(key, value)
+        if value.is_a?(Float)
+          if key.to_s.downcase.include?('rate')
+            "#{(value * 100).to_i}%"
+          else
+            # format('%.2f', value) causes an error
+            # rubocop:disable Style/FormatString
+            '%.2f' % value
+            # rubocop:enable Style/FormatString
```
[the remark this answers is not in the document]
dnoneill: Thank you. I kept googling this and could not find the answer.

lib/generators/spotlight/templates/config/initializers/spotlight_initializer.rb
```diff
@@ -76,6 +76,7 @@
 #   Spotlight::Engine.config.ga_email = 'test@example.com'
```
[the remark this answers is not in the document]
dnoneill: I think it can get removed. I updated the comments to be in line with what needs to be done.

app/components/spotlight/analytics/aggregation_component.html.erb
dnoneill: The render? statement I put in here should have fixed that.

app/models/spotlight/analytics/ga_model.rb
```diff
@@ -1,16 +1,10 @@
 # frozen_string_literal: true
 
-require 'legato'
-
 module Spotlight
   module Analytics
     ##
     # Google Analytics data model for the Exhibit dashboard
     class GaModel
```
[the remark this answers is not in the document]
dnoneill: removed, doesn't look like.

app/models/spotlight/analytics/ga.rb
```diff
@@ -1,80 +1,132 @@
...
 module Spotlight
   module Analytics
     ##
     # Google Analytics data provider for the Exhibit dashboard
     class Ga
-      extend Legato::Model
-
       def enabled?
-        user && site
+        Spotlight::Engine.config.ga_pkcs12_key_path && client
```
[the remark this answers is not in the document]
dnoneill: updated to ga_json_key_path.

app/models/spotlight/analytics/ga.rb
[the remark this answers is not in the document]
dnoneill: Is this widely enough used that it makes sense to add to an open source project?

app/models/spotlight/analytics/ga.rb
```diff
@@ -1,80 +1,131 @@
...
-                                   sub: Spotlight::Engine.config.ga_email
+      private
+
+      def ga_web_property_id
+        Spotlight::Engine.config.ga_web_property_id
+      end
+
+      def model
+        Spotlight::Analytics::GaModel
       end
```
corylown, line 129: I think this can be removed now too.
```suggestion
```

dnoneill: Good catch

app/components/spotlight/analytics/dashboard_component.html.erb
```diff
@@ -0,0 +1,51 @@
+<%= cache current_exhibit, expires_in: 1.hour do %>
```
corylown, line 1: This cache key/block is problematic. On the first render of this partial for an exhibit it will cache the whole partial and then display the same content no matter what start_date or end_date is applied in later requests. I think to fix this you'd need to add the start/end param values to the cache key or remove the conditional display for now, which is what I suggest below.

app/components/spotlight/analytics/dashboard_component.html.erb
```diff
@@ -0,0 +1,51 @@
...
+  <% if params[:start_date] || params[:end_date] %>
```
corylown, line 2: I'm thinking we should remove the `params[:start_date]` and `params[:end_date]` functionality from this partial and from `dashboard_component.rb` until we can more completely think through the UI. The caching isn't set up right for this and we haven't thought through the controls or what to display in the UI -- displaying "User activity from 30daysAgo to today" doesn't seem great. I can create a new issue for this to discuss with Cathy and Alan. The rest of the code supports adding these features without too much trouble later.

app/components/spotlight/analytics/aggregation_component.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Spotlight
+  module Analytics
+    # Display Analtyics aggregations as table
+    class AggregationComponent < ViewComponent::Base
```
corylown, line 6: Could you add a test for this component?

app/components/spotlight/analytics/dashboard_component.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    <h2><%= I18n.t("spotlight.dashboards.analytics.monthly_header_selections", start_date: dates['start_date'], end_date: dates['end_date']) %></h2>
+  <% else %>
+    <h2><%= I18n.t("spotlight.dashboards.analytics.monthly_header") %></h2>
```
corylown, line 5: This key name `monthly_header` no longer makes sense now that we're showing a year of analytics data (and may show other ranges in the future). Maybe we should change it to something that doesn't reference a specific time, like `reporting_period_heading`

corylown (changes requested):
@dnoneill I like this much better with the new component. In addition to my comments in the code, I'm wondering if we need to do something better than display a blank page when there's no analytics data available for an exhibit. Maybe just something like "No analytics data available for this exhibit."?
Screenshot 2024-09-18 at 4 27 11 PM

app/components/spotlight/analytics/dashboard_component.html.erb
dnoneill: This is not the behavior I am seeing in the app. It isn't going to be relevant because I am going to remove the start date, but it does reload with the right variables when you add a start date. My assumption was this line makes it work https://github.com/projectblacklight/spotlight/blob/539d806a2d8b90e618ef18e54070b8ea99edb2a3/app/models/concerns/spotlight/exhibit_analytics.rb#L11.

Definitely could be a problem if there is a different end_date and no new start date. Anyway moot for now due to the removal of the start/end.

app/components/spotlight/analytics/aggregation_component.rb
dnoneill: done

app/components/spotlight/analytics/dashboard_component.html.erb
dnoneill: ok

dnoneill:
@corylown Added a view for when there are no results.
Screenshot 2024-09-18 at 5 17 11 PM

PR #3130, opened 2024-09-18
Add ability to restore default labels when editing metadata fields

### LINKED ISSUE #3126: Feature: Add a "restore default" option for each metadata field in the Dashboard > Metadata section
This is based on https://github.com/sul-dlss/exhibits/issues/2486 -- we decided to propose the change upstream.

marlo-longley:
Closes #3126

Because of space constraints, this PR reduces the font size and width of checkbox columns in order to make space for the new button. See discussion in the ticket.

Since this feature adds the "restore default" functionality to the Metadata page in addition to Appearance page, I moved the JS to a shared location. 

Screenshot 2024-09-18 at 16 51 39


app/javascript/spotlight/admin/edit_in_place.js
```diff
@@ -43,5 +43,24 @@ export default class {
         return false;
       });
     })
+
+    $("[data-behavior='restore-default']").each(function(){
+      var hidden = $("[data-default-value]", $(this));
+      var value = $($("[data-in-place-edit-target]", $(this)).data('in-place-edit-target'), $(this));
+      var button = $("[data-restore-default]", $(this));
+      hidden.on('blur', function(){
```
taylor-steve, line 51: @marlo-longley I've been looking into the test failure you and @corylown were asking about. Two things:
1. I think maybe we need to handle more than blur, depending on the behavior we are looking for.

* If I click "Language", type "Language of Origin" and press Enter, I do not get the "Restore default" button in the current page
*  If I click "Language", type "Language of Origin" and invoke blur by clicking somewhere else (including the save changes button) I immediately get the "Restore default" button in the current page


2. Right now, there is an immediate check for that restore default button as an indication that the save has occurred. I don't think that's going to be reliable as we don't know for sure if that find will happen before or after the page reload.
https://github.com/projectblacklight/spotlight/blob/6fbece146028df810248766d7d342a2f180201ec/spec/features/exhibits/edit_metadata_fields_spec.rb#L41-L43

If we want the "Restore button" to show immediately, could we cheese it a bit for the test and force a page revisit? Add sleep before the config data check?

marlo-longley: Group decision: skip this test.

corylown (changes requested):

app/views/spotlight/metadata_configurations/_metadata_field.html.erb
```diff
@@ -1,12 +1,22 @@
-<tr data-id="<%= key.parameterize %>" class="dd-item">
+<% default_field_label = local_assigns.dig(:config, :original, :label) %>
+
+<tr data-id="<%= key.parameterize %>" class="dd-item" data-behavior="restore-default">
   <%= f.fields_for key do |field| %>
     <td>
       <%= field.hidden_field :weight, 'data-property' => 'weight' %>
       <div class="handle-wrap" data-in-place-edit-target=".edit-in-place" data-in-place-edit-field-target="[data-edit-field-target='true']">
```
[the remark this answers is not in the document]
marlo-longley: This is the current behavior. But I can address it if desired.

corylown (approved):

Issue #3047, opened 2024-07-08
Support Blacklight 8


corylown:
This may have mostly been completed via https://github.com/projectblacklight/spotlight/pull/2853

Relates to:

- https://github.com/projectblacklight/spotlight/issues/3025
- https://github.com/projectblacklight/spotlight/issues/2950
- https://github.com/projectblacklight/spotlight/pull/2949
- https://github.com/projectblacklight/spotlight/pull/3022

[Upgrading to Blacklight 8 Guide](https://github.com/projectblacklight/blacklight/wiki/Upgrading-to-Blacklight-8)

- [ ] #3069
- [x] https://github.com/projectblacklight/spotlight/issues/2966
- [ ] #3070
- [ ] https://github.com/projectblacklight/spotlight/pull/2922
- [ ] Convert admin partials to components? (not sure if this is strictly required): https://github.com/search?q=repo%3Aprojectblacklight%2Fspotlight+blacklight_config.view.admin_table&type=code
- [ ] Figure out if we need to adapt how we're handling constraints: see https://github.com/projectblacklight/blacklight/wiki/Upgrading-to-Blacklight-8#adding-elements-to-constraints and also: https://github.com/search?q=repo%3Aprojectblacklight%2Fspotlight+render_constraints&type=code

Things we can drop if we stop supporting Blacklight 7:
- [ ] `Blacklight::SearchService#fetch` returns just the documents, not the solr response and the documents: https://github.com/search?q=repo%3Aprojectblacklight%2Fspotlight+search_service.fetch&type=code
- [ ]  `convert_to_search_state` has been removed. Just use `search_state`: https://github.com/search?q=repo%3Aprojectblacklight%2Fspotlight+convert_to_search_state&type=code

Update: Look at this comment for current state: https://github.com/projectblacklight/spotlight/issues/3047#issuecomment-2274034424

corylown:
Known issues with Blacklight 8 Spotlight:
- `app/views/layouts/spotlight/base.html.erb` and `app/views/shared/_header_navbar.html.erb` both call Blacklight's `container_classes` helper that in Blacklight 8 calls `blacklight_config`. Seeing some test failures complaining that `blacklight_config` is not defined.
- The only way I've been able to get spotlight to work with Blacklight 8 is to generate spotlight with Blacklight 7 and then switch to version 8 via the gemfile, likely due to different default settings for asset handling in BL7 vs 8.
- Had to add to `app/assets/config/manifest.js`
```
[...]

PR #3131, opened 2024-09-18
ensure string is passed to truncate since errors with integers

hudajkhan:
Relates to having tests pass with Blacklight 8 (specially 8.3).

For 8.3, the following test is failing:
rspec ./spec/views/spotlight/sir_trevor/blocks/_solr_documents_features_block.html.erb_spec.rb:48 # spotlight/sir_trevor/blocks/_solr_documents_features_block.html.erb without a primary caption falls back to the regular document title for the caption

The reason appears to be when the truncate method is passed an integer, e.g. truncate(1, length:95), it throws an error.  However, truncate("1", length:95) will work just fine. Adding a "to_s" ensures the caption text truncation does not throw an error. 

This test should still pass in Blacklight 7. 

How to test with Blacklight 8: Generate test app (.internal_test_app). Update the internal test app's Gemfile to use Blacklight 8.3.  Run bundle install for the internal test app. Then run rspec ./spec/views/spotlight/sir_trevor/blocks/_solr_documents_features_block.html.erb_spec.rb:48.

Relates to #3047 and this comment which includes a checklist of tests to fix: https://github.com/projectblacklight/spotlight/issues/3047#issuecomment-2359456286

### DIFF
--- a/app/views/spotlight/sir_trevor/blocks/_solr_documents_features_block.html.erb
+++ b/app/views/spotlight/sir_trevor/blocks/_solr_documents_features_block.html.erb
@@ -25,7 +25,7 @@
                   <% caption_text = solr_documents_features_block.document_caption(doc_presenter, solr_documents_features_block.primary_caption_field, default: Spotlight::PageConfigurations::DOCUMENT_TITLE_KEY) %>
 
                   <a href="#" title="<%= caption_text %>">
-                    <%= truncate(caption_text, length: 95) %>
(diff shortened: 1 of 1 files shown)


jcoyne:
What causes the caption text to be an integer in the first place?

cbeer:
This might be an artifact of the test setup (falling back on an integer `id` field), but it's plausible (if not likely) that this could happen in reality.

PR #3133, opened 2024-09-19
add property id for ga dashboard

dnoneill:


lib/generators/spotlight/templates/config/initializers/spotlight_initializer.rb
```diff
@@ -68,11 +68,12 @@
 # a) Enable Google Analytics API in https://console.cloud.google.com/
 # b) generate and download the JSON key and make it accessible to your application
 # (https://console.cloud.google.com/iam-admin/iam -> Service accounts -> click on service account -> keys)
-# c) set ga_web_property_id below to your site's property id
+# c) set ga_property_id below to your site's property id (located in admin -> Property -> Property details upper right hand corner)
 # Rails.application.config.to_prepare do
 #   Spotlight::Engine.config.analytics_provider = Spotlight::Analytics::Ga
 #   Spotlight::Engine.config.ga_json_key_path = nil
 #   Spotlight::Engine.config.ga_web_property_id = 'G-XXXXXXXXXX'
+#   Spotlight::Engine.config.ga_property_id = '12345678'
```
corylown, line 76: I could see `ga_web_property_id` and `ga_property_id` causing confusion.

What we're calling `ga_web_property_id` appears to be called "measurement id" by Google.

I think we should clarify in the comments that `ga_web_property_id` is called "measurement id" by Google and is used to identify where to send analytics data. You can find it through admin -> Data collection and modification -> Web stream details and begins with `G-`

It might also be helpful to add to the info about `ga_property_id` that it's used to identify the property when fetching analytics from the API.

corylown: Renaming `ga_web_property_id` to `ga_measurement_id` might also be helpful, but then we have a backwards compatibility issue to deal with. Maybe we save that for v5.

corylown (changes requested):

corylown (approved):

PR #3160, opened 2024-09-25
Add a default document component for the admin table + blacklight 7.

cbeer:
It's possible downstream application:

- render the document admin table partial outside the `catalog#admin` action, or
- customized the `admin_table` view but haven't updated it to use view components yet.

A little more defensive programming should help avoid breaking things 🤷‍♂️

### DIFF
--- a/app/views/spotlight/catalog/_document_admin_table.html.erb
+++ b/app/views/spotlight/catalog/_document_admin_table.html.erb
@@ -11,7 +11,7 @@
   </thead>
 
   <% if Blacklight.version < '8.0' %>
-    <%= render view_config.document_component.with_collection(documents) %>
+    <%= render (view_config.document_component || Spotlight::DocumentAdminTableComponent).with_collection(documents) %>
   <% else %>
     <% document_presenters = documents.map { |doc| document_presenter(doc) } -%>
     <%= render view_config.document_component.with_collection(document_presenters) %>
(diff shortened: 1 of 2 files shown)


app/views/spotlight/catalog/_document_admin_table.html.erb
```diff
@@ -11,7 +11,7 @@
   </thead>
 
   <% if Blacklight.version < '8.0' %>
-    <%= render view_config.document_component.with_collection(documents) %>
+    <%= render (view_config.document_component || Spotlight::DocumentAdminTableComponent).with_collection(documents) %>
```
jcoyne, line 14: I'm curious as to why a fallback is needed here since you set the component in CatalogController.

cbeer: Besides what I put in the PR description?

We had a pre-existing customization in exhibits that broke the admin page without some other change (e.g. https://github.com/sul-dlss/exhibits/pull/2574/commits/7666bf83b52c1c283515226de43920dcb9d5b9d6), and that's something we should try to avoid doing to downstream applications?

jcoyne: Yeah, I'm still not understanding how `view_config.document_component` could be nil. since it was set in CatalogController in this PR.   Wouldn't that change alone have prevented the problem we had in Exhibits?

cbeer: If the partial is rendered outside of the `catalog#admin` action, the contents of the `before_action` block aren't evaluated.

jcoyne: Okay. My assumption was that no one would ever do that.

PR #3161, opened 2024-09-27
Move to Propshaft, stop supporting Sprockets and Blacklight 7

taylor-steve:
This work:
* Provides install generators for bundling (via `jsbundling-rails` with `esbuild`) and non-bundling (via `importmap-rails`) approaches
* Drops explicit support for Sprockets
* Drops explicit support for Blacklight 7
* Defaults to Bootstrap 5 rather than 4
* Replaces uses of `rails-ujs`

Usage:
The README has instructions for template installs. The test app generator defaults to using importmap:
```
rake spotlight:server
```
It can also generate an app that uses bundling:
```
ENGINE_CART_RAILS_OPTIONS="-j esbuild" rake spotlight:server
```

Work for post-merge:
* Script tags for nestable, bootstrap-tagsinput, and typeahead can be removed after open PRs to replace those dependencies are merged
* Script tags for openseadragon-rails can be removed after updates are done to that gem.
* This work introduces additional `sleep` calls in the specs. These should be re-evaluated after the javascript dependencies listed above are merged. The existing typeahead helpers in particular are a primary source of test inconsistency.


dnoneill:
Couple of issues

- right now spotlight:server doesn't work
- CSS not loading or something, the pages don't look right.
-

lib/generators/spotlight/assets/propshaft_generator.rb
```diff
@@ -45,7 +45,14 @@ def add_bootstrap
       def install_gems
         gem 'jsbundling-rails'
         gem 'cssbundling-rails'
-        run 'bundle install'
+      end
+
+      def bundle_install
```
jcoyne, line 50: Is this required?  Generators should always be run in the context of the app.

jcoyne:
It looks to me like this breaks sprockets bundling.  So will this be a major version update?

lib/generators/spotlight/assets/propshaft_generator.rb
taylor-steve: Sort of. I've been searching for a better solution for this. We were loosely copying https://github.com/geoblacklight/geoblacklight/blob/36fa3977309891a026baf43516bff7bc7f4a6803/lib/generators/geoblacklight/assets/vite_generator.rb#L26

The issue is that we're adding cssbundling-rails/jsbundling-rails in the generator. They might not have them installed and I've been on the fence if it's correct to add them to the gemspec, because they aren't necessarily dependencies in a non-propshaft configuration.

taylor-steve:
> It looks to me like this breaks sprockets bundling. So will this be a major version update?

That's the current plan as I understand it.

corylown (approved):

PR #3163, opened 2024-09-27
Configure embed components for testing.

### LINKED ISSUE #3430: Bug: Stop using show.partials.insert
### Description
https://github.com/projectblacklight/spotlight/blob/c7e6517421f7537168d12607b7fe5549dd48b588/spec/test_app_templates/catalog_controller.rb#L16

Blackight 9 no longer has `show.partials` as part of it's configuration.

cbeer:
Fixes #3430

### DIFF
--- a/app/models/spotlight/blacklight_configuration.rb
+++ b/app/models/spotlight/blacklight_configuration.rb
@@ -123,7 +123,7 @@ module Spotlight
         config.view.embed!
         # This is blacklight-gallery's openseadragon partial
         unless config.view.embed.document_component
-          config.view.embed.partials ||= ['openseadragon']
+          config.view.embed.partials ||= ['openseadragon'] unless config.view.embed.embed_component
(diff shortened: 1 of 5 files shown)


app/models/spotlight/blacklight_configuration.rb
```diff
@@ -123,7 +123,7 @@ def blacklight_config
         config.view.embed!
         # This is blacklight-gallery's openseadragon partial
         unless config.view.embed.document_component
-          config.view.embed.partials ||= ['openseadragon']
+          config.view.embed.partials ||= ['openseadragon'] unless config.view.embed.embed_component
```
jcoyne, line 126: Do we want to put something in the template that shows the usage of `embed_component`?

cbeer: No. Some future version of the `blacklight_gallery:install` generator will do that.

hackartisan:
this looks like it was resolved by #3588

PR #3165, opened 2024-10-01
ga similar named exhibit collision fix

dnoneill:
closes https://github.com/sul-dlss/exhibits/issues/2592

### DIFF
--- a/app/models/spotlight/analytics/ga.rb
+++ b/app/models/spotlight/analytics/ga.rb
@@ -37,8 +37,8 @@ module Spotlight
           filter: Google::Analytics::Data::V1beta::Filter.new(
             field_name: 'pagePath',
             string_filter: Google::Analytics::Data::V1beta::Filter::StringFilter.new(
-              match_type: :BEGINS_WITH,
-              value: path.to_s
+              match_type: :PARTIAL_REGEXP,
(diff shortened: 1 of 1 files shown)


corylown (changes requested):
Is there a way to do this that doesn't also remove the exhibit home page from the list of most popular pages?

Exhibits prod (first page listed is the home page):
Screenshot 2024-10-02 at 1 44 02 PM

This PR (home page is missing
Screenshot 2024-10-02 at 1 44 12 PM
 from the list):

app/models/spotlight/analytics/ga.rb
```diff
@@ -37,8 +37,8 @@ def dimension_filter(path)
           filter: Google::Analytics::Data::V1beta::Filter.new(
             field_name: 'pagePath',
             string_filter: Google::Analytics::Data::V1beta::Filter::StringFilter.new(
-              match_type: :BEGINS_WITH,
-              value: path.to_s
+              match_type: :PARTIAL_REGEXP,
```
dnoneill, line 40: FULL_REGEXP works for data and martin-wong as well. Went back and forth about this and decided partial would probably be safer.

corylown (approved):

PR #3169, opened 2024-10-04
add controlled vocab functionality for tag_list

dnoneill:
closes https://github.com/sul-dlss/exhibits/issues/2492

screenshot is for when ```Spotlight::Engine.config.site_tags = ['tag 1', 'tag 2', 'tag 3', 'tag 4', 'tag 5', 'tag 6', 'tag 7', 'tag 8', 'tag 9', 'tag 10', 'tag 11', 'tag 12']```

Screenshot 2024-10-04 at 1 03 41 PM
Screenshot 2024-10-04 at 1 03 17 PM

### DIFF
--- a/app/components/spotlight/tag_list_form_component.html.erb
+++ b/app/components/spotlight/tag_list_form_component.html.erb
@@ -1,0 +1,14 @@
+<% if Spotlight::Engine.config.site_tags %>
+  <div class="form-group row">
+  <label class="col-form-label col-md-2" for="tag_list">Tag list</label>
+  <div class="col-md-10">
+    <div class="overflow-scroll rounded border px-3 py-2 h-25" style="overflow: scroll; height: 25vh!important;">
+      <% Spotlight::Engine.config.site_tags.each do |tag| %>
+        <%= form.check_box :tag_list, { multiple: true, label: tag }, tag , nil %>
+      <% end %>
+    </div>
+  </div>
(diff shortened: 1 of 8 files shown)


app/components/tag_list_form_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<% if Spotlight::Engine.config.site_tags %>
+  <div class="form-group row">
+  <label class="col-form-label col-md-2" for="tag_list">Tag list</label>
+  <div class="col-md-10">
+    <div class="overflow-scroll rounded border px-3 py-2 h-25" style="overflow: scroll; height: 25vh!important;">
```
dnoneill, line 5: This inline style can get removed as soon as we update to bootstrap 5. I can move this to a css file but it didn't seem to make sense due to the upgrade we are doing soon.

app/components/tag_list_form_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Create tag list for exhibit general/create form
+class TagListFormComponent < ViewComponent::Base
```
corylown, line 4: component should be namespaced in `module Spotlight` and moved to `app/components/spotlight`

corylown (changes requested):

corylown (approved):

PR #3170, opened 2024-10-04
add determine_layout to specify spotlight layout

dnoneill:


app/controllers/concerns/spotlight/controller.rb
```diff
@@ -111,5 +111,9 @@ def exhibit_search_facet_path(*args, **kwargs)
       options = Blacklight::Parameters.sanitize(params.to_unsafe_h.with_indifferent_access).merge(options).merge(kwargs).except(:exhibit_id, :only_path)
       spotlight.facet_exhibit_catalog_url(current_exhibit, *args, **options&.symbolize_keys)
     end
+
+    def determine_layout
```
jcoyne, line 115: Was the intention to use this method somewhere? Why are we adding this?

jcoyne, line 115: Can you add a comment here that this is overriding Blacklight's default behavior?

jcoyne: Doesn't https://github.com/projectblacklight/spotlight/blob/main/app/controllers/spotlight/concerns/application_controller.rb#L14 do the same thing?

dnoneill: Not on the login page. It keeps defaulting to the blacklight template.

jcoyne: Is there an issue this is linked to?  The login page is just the devise default, so I'm not sure we should care to add styles to that. Does anyone actually uses that page in a production deployment?

dnoneill: It is a draft ticket. https://github.com/orgs/sul-dlss/projects/67/views/2?pane=issue&itemId=82115217.

jcoyne: @dnoneill this is an community project, so a draft ticket on a DLSS project board isn't sufficient.

jcoyne: @dnoneill if it's intention is to go into the codebase at some point, it's important to log an issue so that the community can understand this.  I don't understand how this is related to propshaft.

dnoneill: We were under the impression it was a regression from current spotlight implementation.

jcoyne: It's not a regression. It's the same behavior as every Blacklight application.

PR #3171, opened 2024-10-04
refactoring navbar as component with slot for prepend component

### LINKED ISSUE #3167: Convert _exhibit_navbar.html.erb to a component with slots for adding content before and after the search bar
This will let us add features like "Search tips" without having to override the whole partial.

hudajkhan:
Closes #3167 .

What this pull request does:

- Capture the logic previously in the views/shared/_exhibit_navbar partial into a component (Spotlight::ExhibitNavbarComponent), providing a slot for content to prepend in the view.  The slot will allow implementation for search tips in Stanford's Spotlight implementation. 
- Replicate the view based tests in the component test, updating specific view configuration/context information.  
- Add a test that checks if overriding the prepend_to_search method will display that content in the page. 
- Remove the view based test from spec/views/shared. 
- The _exhibit_navbar partial now renders the component.  The code checks for catalog controller configuration for the Exhibit Navbar Component to display.  If none is available, Spotlight::ExhibitNavbarComponent  is used.  
- A deprecation method has also been added in order to support the use of a component within blacklight configuration.


app/views/shared/_exhibit_navbar.html.erb
```diff
@@ -1,24 +1,2 @@
...
-          url: search_action_url,
-          advanced_search_url: search_action_url(action: 'advanced_search'),
-          params: search_state.params_for_search.except(:qt),
-          autocomplete_path: suggest_index_catalog_path
-        ) %>
-      </div>
-    <% end %>
-  </div>
-</div>
+<%= render (blacklight_config&.view_config(document_index_view_type)&.exhibit_navbar_component || Spotlight::ExhibitNavbarComponent).new %>
```
dnoneill, line 1: I think this should probably be defined in the controller somewhere. i.e. in the catalogcontroller you should define

config.index.exhibit_navbar_component = Spotlight::ExhibitNavbarComponent

hudajkhan: I think I've addresses both comments @dnoneill

app/views/shared/_exhibit_navbar.html.erb
```diff
@@ -1,24 +1,2 @@
...
+<% Spotlight.deprecator.warn("_exhibit_navbar.html.erb will be deprecated in future versions. Please use the catalog controller's exhibit_navbar_component setting.") %>
```
corylown, line 2:
```suggestion
<% Spotlight.deprecator.warn("_exhibit_navbar.html.erb will be removed; customize the exhibit navbar using components instead") %>
```

corylown (changes requested):

lib/generators/spotlight/templates/catalog_controller.rb
```diff
@@ -16,6 +16,7 @@ class CatalogController < ApplicationController
     # Blacklight 8 sets a default value to 'advanced'
     config.json_solr_path = nil
     config.header_component = Spotlight::HeaderComponent
+    config.exhibit_navbar_component = Spotlight::ExhibitNavbarComponent
```
dnoneill, line 19: Think this is config.index?

app/views/shared/_exhibit_navbar.html.erb
dnoneill: Sorry when I meant update the comment I mean to link or point in some way to https://github.com/projectblacklight/spotlight/pull/3171/files#diff-6052daee924920472438cf797a49bcfdafd3787945da870c8b24e9dac6bf27e4R19

lib/generators/spotlight/templates/catalog_controller.rb
hudajkhan: yes! thanks.

corylown: I guess the question is, does it make sense to configure a different navbar depending on whether you're in the index or show view, or should this be configurable at the app level like the header component?

dnoneill: I don't think it matters, but either this or this https://github.com/projectblacklight/spotlight/pull/3171/files#diff-de18dc9392ca38402d097f5d638e554a17c8b0ba6b2e341fe46ca6188e05cf1cR9 line need to be updated to be inline with each other.

dnoneill (approved):

corylown (approved):

PR #3172, opened 2024-10-04
Address propshaft related CI & test failures

taylor-steve:


.github/workflows/ruby.yml
```diff
@@ -20,20 +20,19 @@ jobs:
...
-            blacklight_version: "~> 7.34"
-            bootstrap_version: ~> 4.0
+            blacklight_version: "~> 7.38"
+            bootstrap_version: "~> 4.0"
             additional_name: Rails 7.1
           - rails_version: "~> 7.2"
             ruby: "3.3"
-            blacklight_version: "~> 8.0"
-            bootstrap_version: ~> 4.0
+            blacklight_version: "~> 8.4.0"
```
corylown, line 33: What's the reason for pinning this to 8.4 and not allowing whatever the latest version of 8 is?

taylor-steve: Matching a known release that has a corresponding NPM frontend release. If there's a mismatch and CI is prompted to pick, it fails. I'm not sure yet what the best solution is here. `yarn info` and pick the closest? Let it fail?

taylor-steve: I think we might be fine on 8 though, testing now and I will change it back if that's the case.

corylown: Thanks for explaining. Makes sense and underscores how annoying it is dealing with rubygem & npm versions...

corylown (approved):

PR #3179, opened 2024-10-08
Edit stimulus controller path in generated test app

marlo-longley:
Closes https://github.com/orgs/sul-dlss/projects/67/views/1?pane=issue&itemId=82115172


corylown:
This error is no longer occurring on https://github.com/projectblacklight/spotlight/pull/3161

PR #3181, opened 2024-10-09
Adjust versions and arguments so tests pass

taylor-steve:
The goal of this PR is to get the tests passing, so we have a real baseline. This way we can get CI feedback if we break things.

* Switches from `~` to `^` for bootstrap.
`yarn build:css`/`rake assets:precompile` was failing in some cases because with the current setup the propshaft generator was literally installing bootstrap 4.0.

* Has a temporary (?) fallback mechanism so that in cases like BL 7.39.0 that do not have a corresponding NPM package it will try and install the latest version for that major release.

* [Temporary Workaround] Somewhere along the way we (probably me!) broke something related to Bootstrap 4 and bundling. When imported in application.js and bundled via esbuild, we were losing functions like `modal` in files like `zpr_links.js`. Long story short our lives would be much easier if we can drop either `jquery-serializejson` or Bootstrap 4, but I will attempt to review the history and find out where the bundling broke when I return. For now, as a stopgap, I'm including the Bootstrap 4 bundle package in the layout when BS4 is installed.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -13,8 +13,8 @@ jobs:
       matrix:
         rails_version: ["~> 7.2"]
         ruby: ["3.2", "3.3"]
-        bootstrap_version: ["~> 4.0"]
-        blacklight_version: ["~> 7.34"]
+        bootstrap_version: ["~> 4.3"]
+        blacklight_version: ["~> 7.38"]
         additional_engine_cart_rails_options: [""]
         additional_name: [""]
         include:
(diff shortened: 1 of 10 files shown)


jcoyne:
@taylor-steve Can these changes be split into separate PRs?  These three changes seem like they could stand on their own.  I don't see how the Dockerfile changes relate to the tests passing.

taylor-steve:
> @taylor-steve can the changes not relating to propshaft (like .ruby.yml) be made as separate PRs on main? Then you can merge those commits into the propshaft branch once they are accepted. This will cause the least departure from the main branch and make it much easier for people to review the propshaft changes.

Yes, that makes sense.

> Can you explain
> 
> > This is our current workaround for preventing importmap:install from wanting to overwrite our application.js.
> 
> Importmap:install will happen well before application.js is written, right?

The behavior I see is that `importmap:install` occurs after the Spotlight template.rb steps. It's the other way around during test app/engine cart generation. The Dockerfile uses the template to build. We spent a little time trying to find a better way, but for now this seemed to suit our immediate need.

jcoyne:
@taylor-steve how about using `after_bundle` instead? https://github.com/projectblacklight/spotlight/pull/3192

corylown (approved):

PR #3182, opened 2024-10-09
use actioncable to track indexing progress

### LINKED ISSUE #3049: Use ActionCable to report indexing progress
Related PR here: https://github.com/projectblacklight/spotlight/pull/1687

dnoneill:
closes #3049


app/components/spotlight/progress_bar_component.html.erb
```diff
@@ -0,0 +1,30 @@
+<div data-behavior='progress-panel' data-exhibit-id="<%= exhibit.id %>">
+  <div class="card index-status border-info mb-3">
+    <h3 class="card-header bg-info text-white h4"><%= t("#{translation_field}.heading") %></h3>
```
jcoyne, line 3: Instead of doing string interpolation with `translation_field` could you instead do `t('heading', scope: translation_field)`?

app/components/spotlight/progress_bar_component.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <p data-behavior='monitor-completed'>
+          <span class="text-muted" data-behavior='date'></span> <%= t("#{translation_field}.completed_html") %>
+        </p>
+      <% end %>
+
+      <p class="bg-warning" data-behavior='monitor-error' style='display:none;'>
+        <%= t("#{translation_field}.error") %>
+      </p>
+
+      <div class="progress">
```
jcoyne, line 25: Could use use the `<progress>` element? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

app/jobs/spotlight/process_bulk_updates_csv_job.rb
```diff
@@ -13,15 +13,25 @@ def perform(exhibit, bulk_update)
...
       csv_path = bulk_update.file.current_path
+      started_at = Time.zone.now
+
       File.open(csv_path) do |f|
         progress&.total = f.each_line.count(&:present?) - 1 # ignore the header
 
         ::CSV.table(f, header_converters: header_converter).each do |row|
           process_row(exhibit, row)
           progress&.increment
+          ActionCable.server.broadcast 'progress_channel',
```
jcoyne, line 24: Should this be a different channel from the one the indexing uses?

package.json
```diff
@@ -29,5 +30,7 @@
     "not IE 11"
   ],
   "dependencies": {
+    "@rails/actioncable": "^7.2.100",
```
jcoyne, line 33: Won't this depend on the rails version the host app is using?

rollup.config.js
```diff
@@ -24,7 +25,7 @@ const rollupConfig = {
     name: ESM ? undefined : 'Spotlight'
   },
   external,
-  plugins: [includePaths(includePathOptions)]
-};
+  plugins: [includePaths(includePathOptions), nodeResolve()]
```
jcoyne, line 28: What is this plugin for?

app/javascript/spotlight/channels/index.js
```diff
@@ -0,0 +1 @@
+// Import all the channels to be used by Action Cable
```
jcoyne, line 1: Is this file needed?

jcoyne (changes requested):

rollup.config.js
dnoneill: Wouldn't work without it.

app/javascript/spotlight/channels/index.js
dnoneill: It is automatically created by ActionCable. I wasn't sure if it was something we would want to keep.

app/jobs/spotlight/process_bulk_updates_csv_job.rb
dnoneill: They use the same JS file. I would need to update a bunch of JS.

app/javascript/spotlight/channels/index.js
jcoyne: Automatically created into an engine? How does that work?

rollup.config.js
jcoyne: What was the nature of the failure?

app/jobs/spotlight/process_bulk_updates_csv_job.rb
jcoyne: To be clear, I'm not saying that it should be. I'm just asking whether you considered it and why you decided that one channel is the best approach.

jcoyne:
Does this impose a new Redis dependency on everyone who wants to run Spotlight?

dnoneill:
@jcoyne Yeah but the way I understood it, it was already a dependency of spotlight?

app/javascript/spotlight/channels/index.js
dnoneill: In a site. I create the site ran the auto creation task and moved everything.

dnoneill:
```
(!) Unresolved dependencies
https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
@rails/actioncable (imported by "app/javascript/spotlight/channels/consumer.js")
created app/assets/javascripts/spotlight/spotlight.esm.js in 184ms
```

app/jobs/spotlight/process_bulk_updates_csv_job.rb
dnoneill: I considered it. Seemed like I would just be adding another listener for a different channel that would do the exact same thing. 

https://github.com/projectblacklight/spotlight/pull/3182/files#diff-a9fcc060ea558da08359b8855bc3feb142bf3f85f9ab5039fa55e82739a310c6R12-R18

package.json
dnoneill: You are suggesting I update the version to be a range?

jcoyne:
@dnoneill redis is not a dependency of spotlight as far as I know.

app/components/spotlight/progress_bar_component.html.erb
dnoneill: No. This is a bootstrap class: https://getbootstrap.com/docs/4.0/components/progress/. Changing this to <progress class="progress"> it breaks the CSS.

jcoyne: Isn't the semantic element better for accessibility than a bootstrap component?

dnoneill: Probably. I think if we are trying to solve the accessibility issue then we should open a new ticket. This is the progress bar that was used in app/views/spotlight/catalog/_reindex_progress_panel.html.erb and app/views/spotlight/bulk_updates/_progress_panel.html.erb

package.json
jcoyne: We want to allow rails 7.1 and rails 8.0 (eventually), right? I think some kind of range would be preferable.

app/javascript/spotlight/channels/consumer.js
```diff
@@ -0,0 +1,6 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
+// You can generate new channels where WebSocket features live using the `bin/rails generate channel` command.
+
+import { createConsumer } from "@rails/actioncable"
```
jcoyne, line 4: I don't think this file belongs in the engine. It should be in the host application.  By doing that you can remove the `@rollup/plugin-node-resolve`

jcoyne (changes requested):

app/javascript/spotlight/channels/consumer.js
dnoneill: You are suggesting this goes in the install generator?

hackartisan:
@dnoneill could this please link to an issue that says what / why

app/javascript/spotlight/channels/index.js
jcoyne: You don't want to move these files into the engine.  They should be generated by the auto creation task in the host app.

app/javascript/spotlight/channels/consumer.js
jcoyne: Isn't it already in the actioncabel generator?

jcoyne: Yes, that is the path I would use.  Yes, we would need instructions for how to install this.  It should include something like "Install and configure Redis and actioncable" (assuming we're not doing Solid + Rails 8).

dnoneill:
moving into draft. Got 90% of the way to moving all the actioncable stuff into the engine. Unfortunately we couldn't figure out how to import channels/consumer into sprockets without it erroring.

It also checks for actioncable, if not falls back to previous behavior.

hackartisan:
agreed to close on PR review ensemble

Issue #3176, opened 2024-10-08
Maintenance: Upgrade to rollup v4


taylor-steve:
We should consider moving to Rollup v4 if we continue to use Rollup.

PR #3184, opened 2024-10-10
Upgrade to rollup 4

marlo-longley:
Closes https://github.com/projectblacklight/spotlight/issues/3176


taylor-steve:
@marlo-longley could you please rebase this whenever you get a chance? Also, I'm slightly confused why the bundled output would change? When you rebase could you confirm that your local repo doesn't have any modified JS? It might simply be differences in Rollup or we forgot to rebundle at some point? Not a big deal, just a mystery to me. Thanks!

marlo-longley:
This is now rebased. 

To answer the question from @taylor-steve, it looks like yes there was 1 minor change to the bundled JS after the reabse. I also tried this on a fresh branch based off main which resulted in the same diff. I don't know the exact reason behind this change in the delimiter logic but it must be something internal to the new version of Rollup, which in turn triggered a slightly different JS bundle. If you expand the file diffs in this PR for`spotlight.js` and `spotlight.esm.js`, you will see that is the only change.

Screenshot 2024-10-24 at 9 22 36 AM

corylown (approved):

taylor-steve:
Thanks for resolving the mystery @marlo-longley

PR #3198, opened 2024-10-16
Restore the Blacklight logo for generated apps

taylor-steve:
The Blacklight default logo-image is currently using the sprockets image_url helper. This overrides logo-image using the propshaft url helper.

This is the same convention Geoblacklight is using.


jcoyne:
@taylor-steve is it possible to fix this in Blacklight instead?

taylor-steve:
@jcoyne I haven't been able to think of a good method yet. Two separate SASS entry points could work? E.g., `blacklight_sprockets.scss`, `blacklight_propshaft.scss`. As far as I could tell this is the only use of `image_url` in Blacklight. I think changing the logo would be a very common first step for a new site, so this didn't feel like much of a workaround, more of an example.

jcoyne:
@taylor-steve I don't think that blacklight really wants to continue supporting Sprockets primarily, so I think we could accept this change upstream.  What about doing `url(https://github.com/projectblacklight/blacklight/blob/main/app/assets/images/blacklight/logo.png?raw=true)`?   Would that work for both sprockets and propshaft?

taylor-steve:
@jcoyne yes, that definitely works with propshaft and I'd expect it to work with sprockets. I don't love that it would separate the logo from the release versions. Although maybe that's better for a logo anyway, defaulting to using the most current branding? I'll think on it a bit more, but you're probably right that it's the best solution.

hackartisan:
ensemble PR review: agreed to close

PR #3202, opened 2024-10-17
Preventing exhibit navb deprecation messages when not needed

### LINKED ISSUE #3200: Bug: Deprecation warning being thrown even when exhibit navbar component is configured
### Description
https://github.com/projectblacklight/spotlight/blob/main/app/views/shared/_exhibit_navbar.html.erb always throws deprecation errors, even when the exhibit navbar component has been configured. 

### Steps to reproduce
When running any tests that trigger the use of this partial, the deprecation message is thrown even when the component has been configured. 

### Expected behavior
When the component has been configured, the deprecation message should not be thrown.

hudajkhan:
Closes #3200 .

What this PR does:
Updates the code to only throw a deprecation warning if the exhibit navbar component is not set in the configuration.


jcoyne (approved):

jcoyne:
Do we also want to avoid calling the exhibit_navbar partial if the component is defined?

hudajkhan:
@jcoyne not sure about that just yet, given how it is called from the masthead file: https://github.com/projectblacklight/spotlight/blob/main/app/views/shared/_masthead.html.erb#L11 . I think we would have to rewrite this part as components as well first.

PR #3203, opened 2024-10-17
Update language spec to ruby 3.1

jcoyne:
Autocorrect rubocops.

### DIFF
--- a/blacklight-spotlight.gemspec
+++ b/blacklight-spotlight.gemspec
@@ -18,7 +18,7 @@ these collections.)
 
   s.files = Dir['{app,config,db,lib,vendor}/**/*', 'Rakefile', 'README.md', 'LICENSE', 'spec/{factories,fixtures}/*', 'spec/support/**/*']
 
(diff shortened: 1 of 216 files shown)


blacklight-spotlight.gemspec
```diff
@@ -18,7 +18,7 @@ these collections.)
 
   s.files = Dir['{app,config,db,lib,vendor}/**/*', 'Rakefile', 'README.md', 'LICENSE', 'spec/{factories,fixtures}/*', 'spec/support/**/*']
 
-  s.required_ruby_version = '>= 2.6'
+  s.required_ruby_version = '>= 3.1'
```
corylown, line 21: Does this imply a major version bump? I think we're close to starting work toward Spotlight v5, but are still adding features and fixing bugs on v4 for now.

jcoyne: I dont't think so.  All of the other Ruby's are EOL.  It's long been the practice of the BL project to not support out of support Ruby.

corylown (approved):

PR #3204, opened 2024-10-17
Add a test to check if the Sir Trevor icons loaded

taylor-steve:
I think we should add a test for checking if the Sir Trevor icons were successfully loaded. They are currently broken.

### DIFF
--- a/spec/features/javascript/block_controls_spec.rb
+++ b/spec/features/javascript/block_controls_spec.rb
@@ -25,6 +25,12 @@ describe 'Block controls' do
     # click to add widget
     click_add_widget
 
+    # Check if the Sir Trevor icons are loading. They are in the same SVG so a single check should be sufficient.
+    within('.st-block-replacer') do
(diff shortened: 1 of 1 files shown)


dnoneill:
@taylor-steve ready for a rebase.

spec/features/javascript/block_controls_spec.rb
```diff
@@ -25,6 +25,12 @@
     # click to add widget
     click_add_widget
 
+    # Check if the Sir Trevor icons are loading. They are in the same SVG so a single check should be sufficient.
+    within('.st-block-replacer') do
+      href_value = find('use')['xlink:href']
+      expect(href_value).not_to eq('undefined#add-block')
```
corylown, line 31: Would it be better to check for the presence of `*.svg#add-block`?

taylor-steve: Yes, that's better.

corylown (approved):

PR #3205, opened 2024-10-17
have csv export template use slug for custom fields to match import

### LINKED ISSUE #2994: CSV loader does not convey data for exhibit-specific fields into Solr
### Summary
The "Add items" function includes a tab "Upload multiple items" which allows you to load in a CSV file. If there are exhibit-specific fields, these are given as column headings in the CSV template provided by that function, but any data provided in those columns does not get stored or indexed in Solr.

### Steps to reproduce
On a minimally-customised Spotlight application based on the latest Spotlight engine, create an exhibit, and in the Metadata area of the dashboard create a new field. For example, you might use the Label "Subject". Depending on whether you tick the "Controlled vocabulary" option, the field subject_tesim or subject_ssim will be created.

You can then go to "Upload multiple items" and download the template, which will include the new field.

If you load the CSV file with data in that column, the data appears in the spotlight_resources record, serialised in the "data" field, but it does not appear in spotlight_solr_document_sidecars, and is not stored or indexed in Solr. This means it is not available to be shown in the results.

If we find one of the records uploaded in this way, and click Edit, we get a form with the Subject field available. If we enter data in there and save it, the information is correctly stored in spotlight_solr_document_sidecars and Solr, and will appear in the results.

Fields which have been configured globally across all exhibits by defining them in config/initializers/spotlight_initializer.rb are not affected in this way: you can load those via the CSV loader without any trouble.

dnoneill:
closes #2994

### DIFF
--- a/app/controllers/spotlight/resources/csv_upload_controller.rb
+++ b/app/controllers/spotlight/resources/csv_upload_controller.rb
@@ -36,7 +36,7 @@ module Spotlight
       end
 
       def data_param_keys
-        Spotlight::Resources::Upload.fields(current_exhibit).map(&:field_name) + current_exhibit.custom_fields.map(&:field)
+        Spotlight::Resources::Upload.fields(current_exhibit).map(&:field_name) + current_exhibit.custom_fields.map(&:slug)
       end
 
       # Gets an IO-like object for the CSV parser to use.


taylor-steve:
Do you know if this PR will re-introduce the bug that this previous PR was addressing?

https://github.com/projectblacklight/spotlight/pull/2506/files

Just curious because this PR seems to essentially be a revert of that.

corylown:
I think this does re-introduce the bug fixed here: https://github.com/projectblacklight/spotlight/pull/2506
(hinting at a missing test?)

Maybe the fix is to instead export the custom field slugs instead of the  field name when the template is generated?

dnoneill:
I was wondering about this. I am wondering if we want to support both. I could see instances where the solr field is preferred as well.

dnoneill:
Decision was made to update the template download

corylown (approved):

PR #3208, opened 2024-10-18
Remove Postgresql from the Docker build

jcoyne:
It is not a required dependency of Spotlight.

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,3 @@
-DATABASE_URL=postgresql://spotlight_user:spotlight_password@db/spotlight?pool=5
-POSTGRES_DATABASE=spotlight
-POSTGRES_PASSWORD=spotlight_password
-POSTGRES_USER=spotlight_user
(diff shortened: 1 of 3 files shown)


taylor-steve:
@jcoyne this looks right to me, but could you speak to how you're testing this locally? I think a few of us are struggling with how to review some of the Dockerfile changes because Docker in general isn't working. Are you incorporating some of your other, yet unmerged, Dockerfile PRs? I think if we better understood the direction, we'd have an easier time getting these in. Thanks.

jcoyne:
@taylor-steve I'm not sure how you test a removal of something.  Do we need this?  Does it break anything that isn't already broken? Does this make it less complicated?

taylor-steve:
@jcoyne I don't disagree. I'm only looking to remove our confusion by understanding your intent. Thanks for clarifying.

taylor-steve (approved):

PR #3209, opened 2024-10-18
Prefer docker for starting solr

jcoyne:


taylor-steve:
Could we add some method to opt out of using Docker? Or opt in for CI but opt out by default otherwise? 

I don't leave Docker running because it makes my system super slow, so `docker compose version` will succeed but `docker compose up -d solr` will fail without me manually starting Docker.

Also, I know some devs prefer not to use Docker, but still need it installed.

taylor-steve (approved):

PR #3212, opened 2024-10-18
Upgrade sir-trevor to 0.6.6

jcoyne:
Via https://github.com/madebymany/sir-trevor-js/tree/v0.6.6/build


taylor-steve:
@jcoyne in the new propshaft work we've been testing out using Sir Trevor 0.8.2 from NPM. Do you see an advantage to keeping it locally in vendor?

jcoyne:
@taylor-steve No, I was just trying to migrate step by step.  I don't think we want to pile on all the changes at once.

taylor-steve:
Looks like we have a local modification to sir-trevor from https://github.com/projectblacklight/spotlight/pull/3075 that we'd need to deal with.

Is this 0.6.6? Seems like we already have 0.8.2: https://github.com/projectblacklight/spotlight/pull/2914

corylown:
I'm not sure that local modification to sir-trevor is needed.

jcoyne:
Oh, I misunderstood that we already had 0.8.2.

PR #3214, opened 2024-10-18
Precompile assets before testing

taylor-steve:
I think this is better suited here rather than in the generator.


Rakefile
```diff
@@ -39,6 +39,9 @@ task ci: ['engine_cart:generate'] do
   SolrWrapper.wrap(port: '8983') do |solr|
     solr.with_collection(name: 'blacklight-core', dir: 'lib/generators/spotlight/templates/solr/conf') do
       Rake::Task['spotlight:fixtures'].invoke
+      within_test_app do
+        system 'bin/rake assets:precompile'
```
jcoyne, line 43: typically this is done by spec:prepare. I believe this is done automatically by rake spec

jcoyne (approved):

Rakefile
taylor-steve: Thanks Justin

PR #3219, opened 2024-10-18
Remove bundle install step from generator

taylor-steve:


template.rb
```diff
@@ -14,6 +14,11 @@
 gem 'sidekiq'
 gem 'bootstrap_form', /(\d)(?:\.\d){0,2}/.match(bootstrap_version)[1].to_i == 5 ? '~> 5.4' : '~> 4.5'
 
+if options[:asset_pipeline] == 'propshaft'
```
jcoyne, line 17: Isn't propshaft orthagonal to the two gems below.  You can use these gems with Sprockets as well.

taylor-steve: Yes, agreed. The current state is now closer to what I was imagining, but I'm still uncertain about the Rakefile change.

corylown (approved):

PR #3222, opened 2024-10-21
add date selector to analytics dashboard

dnoneill:
Screenshot 2024-10-21 at 12 57 54 PM
Screenshot 2024-10-21 at 12 57 27 PM
Screenshot 2024-10-21 at 12 57 03 PM


app/components/spotlight/analytics/dashboard_component.html.erb
```diff
@@ -1,6 +1,27 @@
...
+    <%= form.label :end_date, I18n.t("spotlight.dashboards.analytics.end_date_label"), class: 'ml-2 ms-2' %>
+    <%= form.date_field :end_date, value: dates['end_date'], min: min_date, max: max_date %>
+  </div>
+  <div class="ml-auto ms-auto">
+    <%= form.submit I18n.t("spotlight.dashboards.analytics.submit_date_label"), class: 'btn btn-primary' %>
+  </div>
+<% end %>
+
 <% if results? %>
 <%= cache current_exhibit, expires_in: 1.hour do %>
```
corylown, line 18: With caching turned on in development the first request to the page gets cached and then none of the date range parameters I apply get reflected on the page. I think you need to change the cache key to account for any parameters applied.

app/components/spotlight/analytics/dashboard_component.html.erb
```diff
@@ -1,6 +1,23 @@
...
+    <%= form.date_field :start_date, value: dates['start_date'], min: min_date, max: max_date %>
+    <%= form.label :end_date, I18n.t("spotlight.dashboards.analytics.end_date_label"), class: 'ml-2 ms-2' %>
+    <%= form.date_field :end_date, value: dates['end_date'], min: min_date, max: max_date %>
+  </div>
+  <div class="ml-auto ms-auto">
+    <%= form.submit I18n.t("spotlight.dashboards.analytics.submit_date_label"), class: 'btn btn-primary' %>
+  </div>
+<% end %>
+
 <% if results? %>
```
corylown, line 17: Doesn't this always trigger a request to GA API even when the view itself is cached?

app/components/spotlight/analytics/dashboard_component.rb
```diff
@@ -9,7 +9,47 @@ class DashboardComponent < ViewComponent::Base
...
+      end
+
+      def date(param_var, default_date)
+        param_value = params[param_var]
+        return default_date unless param_value
+
+        begin
+          Date.parse(param_value)
+          param_value
+        rescue StandardError
```
corylown, line 29: Any reason not to rescue the more specific `Date::Error` instead?

corylown: The other error case I'm seeing is the date is parseable but it's outside the allowable range from GA:

```
Google::Cloud::InvalidArgumentError
3:start_date = 9999-10-10 must be greater than 2015-08-13 and less than 3000-01-01..
```

app/components/spotlight/analytics/dashboard_component.rb
```diff
@@ -9,7 +9,63 @@ class DashboardComponent < ViewComponent::Base
       def initialize(current_exhibit:)
         super
         @current_exhibit = current_exhibit
-        @dates = { 'start_date' => '365daysAgo', 'end_date' => 'today' }
+        @default_start_date = 1.year.ago.to_date.to_s
+        @default_end_date = Time.zone.today.to_date.to_s
```
corylown, line 13: Things seem to get a little wonky if the configured max/min dates fall outside this range. The UI displays the range set here instead of the configured min/max which is probably unexpected.

app/components/spotlight/analytics/dashboard_component.rb
```diff
@@ -9,7 +9,63 @@ class DashboardComponent < ViewComponent::Base
...
+        if params[:start_date] && params[:end_date] && params[:start_date] > params[:end_date]
+          flash[:error] = I18n.t('spotlight.dashboards.analytics.error_heading', date: "#{params[:start_date]} > #{params[:end_date]}")
+          return
+        end
+        update_dates
+      end
+
+      def update_dates
+        date('start_date') if params[:start_date] && params[:start_date] >= min_date
+        date('end_date') if params[:end_date] && params[:end_date] <= max_date
```
corylown, line 34: Should the checks happening here -- that the start and end dates conform to the configured min/max be part of the validation step?

spec/components/spotlight/analytics/dashboard_component_spec.rb
```diff
@@ -39,6 +57,66 @@
...
+
+    it 'has start and end date selectors' do
+      expect(get_min_max('start_date')).to eq({ max: '2024-12-10', min: '2024-10-10' })
+      expect(get_min_max('end_date')).to eq({ max: '2024-12-10', min: '2024-10-10' })
+      expect(rendered).to have_field('start_date', with: default_start_date)
+      expect(rendered).to have_field('end_date', with: default_end_date)
+      expect(rendered).to have_content 'User activity over the past year'
+    end
+
+    it 'has invalid start and end date params' do
```
corylown, line 109: This applies generally to most of the `it` statements here, but I think the tests would be clearer with context blocks that describe the condition followed by an it block that describes the expected behavior, (or just an it block that describes both the condition and expected behavior), such as:

```
context 'when invalid start and end dates are supplied ' do
  it 'displays an error and falls back to the default start and end dates'
```

corylown (changes requested):

lib/generators/spotlight/templates/config/initializers/spotlight_initializer.rb
```diff
@@ -70,13 +70,16 @@
 # (https://console.cloud.google.com/iam-admin/iam -> Service accounts -> click on service account -> keys)
 # c) set ga_property_id below to your site's property id (located in admin -> Property -> Property details upper right hand corner)
 # d) Set the ga_web_property_id. (located in admin -> Data collection and modification -> Web stream details and begins with G-)
+# e) (optional) set ga_date_range. This allows you throttle the dates the user can filter by.
+# ga_date_range values should use a Date object i.e. (Date.new(YYYY, MM, DD)).
```
corylown, line 74: Is this true? It looks like they should be strings like `YYYY-MM-DD` to me.

app/components/spotlight/analytics/dashboard_component.rb
dnoneill: The range set here is a year. The max/min are there to make sure people can't set a date using the drop downs that the admin doesn't want them to set. It doesn't set the default range.

lib/generators/spotlight/templates/config/initializers/spotlight_initializer.rb
corylown: Everywhere else these dates are strings so in the validation step when you try to compare a date object set in the config with the string "date" from the params I think you'll get an error. It also seems critical that the date format set here is 'YYYY-MM-YY' otherwise the string comparisons in the validate step won't evaluate correctly.

app/components/spotlight/analytics/dashboard_component.rb
corylown: But if I set my app's config to a start_date that's less than a year ago when I first load the analytics page it's going to try to show the last year even though that's outside the restricted range I can select within the app. That seems odd.

corylown (changes requested):

corylown (approved):

PR #3235, opened 2024-10-23
Use data-bl-dismiss attribute for blacklight modal

### LINKED ISSUE #3227: Bug: Modal will not close
### Description
This is for the `migrate-asset-pipeline` branch.

The reindexing status modal will not close when clicking the "x".

### Steps to reproduce
* In Curation > Items, click "Reindex items"
* In the Exhibit Dashboard, open the modal by clicking the status link for the job (e.g., "Successful")
* Try to close the modal

corylown:
Fixes #3227 

This fixes the issue described in #3227 for Blacklight 8.


dnoneill:
This didn't work for me locally. Maybe something weird with my local instance because I was also getting a TypeError: dom.showModal is not a function

corylown:
@dnoneill I suspect a blacklight-frontend version mismatch could be the culprit there.

taylor-steve:
I saw the same behavior as @dnoneill on BL7. I will double check the versions next time I rebuild.

taylor-steve:
Confirmed this is an issue specific to the `migrate-asset-pipeline` branch where the test app is inadvertently bundling both BL7 and BL8 JS when building on BL7. With that fixed this PR functions as expected.

PR #3243, opened 2024-10-25
Specify methods as private in Analytics::DashboardComponent

corylown:
Adjusts tests to use fixed dates so we don't need to rely on private`formatted_date` method in the tests.


dnoneill:
@corylown This is a problem and will break tests. The default date is whatever day is being run. So you are setting it for 10/25/2024. It will break tomorrow

corylown:
Thanks @dnoneill. I think if we set the response from Time.now to always return Oct. 24, 2024 then it works.

dnoneill (approved):

PR #3251, opened 2024-11-01
fix view link to work correctly

dnoneill:
Before this fix, when clicking on the link the url will look like this: http://localhost:3000/test-f71f4f09-8669-4b2f-8121-23ff6ceaaa30/catalog/%3Ca%20href=%22/test-f71f4f09-8669-4b2f-8121-23ff6ceaaa30/catalog/7-113%22%3ESHEPARD'S%20ALABAMA%20CITATIONS%3C/a%3E

Screenshot 2024-11-04 at 1 36 00 PM

### DIFF
--- a/app/components/spotlight/document_admin_table_component.html.erb
+++ b/app/components/spotlight/document_admin_table_component.html.erb
@@ -21,7 +21,7 @@
       </div>
 
       <div class="page-links">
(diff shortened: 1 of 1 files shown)


corylown (approved):
Good find

PR #3253, opened 2024-11-05
remove tophat

### LINKED ISSUE #3052: Investigate where tophat gem features are used
Gem hasn't been updated in 7 years :https://github.com/stve/tophat
Can we remove it or replace it?

dnoneill:
closes #3052


app/controllers/spotlight/catalog_controller.rb
```diff
@@ -30,7 +30,6 @@ class CatalogController < ::CatalogController
         if Blacklight::VERSION > '8'
           blacklight_config.show.document_component = Spotlight::DocumentComponent
         else
-          blacklight_config.show.partials.unshift 'tophat'
```
dnoneill, line 33: I moved this into base instead. I think this is an okay way of doing this but I can change it back.

dnoneill: Yeah I was worried about that. It is an easy enough fix to remove the add_document_meta_content from the DocumentComponent. I wasn't sure if the way this was being done was for a reason? Or if this new approach of using base.html.erb made sense.

corylown: It seems a little nicer to keep it with the component, keeping some conditional logic out of the view.

dnoneill: So I tried to keep it in the component and for some reason the component was being hit twice and loaded twice. I couldn't figure out why, so I moved it back into the base.html.erb file. Let me know if you have better ideas of how to approach this.

PR #3260, opened 2024-11-08
Modify "select all/deselect all" button to become an "All" checkbox to save horizontal space

### LINKED ISSUE #3259: Feature: Use All checkbox instead of Select button in metadata configuration to free up horizontal space
Related issue: https://github.com/sul-dlss/exhibits/issues/2624

Currently, the "Select/Deselect All" buttons on the top of the metadata configuration editing page take up more space than a checkbox for selecting or deselecting all the configuration options in that column.

Current page example:
image

Intended modification:
image

hudajkhan:
Closes #3259 

What this PR does:

- Replace "Select all"/"Deselect all" with a checkbox with "All" as a label.
- Checking the checkbox will check all the metadata field checkboxes in that column.
- Unchecking the checkbox will uncheck all the metadata field checkboxes in that column.
- If any checkbox in a column is unchecked, the "All" checkbox is unchecked. If all the checkboxes are checked, the "All" checkbox is checked. 
- Update related tests
- Update the text strings in spotlight.en.yml since we no longer need "Select all" or "Deselect all"
- Add ids to the "All" checkboxes based on which column they are in/type of information in that column.  This prevents odd JavaScript behavior where clicking next to the label of one of the All checkboxes may affect another column


config/locales/spotlight.en.yml
```diff
@@ -710,8 +710,9 @@ en:
         header: Metadata
         instructions: Select metadata fields to display on each type of page. Select a field name to edit its display label. Drag and drop fields to specify the order in which they are displayed.
         order_header: Display and order metadata fields
-        select_all: Select all
+        select_check: All
         view:
+          select_id: item_details
```
corylown, line 715: Does this need to be a translation?

config/locales/spotlight.en.yml
```diff
@@ -710,8 +710,9 @@ en:
         header: Metadata
         instructions: Select metadata fields to display on each type of page. Select a field name to edit its display label. Drag and drop fields to specify the order in which they are displayed.
         order_header: Display and order metadata fields
-        select_all: Select all
+        select_check: All
```
corylown, line 713: Should we leave the translation key the same in case anyone has customized it?

corylown (changes requested):

config/locales/spotlight.en.yml
hudajkhan: I put it back in.

config/locales/spotlight.en.yml
hudajkhan: No : ) I changed that.

config/locales/spotlight.en.yml
corylown: Sorry, I didn't mean keep both. I mean continue to use the existing select_all translation and set the value to "All" for display.

hudajkhan: ok, will make that change.

corylown (approved):

PR #3266, opened 2024-11-13
update content_for check when annotate_rendered_view_with_filenames i…

### LINKED ISSUE #3261: Bug: show view layout breaks when annotate_rendered_view_with_filenames is enabled
### Description
It looks like in newer versions of rails  `config.action_view.annotate_rendered_view_with_filenames = true` is set in development by default. For some reason this causes the layout to preserve space for the left sidebar when it should not.

### Steps to reproduce

1. Set  `config.action_view.annotate_rendered_view_with_filenames = true` in `development.rb`
2. View an item at a path like: `/spotlight/#{EXHIBIT}/catalog/#{DOCUMENTID}`
3. See that there is empty space for a sidebar to the left:

Screenshot 2024-11-08 at 9 47 28 AM

### Expected behavior

Show page should render like this whether `config.action_view.annotate_rendered_view_with_filenames` is on or off:

Screenshot 2024-11-08 at 9 53 11 AM

dnoneill:
…s enabled

content_for?(:sidebar) returns true because the content is 

closes #3261

### DIFF
--- a/app/helpers/spotlight/application_helper.rb
+++ b/app/helpers/spotlight/application_helper.rb
@@ -30,6 +30,13 @@ module Spotlight
       current_site.title.presence
     end
 
+    def content?(field)
+      return content_for?(field) unless Rails.configuration.action_view.annotate_rendered_view_with_filenames
+
+      stripped_content = content_for(field).gsub(/<!-- BEGIN .+? -->/, '').gsub(/<!-- END .+? -->/, '').strip
+      stripped_content.present?
+    end
+
     # Returns the url for the current page in the new locale. This may be
     # overridden in downstream applications where our naive use of `url_for`
     # is insufficient to generate the expected routes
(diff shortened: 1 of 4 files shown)


hudajkhan:
@dnoneill I think we may not even need to check if the flag is set to see if the content block is empty.  It seems fine to not display content that is only HTML comments and nothing else.

@corylown I looked into the component structure a bit more. 
All the partials that are listed in the sidebar section come from Blacklight.  They start with catalog/_show_sidebar.html.erb but that file includes a deprecation warning stating that we should be setting the sidebar component using blacklight_config.show.sidebar_component .  Right now it defaults to Blacklight::Search::SidebarComponent (although I think the show view uses Blacklight::Document::SidebarComponent).  
I tried setting up a Spotlight specific sidebar component and configuring the catalog controller but even with that, it is still using the _show_sidebar.html.erb partial which means we will still have an HTML comment in there.

Perhaps there's a better way to do this in the future with Blacklight 9 when it moves to using the component instead of relying on the partial, but I am not sure what else we can do right now.

hudajkhan (approved):

PR #3273, opened 2024-11-21
backport remove tophat

dnoneill:


corylown:
@dnoneill this is a breaking change for anyone who overrides the base partial. I'm not sure we should backport this to version 4.

PR #3277, opened 2024-11-21
Set items-block class on uploadeditems block widget so layout is correct in Bootstrap 5

corylown:
Text and image should be able to be placed on the left or right side and the text should wrap around the image.

Before
Screenshot 2024-11-21 at 5 38 44 PM
Screenshot 2024-11-21 at 5 38 53 PM
After
Screenshot 2024-11-21 at 5 40 22 PM
Screenshot 2024-11-21 at 5 40 27 PM


taylor-steve (approved):

corylown:
I need to do more work on this so it will work for more than one image.

taylor-steve (approved):

PR #3286, opened 2024-11-22
add alt-text spec

dnoneill:
cleanup in the controller for initial cruft that I missed.


corylown (commented):
Controller specs are no longer recommended by Rails and Rspec and request and/or feature tests would be better.

corylown (approved):

PR #3293, opened 2024-11-25
fix public/private check in bl8

dnoneill:
closes https://github.com/sul-dlss/exhibits/issues/2650


corylown:
@dnoneill does this need fixing on main as well?

corylown:
I still see the broken behavior in exhibits with this change to spotlight.

dnoneill:
@corylown No. It got removed in the propshaft stuff. 

I will double check but this should fix it. Is it possible it isn't working because we need to do a new release to npm?

corylown (approved):

PR #3315, opened 2024-12-04
Update generators to handle Blacklight 8.7.0

taylor-steve:
Blacklight 8.7.0 contains a backport of https://github.com/projectblacklight/blacklight/pull/3340 which alters the ordering of the install generators.

This PR:
* Removes the workarounds we had in place, waiting for that change
* Sets Blacklight 8.7.0 as the minimum required version. After a conversation with @corylown we think that's fine because no version of Spotlight v5 has been released yet.
* Moves to assuming Spotlight will be built using `--css bootstrap`
* Abandons Rails 7.0 as it handles our rails new options differently than subsequent releases

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -44,7 +44,7 @@ jobs:
       RAILS_VERSION: ${{ matrix.rails_version }}
       BLACKLIGHT_VERSION: ${{ matrix.blacklight_version }}
       BOOTSTRAP_VERSION: ${{ matrix.bootstrap_version }}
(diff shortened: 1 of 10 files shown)


jcoyne:
@taylor-steve is there a reason we don't just run the `rails new` with `--css=bootstrap` rather than hacking around it to make it look like we did?

taylor-steve:
@jcoyne Yes. With `--css bootstrap`, something (probably me...) is forcing the installation of jsbundling-rails with esbuild. I can do `rails new bootstrap_test --css bootstrap -j importmap -a propshaft` and get what I expect, so it's clearly a solvable problem. Just not by me, today. I'm sure it'll be obvious tomorrow.

corylown (approved):

PR #3329, opened 2024-12-09
Extract Translations::SubheadingComponent

jcoyne:


corylown (changes requested):
The translation paths don't work from the component:
```
<span class="translation_missing" title="translation missing: en.spotlight.translations.subheading_component.basic_settings.label">Label</span>
```

corylown (approved):

PR #3334, opened 2024-12-10
Switch back to default chrome

jcoyne:


corylown:
Completed via https://github.com/projectblacklight/spotlight/pull/3337

PR #3346, opened 2024-12-11
Extract SaveSearchComponent

jcoyne:
This will enable downstream apps to easily set the css classes on the button


app/components/spotlight/save_search_component.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Spotlight
+  # Displays the "Save this search" button and modal
+  class SaveSearchComponent < ViewComponent::Base
+    def initialize(ibutton_classes: 'btn btn-secondary')
```
corylown, line 6: Fix typo so default button_classes are set:
```suggestion
    def initialize(button_classes: 'btn btn-secondary')
```

corylown (changes requested):

corylown (approved):

PR #3350, opened 2024-12-12
Enable selection of image area for non-embed item widgets (other than uploaded item)

### LINKED ISSUE #3186: Enable image area selection for non-Embed widgets
For widgets that display images statically (i.e., do not afford panning and zooming interactions within an embedded viewer).
- Item Row
- Uploaded Item Row
- Item Grid
- Item Carousel
- Item Slideshow

Add an option `Select image area` to specify the region of interest.
image9
Clicking `Select image area` opens a modal for the cropping interface. Repurpose the existing cropping interface and controls—currently used for the exhibit masthead and thumbnail images. However, unlike the masthead and thumbnail images, the blue area selector should be freeform rather than constrained to a fixed aspect ratio.
image6

hudajkhan:
Closes #3186 

What this pull request does:

- Sets up an image selection link for resources, with a flag indicating if a specific resource type (e.g. embed or uploaded item) should or should not display (this version of) the image selection link.  The link passes along the id of the block item as well as the index of the item (e.g. the block item corresponds to the first item row widget on the page, whereas the index corresponds to the specific item that has been selected from autocomplete)
- Sets up a modal that opens on clicking the image selection link.  
- Updates the crop module and some of the other pages that also use the crop module to enable a consistent mechanism for identifying the IIIF fields needed for the cropping functions.  Also allows for not preserving the aspect ratio of the original crop box for the item widgets (while preserving that functionality for masthead/thumbnail cropping)
- Sets up listeners for opening and saving the modal that are able to transfer information properly for the IIIF fields between the modal and the widget form.

How to test:

- Click edit on the home page. 
- Add an item row (or item grid or item carousel or item slideshow) widget.
- Select an item from the autocomplete
- Click "select image area" under the item
- Select an area in the cropping tool
- Click "save changes" on the modal

The thumbnail should change to reflect the cropping selection. 

- Click save changes on the form.

The home page should now show the image from the added item where the image should display according to the cropped selection.


app/javascript/spotlight/admin/crop.js
```diff
@@ -257,6 +271,7 @@ export default class Crop {
     var url = this.fileInput.data('endpoint')
     // Every post creates a new image/masthead.
     // Because they create IIIF urls which are heavily cached.
+    var data = this.getData();
```
taylor-steve, line 274: I think this version of data is unused. I think it is fetched in the ajax call below.

app/javascript/spotlight/admin/crop.js
```diff
@@ -2,18 +2,26 @@ import { addImageSelector } from 'spotlight/admin/add_image_selector'
...
+      this.iiifManifestField = $('#' + this.formPrefix + '_iiif_manifest_url');
+      this.iiifCanvasField = $('#' + this.formPrefix + '_iiif_canvas_id');
+      this.iiifImageField = $('#' + this.formPrefix + '_iiif_image_id');
+    } else {
+      this.iiifUrlField = iiifFields["iiifUrlField"];
+      this.iiifRegionField = iiifFields["iiifRegionField"];
+      this.iiifManifestField = iiifFields["iiifManifestField"];
+      this.iiifCanvasField = iiifFields["iiifCanvasField"];
+      this.iiifImageField = iiifFields["iiifImageField"];
+    }
```
taylor-steve, line 24: Did you explore extending crop/otherwise breaking this out into a separate class? It seems like we may be able to drop this and one other conditional if we did. Maybe we'd have to create a cropper factory? I'm not sure if that's worth it or not.

app/components/spotlight/select_image_component.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= render Blacklight::System::ModalComponent.new do |component| %>
+  <% component.with_title { t('spotlight.pages.form.select_image') } %>
+  <% component.with_body do %>
+    <div class='p-3 m-2'>
+      <input type="hidden" name="select_image_region">
+      <p class="instructions"><%= help_text.html_safe %></p>
+      <div class="mx-3" id="resource_block_iiif_cropper" data-behavior="iiif-cropper" data-cropper-key="select_image_<%=@block_item_id%>" data-index-id="<%=@index_id%>" data-crop-width="120" data-crop-height="120">
```
taylor-steve, line 7: Can the ID `resource_block_iiif_cropper` be removed? I didn't see it used later.

app/components/spotlight/select_image_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    </div>
+  <% end %>
+   <% component.with_footer do %>
+    <div> 
+      <input type='button' class='btn btn-outline-primary' value='Cancel' data-bl-dismiss='modal'>
+      <input id='saveimage' type='button' class='btn btn-primary' value='Save changes' data-bl-dismiss='modal'>
```
taylor-steve, line 13: Maybe a nit? Does this need to be an ID? I think you use data-behavior someplace else to target your element. In the handler you have `'#blacklight-modal input#saveimage` which sort of implies you don't trust your own ID. If it needs to be an ID, I think something more specific/less prone to accidental collision would be nice.

app/components/spotlight/select_image_component.rb
```diff
@@ -0,0 +1,24 @@
...
+    def render?
+      true
+    end
+
+    def initial_crop_selection
+      Spotlight::Engine.config.thumbnail_initial_crop_selection
+    end
+
+    def help_text
+      t(:'spotlight.pages.form.instructions_html', thing: :thumbnail)
```
taylor-steve, line 21: I don't see thing in the translation string, can we drop it here?

app/javascript/spotlight/admin/crop.js
taylor-steve: Oh, I see, I misread this the first time sorry. You've added `iiifFields` here to accept the data from `getIIIFObject` that gets the main page widget data and transfers it to the cropper in the modal via the constructor.

Could we rewrite the data for `dataCropperDiv` in `CroppableModal#attachModalLoadBehavior` rather than passing this around as a separate argument? We might need another piece of data and to abstract away what the `'#' + this.formPrefix` is doing now. That or maybe a lightweight extension to this class is the way to go. I just think the only mandatory parameter in the constructor being mostly ignored if the second argument is present should likely be untangled.

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,104 @@
...
+      // Get the hidden input field on the main edit page corresponding to this item
+      var thumbnailSaveField = context.iiifInputField(itemIndex, 'thumbnail_image_url', itemElement);
+      var fullimageSaveField = context.iiifInputField(itemIndex, 'full_image_url', itemElement);
+      var iiifTilesource = context.iiifInputField(itemIndex, 'iiif_tilesource', itemElement).val();
+      // Get the region value saved in the modal for the selected area
+      var regionElement = $('#blacklight-modal input[name="select_image_region"]');
+      var regionValue = regionElement.val();
+      // Extract the region string to incorporate into the thumbnail URL
+      var urlPrefix = iiifTilesource.substring(0, iiifTilesource.lastIndexOf('/info.json'));
+      var thumbnailUrl = urlPrefix + '/' + regionValue + '/!400,400/0/default.jpg';
```
taylor-steve, line 94: Should these sizes be driven by config? E.g., https://github.com/projectblacklight/spotlight/blob/40f92e657e4afdbcde9c243f1ba5bc525c57c130/lib/spotlight/engine.rb#L174

I say this with little to no understanding of how these existing work, or even if they do work...

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,104 @@
...
+      var dataCropperDiv = $('#blacklight-modal [data-behavior="iiif-cropper"]');
+      
+      if(dataCropperDiv) {
+        var dataCropperKey = dataCropperDiv.data('cropper-key');
+        var itemIndex = dataCropperDiv.data('index-id');
+        var iiifFields = context.getIIIFObject(dataCropperKey, itemIndex);
+        // The region field is set separately within the modal div
+        iiifFields['iiifRegionField'] = context.setRegionField(dataCropperKey, itemIndex);
+        new Crop(dataCropperDiv, iiifFields, false).render();
+        //context.attachModalSaveHandler(dataCropperKey);
```
taylor-steve, line 25: Remove or uncomment?

app/javascript/spotlight/admin/crop.js
hudajkhan: Thanks. Will remove.

app/components/spotlight/select_image_component.rb
hudajkhan: Yes

app/javascript/spotlight/admin/croppable_modal.js
hudajkhan: Remove

app/javascript/spotlight/admin/blocks/resources_block.js
```diff
@@ -79,13 +90,20 @@ Core.Block.Resources = (function(){
...
-                      <img class="img-thumbnail" src="${(data.thumbnail_image_url || ((data.iiif_tilesource || "").replace("/info.json", "/full/!100,100/0/default.jpg")))}" />
+                    <div class="d-inline-block">
+                      <div class="d-flex">
+                        <div class="checkbox">
+                          <input name="item[${index}][display]" type="hidden" value="false" />
+                          <input name="item[${index}][display]" id="${this.formId(this.display_checkbox + '_' + data.id)}" type="checkbox" ${checked} class="item-grid-checkbox" value="true"  />
+                          <label class="sr-only visually-hidden" for="${this.formId(this.display_checkbox + '_' + data.id)}">${i18n.t("blocks:resources:panel:display")}</label>
+                        </div>
+                        <div class="pic">
+                          <img class="img-thumbnail" src="${(data.thumbnail_image_url || ((data.iiif_tilesource || "").replace("/info.json", "/full/!100,100/0/default.jpg")))}" />
```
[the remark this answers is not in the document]
hudajkhan: Discussion: Replace "/full/!100,100/0/default.jpg" with a function or property call in resources_block instead, so only the method needs to be overridden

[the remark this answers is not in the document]
taylor-steve: You didn't write this. It can wait for a refactor. My fault!

app/javascript/spotlight/admin/croppable_modal.js
taylor-steve: Okay so the Sir Trevor widgets are hardcoded to 400,400 here when they don't have a thumbnail_image_url: https://github.com/projectblacklight/spotlight/blob/9550f54007d23c05b53666ae65cf3a0690198094/app/views/spotlight/sir_trevor/blocks/_solr_documents_block.html.erb#L14

RIIIF has another hardcoded 400,400. Featured Images uses the config I've linked to above in the model. It doesn't seem like there's an easy way to resolve any of this in this issue. Unless @corylown has other ideas, I think @dnoneill is right that there's a refactor needed that is outside the scope of this. I'm happy to stick with the hardcoded values until then.

app/javascript/spotlight/admin/crop.js
hudajkhan: This has been removed now.

app/javascript/spotlight/admin/crop.js
hudajkhan: I tried rewriting the constructor specifically.  Most of the code no longer relies on formPrefix but there is the upload field that does depend on it.  The upload id part is only used on the exhibit/masthead pages and not in the cropping modal. I could also work on changing that to the new method so thoughts welcome.

app/components/spotlight/select_image_component.html.erb
hudajkhan: The id is required by the cropping code, specifically in renderCropperMap in crop.js: this.cropArea.attr('id').  So we will need to keep the id for the code as it currently stands.

app/components/spotlight/select_image_component.html.erb
hudajkhan: Changed this with input from @dnoneill and @taylor-steve

hudajkhan: But it still does use and id. That said, I think we were fine with it for the time being. Also I tried to make the id more understandable "save-cropping-selection"

app/components/spotlight/select_image_component.rb
hudajkhan: Removed

app/components/spotlight/select_image_component.html.erb
taylor-steve: Yes. I see now, thank you for pointing that out.

app/components/spotlight/select_image_component.html.erb
taylor-steve: Thanks!

app/javascript/spotlight/admin/crop.js
taylor-steve: I think this is great, thanks.

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,106 @@
...
+      var dataCropperDiv = $('#blacklight-modal [data-behavior="iiif-cropper"]');
+      
+      if(dataCropperDiv) {
+        var dataCropperKey = dataCropperDiv.data('cropper-key');
+        var itemIndex = dataCropperDiv.data('index-id');
+        //var iiifFields = context.getIIIFObject(dataCropperKey, itemIndex);
+        // The region field is set separately within the modal div
+        //iiifFields['iiifRegionField'] = context.setRegionField(dataCropperKey, itemIndex);
+        new Crop(dataCropperDiv, false).render();
+        //context.attachModalSaveHandler(dataCropperKey);
```
taylor-steve, line 25: I think you might have meant to delete or uncomment some of these?

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,106 @@
...
+      var itemIndex = dataCropperDiv.data("index-id");
+      // Get the element on the main edit page whose select image link opened up the modal
+      var itemElement = $('[data-cropper="' + dataCropperKey + '"]');
+      // Get the hidden input field on the main edit page corresponding to this item
+      var thumbnailSaveField = context.iiifInputField(itemIndex, 'thumbnail_image_url', itemElement);
+      var fullimageSaveField = context.iiifInputField(itemIndex, 'full_image_url', itemElement);
+      var iiifTilesource = context.iiifInputField(itemIndex, 'iiif_tilesource', itemElement).val();
+      // Get the region value saved in the modal for the selected area
+      //var regionElement = $('#blacklight-modal input[name="select_image_region"]');
+      //var regionValue = regionElement.val();
```
taylor-steve, line 92: Delete or uncomment?

app/javascript/spotlight/admin/croppable_modal.js
hudajkhan: yes

app/javascript/spotlight/admin/croppable_modal.js
hudajkhan: Delete

app/javascript/spotlight/admin/blocks/resources_block.js
```diff
@@ -59,7 +60,16 @@ Core.Block.Resources = (function(){
           .map(word => word.charAt(0).toUpperCase() + word.slice(1))
           .join('');
     },
-
+    _itemSelectImageLink: function(block_item_id, doc_id, index) {
+      // If image selection is not possible for this block, then do not show
+      // image selection link
+      if (!this.show_image_selection) return ``;
+      var url = $('form[data-exhibit-path]').data('exhibit-path') + '/select_image?';
```
dnoneill, line 67:
```suggestion
      var url = $('form[data-exhibit-path]').data('exhibit-path') + '/select_image?';
      var url = document.querySelector('form[data-exhibit-path]').dataset.exhibitPath + '/select_image?'
```

app/javascript/spotlight/admin/blocks/solr_documents_base_block.js
```diff
@@ -50,16 +50,21 @@ SirTrevor.Blocks.SolrDocumentsBase = (function(){
...
         '<input type="hidden" name="item[' + index + '][iiif_image_id]" value="' + (autocomplete_data.iiif_image_id) + '"/>',
-      ].join("\n");
+      ];
+
+      // The region input is required for widgets that enable image cropping but not otherwise
+      if(this.show_image_selection) {
+        iiifFields.push('<input type="hidden" name="item[' + index + '][iiif_region]" value="' + (autocomplete_data.iiif_region || "") + '"/>');
+      }
+
+      return iiifFields.join("\n");
```
dnoneill, line 67: I wonder if we can somehow split this into it's own function. I had to do an override of this function for exhibits and the mirador functionality. This change isn't going to be caught since I have overridden it. I wonder if we can make it more seamless so that users don't constantly have to keep up to date with this?

app/javascript/spotlight/admin/crop.js
```diff
@@ -2,22 +2,36 @@ import { addImageSelector } from 'spotlight/admin/add_image_selector'
...
     this.cropTool = $(this.cropSelector);
-    this.formPrefix = this.cropTool.data('form-prefix');
-    this.iiifUrlField = $('#' + this.formPrefix + '_iiif_tilesource');
-    this.iiifRegionField = $('#' + this.formPrefix + '_iiif_region');
-    this.iiifManifestField = $('#' + this.formPrefix + '_iiif_manifest_url');
-    this.iiifCanvasField = $('#' + this.formPrefix + '_iiif_canvas_id');
-    this.iiifImageField = $('#' + this.formPrefix + '_iiif_image_id');
-
+    // Exhibit and masthead cropping preserves aspect ratio, while item 
+    // and other widget related cropping does not preserve aspect ratio in the cropping area.
```
dnoneill, line 14: I understand what this means but I think it might be made more clear by saying that masthead and exhibit cropping requires a fixed width/height for cropping while items don't.

app/javascript/spotlight/admin/crop.js
```diff
@@ -2,22 +2,36 @@ import { addImageSelector } from 'spotlight/admin/add_image_selector'
...
+    // Get the closest form element
     this.form = cropArea.closest('form');
     this.tileSource = null;
   }
 
+  // Return the iiif input element based on the fieldname.
+  // Multiple input fields with the same name on the page may be related 
+  // to a cropper. We thus need to pass in a parent element. 
+  iiifInputElement(inputPrefix, fieldName, inputParentElement) {
+    return $('input[name="' + inputPrefix + '[' + fieldName + ']"]', inputParentElement);
```
dnoneill, line 32:
```suggestion
    return $('input[name="' + inputPrefix + '[' + fieldName + ']"]', inputParentElement);
    return inputParentElement.querySelector(`input[name="${inputPrefix}[${fieldName}]");
```

app/javascript/spotlight/admin/crop.js
```diff
@@ -228,9 +248,12 @@ export default class Crop {
     }
 
     var input = $('[data-behavior="autocomplete"]', this.cropTool);
-    var panel = $(input.data('target-panel'));
-
-    addImageSelector(input, panel, this.iiifManifestField.val(), !this.iiifImageField.val());
+    
+    // Not every page which uses this module has autocomplete linked directly to the cropping tool
+    if(input.length) {
+      var panel = $(input.data('target-panel'));
```
dnoneill, line 254:
```suggestion
      var panel = $(input.data('target-panel'));
      var panel = document.querySelector(input.dataset.targetPanel);
```

app/javascript/spotlight/admin/crop.js
```diff
@@ -280,7 +303,11 @@ export default class Crop {
   }
 
   setUploadId(id) {
-    $('#' + this.formPrefix + "_upload_id").val(id);
+    // This input is currently used for exhibit masthead or thumbnail image upload.
+    // The name should be sufficient in this case, as we don't use this part of the
+    // code for solr document widgets where we enable cropping. 
+    // If we require more specificity, we can scope this to this.cropTool. 
+    $('input[name="' + this.inputPrefix + '[upload_id]"]').val(id);
```
dnoneill, line 310:
```suggestion
    $('input[name="' + this.inputPrefix + '[upload_id]"]').val(id);
    document.querySelector(`input[name="${this.inputPrefix}[upload_id]"]`).id;
```

dnoneill: I would double check all the suggested jquery to js changes.

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,99 @@
...
+    this.attachModalLoadBehavior();
+    // Attach handler for save by checking if clicking in the modal is on a save button
+    this.attachModalSaveHandler();
+  }
+
+  attachModalLoadBehavior() {
+    var context = this;
+    // Listen for event thrown when modal is displayed with content
+    document.addEventListener('show.blacklight.blacklight-modal', function(e) {      
+      var dataCropperDiv = $('#blacklight-modal [data-behavior="iiif-cropper"]');
```
dnoneill, line 16:
```suggestion
      var dataCropperDiv = $('#blacklight-modal [data-behavior="iiif-cropper"]');
      var dataCropperDiv = document.querySelector('#blacklight-modal [data-behavior="iiif-cropper"]');
```

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,99 @@
...
+    var thumbnailUrl = this.iiifInputField(itemIndex, 'thumbnail_image_url', itemElement).val();
+    var region = this.extractRegionField(thumbnailUrl);
+    regionField.val(region);
+    return regionField;
+  }
+  //When editing an existing/saved item, extract region values based on url
+  extractRegionField(iiifThumbnailUrl) {
+    if (iiifThumbnailUrl != null && iiifThumbnailUrl.length == 0) return null;
+
+    var regex = /\/[0-9]+,[0-9]+,[0-9]+,[0-9]+\//;
```
dnoneill, line 38: This seems like a kind of messy way to extract the data. I think if you really need this data you can add it to the crop dom with one line of code here: https://github.com/projectblacklight/spotlight/blob/105986e9bfc48081d7a754398a9eb1a17cb5f787/app/javascript/spotlight/admin/crop.js#L177-L178

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,99 @@
...
+    iiifFields['iiifCanvasField'] = this.iiifInputField(itemIndex, 'iiif_canvas_id', itemElement);
+    iiifFields['iiifImageField'] = this.iiifInputField(itemIndex, 'iiif_image_id', itemElement);
+    return iiifFields;
+  }
+
+  // Field names are of the format item[item_0][iiif_image_id]
+  iiifInputField(itemIndex, fieldName, parentElement) {
+    var itemPrefix = 'item[' + itemIndex + ']';
+    var selector = 'input[name="' + itemPrefix + '[' + fieldName + ']"]';
+    return $(selector, parentElement);
```
dnoneill, line 58:
```suggestion
    return parentElement.querySelector(selector)
```

app/javascript/spotlight/admin/croppable_modal.js
```diff
@@ -0,0 +1,99 @@
...
+        context.saveCroppedRegion();
+      });
+    });
+  }
+
+  saveCroppedRegion() {
+    //On hitting "save changes", we need to copy over the value
+    //to the iiif thumbnail url input field as well as the image source itself
+    var context = this;
+    var dataCropperDiv = $('#blacklight-modal [data-behavior="iiif-cropper"]');
```
dnoneill, line 75:
```suggestion
    var dataCropperDiv = $('#blacklight-modal [data-behavior="iiif-cropper"]');
    var dataCropperDiv = document.querySelector('#blacklight-modal [data-behavior="iiif-cropper"]');
```

app/javascript/spotlight/admin/blocks/resources_block.js
hudajkhan: I started down the path of changing things to (non JQuery) JavaScript, but objects are passed back and forth between modules/functions, and what works for one does not work for the other.  I've set up another branch to clean this up and use straight up JavaScript but I will plan on doing that in another PR.  FYI: After discussion with @dnoneill , this suggested approach seems fine.

app/javascript/spotlight/admin/blocks/solr_documents_base_block.js
hudajkhan: After discussion, we are ok with how this is right now, but we may want to revisit how to make this more streamlined in the future.

taylor-steve (approved):

PR #3353, opened 2024-12-12
Increase reliability of Sir Trevor feature tests

taylor-steve:
This dramatically reduces or eliminates erroneous test failures surrounding Sir Trevor widgets. It does increase the time to run the tests, but I feel that's better than having to restart them 2-4 times. It does this by moving and improving some of our in-test checks/guards/sleeps to the helpers.

I think after this, there is only one major flappy test remaining (in `spec/features/javascript/edit_in_place_spec.rb`).

### DIFF
--- a/spec/features/autocomplete_typeahead_spec.rb
+++ b/spec/features/autocomplete_typeahead_spec.rb
@@ -47,9 +47,11 @@ describe 'Autocomplete typeahead', js: true, type: :feature do
         # Open the multi-image selector and choose the last one
         click_link('Change')
         find('.thumbs-list li:last-child').click
+        sleep 1
         expect(page).to have_css('.leaflet-container', visible: true)
 
         click_button 'Save changes'
+        sleep 1
 
         expect(page).to have_content('The exhibit was successfully updated.')
 
--- a/spec/support/features/test_features_helpers.rb
+++ b/spec/support/features/test_features_helpers.rb
@@ -5,19 +5,17 @@ module Spotlight
     def fill_in_typeahead_field(opts = {})
       type = opts[:type] || 'default'
 
-      # This first click here is not needed in terms of user interaction, but it
-      # makes the test more reliable. Otherwise the fill_in can happen prior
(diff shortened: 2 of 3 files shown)


spec/features/autocomplete_typeahead_spec.rb
```diff
@@ -47,9 +47,11 @@
         # Open the multi-image selector and choose the last one
         click_link('Change')
         find('.thumbs-list li:last-child').click
+        sleep 1
```
jcoyne, line 50: Why do we need a sleep here.  Shouldn't capybara's `default_max_wait_time` be enough?

spec/support/features/test_features_helpers.rb
```diff
@@ -37,12 +35,17 @@ def click_add_widget
       first('.st-block-replacer').click
     end
 
+    def wait_for_sir_trevor
+      expect(page).to have_selector('.st-blocks.st-ready')
+      sleep 1
```
jcoyne, line 40: Why do we need a sleep here?

taylor-steve: In theory we shouldn't. Sir Trevor adds the `st-ready` class right at the end of their initialization. Nevertheless, the rate of false alarms seems to increase without it. I don't have a good sense of if this is a Sir Trevor issue or if the extra second is giving some other source of CI unreliability the time it needs.

jcoyne: This doesn't make sense to me.  How have you proved your assertion that the rate increases without this `sleep`?

jcoyne: What is the current capybara timeout setting for this test?

taylor-steve: This is most of them:

https://github.com/projectblacklight/spotlight/blob/20b51cb9b21e3944f1d283bbe3e2298a032f32aa/spec/features/javascript/blocks/solr_documents_block_spec.rb#L3

PR #3369, opened 2024-12-17
[backport] Test on Rails 8

jcoyne:
Ref #3332


hackartisan:
This looks good, but we know from #3660 that the 4.x build isn't currently passing, so we'll leave this open for now as a reference while working that out

PR #3386, opened 2024-12-20
Pin OSD & restrict Blacklight Gallery

taylor-steve:


jcoyne:
@taylor-steve is this still necessary?

taylor-steve:
@jcoyne yes, it's necessary in the sense that OSD doesn't work on the 4.x branch currently. We'd either need this or to address the asset issues (I need to refamiliarize where that was left off before break). I don't think it's urgent to accept this, I was leaving it as a reminder that we'd need to do something should we need to cut another 4.x release.

jcoyne:
@taylor-steve wouldn't that prevent https://github.com/sul-dlss/exhibits/pull/2773/files from working?

taylor-steve:
@jcoyne Yes. I'm not sure what to do here, regarding sprockets & OSD/Blacklight gallery. Current blacklight gallery includes the OSD-rails dom.js via https://github.com/projectblacklight/blacklight-gallery/blob/50333265441ccf794ca152e46903058da28b3f36/app/assets/javascripts/blacklight_gallery/osd_viewer.js#L1C8-L1C9. That file has an import statement that fails in Spotlight 4.x. I don't see an obvious way to use current Blacklight Gallery with the old OSD (jquery) setup without adding a bunch of manual includes for gallery/OSD.

jcoyne:
>  I don't see an obvious way to use current Blacklight Gallery with the old OSD (jquery) setup

Why would anyone want to do that?  Shouldn't you use the old BL gallery with the old OSD, or switch to the new ones?

taylor-steve:
Closing this because there are paths to using the new versions in Spotlight v4

PR #3400, opened 2025-01-10
migrate secondary-btns to primary

dnoneill:
Stanford is currently migrating away from two different color buttons on the page based on UX designers. Moving away from this might also be preferable to community but we want to check with the community before merging.


jcoyne:
This is really going to be a pain for anyone who has customized based on secondary button.

dnoneill:
No objections from the community

corylown (approved):

PR #3402, opened 2025-01-13
Add config for updated gems

taylor-steve:
Update the v4 installer to work with the newest openseadragon/blacklight-oembed gems.

Currently I think it's best to leave all the sprockets config work out of the OSD gem. If Spotlight v4 still cares about sprockets we can make it do the work, especially since Spotlight v4 is a bit of a special case, deleting `app/javascript/application.js`.

### DIFF
--- a/lib/generators/spotlight/install_generator.rb
+++ b/lib/generators/spotlight/install_generator.rb
@@ -122,6 +122,25 @@ module Spotlight
       generate 'blacklight_gallery:install'
     end
 
+    def configure_osd_for_sprockets
+      return unless defined?(Sprockets)
+
+      append_to_file 'app/assets/javascripts/application.js', "\n//= require openseadragon\n//= require openseadragon-rails/index\n"
+
+      append_to_file 'config/initializers/assets.rb' do
+        <<~CONTENT
+          Rails.application.config.assets.paths << Rails.root.join('node_modules/openseadragon/build/openseadragon')
+          Rails.application.config.assets.paths << Rails.root.join('node_modules/openseadragon/build/openseadragon/images')
+        CONTENT
+      end
+
+      append_to_file 'app/assets/config/manifest.js', "//= link_tree ../../../node_modules/openseadragon/build/openseadragon/images\n"
+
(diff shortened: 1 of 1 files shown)


lib/generators/spotlight/install_generator.rb
```diff
@@ -122,13 +122,38 @@ def add_osd_viewer
...
+      append_to_file 'config/initializers/assets.rb' do
+        <<~CONTENT
+          Rails.application.config.assets.paths << Rails.root.join('node_modules/openseadragon/build/openseadragon')
+          Rails.application.config.assets.paths << Rails.root.join('node_modules/openseadragon/build/openseadragon/images')
+        CONTENT
+      end
+
+      append_to_file 'app/assets/config/manifest.js', "//= link_tree ../../../node_modules/openseadragon/build/openseadragon/images\n"
+
+      inject_into_file 'app/controllers/application_controller.rb', after: 'class ApplicationController < ActionController::Base' do
```
jcoyne, line 139: Isn't this done by https://github.com/sul-dlss/openseadragon-rails/blob/eb8c5510b705a0c9785ad22c982d8de9dfb9c271/lib/generators/openseadragon/install_generator.rb#L32 ?

jcoyne: Furthermore, it seems we already have this here: https://github.com/projectblacklight/spotlight/blob/242e7ee5f418256554a98faf586c2439efc642a4/app/controllers/spotlight/pages_controller.rb#L12

taylor-steve: Yes. Those steps don't run because the prior `append_to_file 'app/javascript/application.js'` fails because Spotlight removes that file earlier in the install. We could guard against the missing file in the openseadragon gem, but I increasingly don't like the idea of "partial" success on gem installers so for this I let Spotlight take care of its own edge case. Curious what your take is.

taylor-steve:
We should decide on https://github.com/sul-dlss/openseadragon-rails/pull/17 first.

PR #3429, opened 2025-02-06
Test on Blacklight 9

### LINKED ISSUE #3430: Bug: Stop using show.partials.insert

jcoyne:
We need to resolve https://github.com/projectblacklight/spotlight/issues/3430 for this to work


hackartisan:
close in favor of #3651

PR #3442, opened 2025-02-14
Add description to heading block widget

### LINKED ISSUE #3434: Feature: Add explanatory text to the Heading widget
### When the Heading widget has been used on a page and an exhibit admin looks at that page in edit mode, the widget should display the following text, with a horizontal line under the text as a visual separator:

This widget produces an H2 heading.

taylor-steve:
Closes #3434
Screenshot 2025-02-14 at 1 38 54 PM

I'm not sure how I feel about this. It's probably fine as a stopgap.

The [heading block template](https://github.com/projectblacklight/spotlight/blob/4e5bb1670c0cc565ec6797dcb38e4bec7a6c8c04/app/views/spotlight/sir_trevor/blocks/_heading_block.html.erb#L1) can be overridden by the host app making this inaccurate unless they also remember to update this description.

Sir Trevor supports multiple heading levels in the default config via:
```js
      headingLevels: [2,3,4,5,6],
      defaultHeadingLevel: 2,
```

We'd need to start using that level data attribute in the template to select the tag.

It does use scribe for [determining the levels automatically](https://github.com/madebymany/sir-trevor-js/blob/8b5807a94f6e4486ee00ed815f898c3eb72ef19d/src/blocks/scribe-plugins/scribe-heading-plugin.js) which seems to be non-functional, at least in Spotlight, and defunct: https://github.com/guardian/scribe

If we accept this and continue to use Sir Trevor for any length of time, I think we should make a new issue to create a view component for the heading block and either allow the user to select the heading level or have our own automatic scheme. Either may require design.


hackartisan:
Agreed to close during PR review ensemble -- the default heading level was changed in another PR and we're pursuing a more comprehensive fix via the sirtrevor overhaul.

PR #3451, opened 2025-02-19
Remove Bootstrap 4 data attributes

corylown:


spec/views/spotlight/roles/index.html.erb_spec.rb
```diff
@@ -22,8 +22,7 @@
       assert_select "tr[data-edit-for='#{admin_role.id}']"
       assert_select 'td', /jane@example.com/
       assert_select 'td', 'Admin'
-      assert_select "input[type='submit'][data-behavior='destroy-user'][data-target='#{admin_role.id}']," \
-                    "input[type='submit'][data-behavior='destroy-user'][data-bs-target='#{admin_role.id}']"
+      assert_select "input[type='submit'][data-behavior='destroy-user'][data-target='#{admin_role.id}']"
```
corylown, line 25: these are not bootstrap targets so we should just use `data-target` not `data-bs-target`

spec/views/spotlight/search_configurations/_search_fields.html.erb_spec.rb
```diff
@@ -27,8 +27,7 @@
   end
 
   it 'has a checkbox to enable or disable fielded search' do
-    expect(rendered).to have_selector 'input[data-behavior="enable-feature"][data-target="#search_fields"], ' \
-                                      'input[data-behavior="enable-feature"][data-bs-target="#search_fields"]'
+    expect(rendered).to have_selector 'input[data-behavior="enable-feature"][data-target="#search_fields"]'
```
corylown, line 30: these are not bootstrap targets so we should just use `data-target` not `data-bs-target`

taylor-steve (approved):

PR #3461, opened 2025-03-21
Ensure we do not set data attributes to the string "undefined"

corylown:
We should not be setting the value of data attributes to "undefined" when there is no value to set.

Screenshot 2025-03-21 at 3 43 28 PM


jcoyne (approved):
Approved, but needs a rebase.

PR #3463, opened 2025-03-24
Fix specs: purl now encodes the slash between the druid and filename in the IIIF identifier

corylown:
Due to this change in purl: https://github.com/sul-dlss/purl/pull/1236/commits/1f3be24d123007383ba373d0c2d12ac0d6769785.

I think this dependency on production purl is due to cases where the solr document fixtures indexed for tests reference production iiif manifests on purl and gets fetched live via javascript.


jcoyne (approved):

corylown:
Agreed. I created an issue for it: https://github.com/projectblacklight/spotlight/issues/3464

PR #3487, opened 2025-07-30
Avoid deprecated method set_slot

jcoyne:


corylown:
Fixed via https://github.com/projectblacklight/spotlight/commit/a4507aae1dbf67459ee3bb30e936763291419a14

PR #3531, opened 2025-08-07
Address rubocop complaints

corylown:


jcoyne:
This seems obsolete. Please reopen if this is a mistake.

PR #3532, opened 2025-08-08
Update for Blacklight 8.12.0 & View Component 4

taylor-steve:
Rubocop + Blacklight 8.12.0/view component 4 changes

### DIFF
--- a/app/models/spotlight/resource.rb
+++ b/app/models/spotlight/resource.rb
@@ -4,6 +4,7 @@ module Spotlight
   ##
   # Exhibit resources
   class Resource < ActiveRecord::Base
+    # rubocop:disable Style/RedundantParentheses
     class_attribute :indexing_pipeline, default: (Spotlight::Etl::Pipeline.new do |pipeline|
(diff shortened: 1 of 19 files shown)


jcoyne:
Most of these changes seem to be related to view_component 4.0, not Blacklight 8.12.

app/models/spotlight/resource.rb
```diff
@@ -4,6 +4,7 @@ module Spotlight
   ##
   # Exhibit resources
   class Resource < ActiveRecord::Base
+    # rubocop:disable Style/RedundantParentheses
```
jcoyne, line 7: Is there a reason not to remove the redundant parenthesis?

taylor-steve: Yes, it changes the behavior

jcoyne: That sounds like a bug in rubocop.  Have you filed an issue?  They are usually responsive to fixing these sorts of things.

Issue #3499, opened 2025-07-31
Feature: Ability to add anchor and internal links for Footnotes and Table of Contents


alundgard:
_Content copied over from https://github.com/sul-dlss/exhibits/issues/1979 and the Spotlight Community Roadmap._

#### Footnotes user story

As an exhibit creator who is publishing exhibits with scholarly essays, I want to be able to add footnotes in the essay that link from the footnote indicator to the footnote itself and back. I would also like lengthy essays to have a table of contents that has anchor links for the page. This is so that scholarly essays can include proper references, and readers can navigate easily across headings, the footnotes, and essay text.

We are producing more Spotlight exhibits with scholarly content. Currently, we don’t have a good way to have readers be able to easily jump to a footnote and back to the running text, or between sections of the essay. Many new exhibits with scholarly essays are being planned.

See current workaround examples (all for footnotes):

- https://exhibits.stanford.edu/watkins 
- https://exhibits.stanford.edu/seaside/feature/david-starr-jordan-m-d-ph-d-ll-d
- https://exhibits.stanford.edu/sunset/feature/eras-1898-1914

#### Accessibility use case

Stanford note: Anchor links are also important for accessibility. Use case: when several images on a feature page contain extensive text, the text must be transcribed but it is too long to include as alt text (because it is bad practice to have lengthy alt text way in excess of 250 characters). Our accessibility solution is to create an additional feature page, and put all transcribed text on the page. But, the text for each image must be separated by a heading indicating what the image is. We can put a link in the widget to the page with transcriptions, but we should be able to link to the heading on the page for the transcribed text for each image.

### Feature details & acceptance criteria

- Ability to add anchor links as `id` attributes on various HTML elements (likely including headers, within-paragraph text spans, and list items).
- Ability to add anchor links within text paragraphs that link to footnotes on the page, and then link back to the text paragraph position.
- Ability to add anchor links within a list to create a Table of Contents linking to section headings on the page.
- Ability to add anchor links within image widget text paragraphs that link to specific headings appearing on a separate feature page.

### Other notes & comments
- Could be resolved by implementing TinyMCE WYSIWYG as a Sir Trevor widget.
- NLM note: Should this feature request be separated into two?  NLM provides a link to OCR/transcripts from a feature page to a new page to display the item.  See "Download" button at https://profiles.nlm.nih.gov/spotlight/kk/catalog/nlm:nlmuid-101584571X121-doc

ebtoner:
[...]

PR #3563, opened 2025-09-18
Remove Github Markup and add RedCarpet markdown renderer with automatic heading ids

ebtoner:
### Description

Advances #3499, adding automatic heading ids for section linking / table of contents

This ticket removes the gem `github-markup`, which served no purpose other than to [run a default version of the redcarpet markdown renderer](https://github.com/github/markup/blob/ebfff7059dcd84c8ec32df138392decfc6544fda/lib/github/markup/markdown.rb#L15). This update instead uses the `redcarpet` gem directly, including an extension to "add HTML anchors to each header in the output HTML, to allow linking to each section" (see [redcarpet readme](https://github.com/vmg/redcarpet) for details).

### DIFF
--- a/app/helpers/spotlight/rendering_helper.rb
+++ b/app/helpers/spotlight/rendering_helper.rb
@@ -3,7 +3,7 @@
 module Spotlight
   module RenderingHelper # :nodoc:
     def render_markdown(text)
-      # Extend Redcarpet renderer to strip images and to add HTML anchors to each heading in the output HTML
+      # Extend Redcarpet renderer to add HTML anchors to each heading in the output HTML
       renderer = Redcarpet::Render::HTML.new(with_toc_data: true)
       # Use Redcarpet to render markdown as html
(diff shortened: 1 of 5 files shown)


app/helpers/spotlight/rendering_helper.rb
```diff
@@ -3,7 +3,11 @@
 module Spotlight
   module RenderingHelper # :nodoc:
     def render_markdown(text)
-      GitHub::Markup.render('.md', text).html_safe
+      # Extend Redcarpet renderer to add HTML anchors to each heading in the output HTML
+      renderer = Redcarpet::Render::HTML.new(with_toc_data: true)
+      # Use Redcarpet to render markdown as html
+      html = Redcarpet::Markdown.new(renderer).render(text)
+      sanitize(html, attributes: %w[href id])
```
corylown, line 10: Is there a reason to use `sanitize` here instead of marking the response from `render` as `html_safe`?

corylown (approved):

Issue #3535, opened 2025-09-03
Accessibility:  Review existing heading issues and propose a plan for updating heading structure


VanessaVenti:
Review the following tickets and propose a plan to update the heading structure:

- https://github.com/projectblacklight/spotlight/issues/3525
- https://github.com/projectblacklight/spotlight/issues/3434
- https://github.com/projectblacklight/spotlight/issues/3473

taylor-steve:
For pages:

What are folk's thoughts about the intent/current usage of the heading block?

Assume that all non-heading block (e.g., Item Row) headings are `h3`, and the heading block is an `h2`. For the sake of this discussion, assume the structural consideration with other headings outside the context of a `page` (collection of blocks) aren't an issue (which they are).

Given a page with the following blocks, where 'Block *' is any non-heading block:
* Heading 1
* Block 1
* Block 2
* Heading 2
* Block 3
* Block 4

Currently that yields a single section of content, which has invalid heading structure:
* h2 - Heading 1
* h3 - Block 1
* h3 - Block 2
* h2 - Heading 2
* h3 - Block 3
* h3 - Block 4

We could treat the heading blocks as inferred section boundaries:
* h2 - Heading 1
  * h3 - Block 1
  * h3 - Block 2
* h2 - Heading 2
  * h3 - Block 3
  * h3 - Block 4

We could treat leading heading blocks differently:
* h2 - Heading 1
  * h3 - Block 1
  * h3 - Block 2
  * h3 styled as h2 - Heading 2
  * h3 - Block 3
  * h3 - Block 4

We could keep the flat structure and differentiate heading blocks by visual style:
* h3 styled as h2 - Heading 1
* h3 - Block 1
* h3 - Block 2
* h3 styled as h2 - Heading 2
* h3 - Block 3
* h3 - Block 4

What is the intent of the heading block? Is that intent inherent to Spotlight or does it vary by implementing institution/exhibit?

alundgard:
Here is a proposal:
- Leave the exhibit title in the masthead as h1.
- Change the page title h1 to h2. 
- Visually style the page title h2 to look more like an h1 (i.e., keep the slightly larger font size).
- If a user has unchecked the "Show title" box for that page, visually-hide the title but still render the title h2 in the DOM.
- If a page doesn't have any title, visually-hide but render a generic "Main content" title h2 in the DOM.
- Leave the within-page headings (h2) and widget headings (h3) as they are.

PR #3570, opened 2025-09-22
Add basic axe core testing for blocks

taylor-steve:
Part of #3535

Adds the `axe-core-rspec` gem as a dev dependency. Axe core has good support for heading ordering rules, so this will start to give us some feedback as we make changes.

I've included these directly in the block specs, sometimes piggy-backing on an existing test, in an attempt to keep the number of new/mostly duplicated feature tests to a minimum. It looks like Blacklight moved away from the `be_accessible` helper and is now using `be_axe_clean` directly so I've done the same.

### DIFF
--- a/blacklight-spotlight.gemspec
+++ b/blacklight-spotlight.gemspec
@@ -51,6 +51,7 @@ these collections.)
   s.add_dependency 'roar-rails'
   s.add_dependency 'signet'
   s.add_dependency 'view_component', '>= 2.66', '< 4'
+  s.add_development_dependency 'axe-core-rspec'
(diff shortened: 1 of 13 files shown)


hackartisan (changes requested):

hackartisan (approved):
👍🏻 This is a good idea for how to demonstrate the issue and enforce its resolution once that resolution is implemented. Until then, it's possible we're adding several pending tests which may remain so indefinitely (at least 1 doesn't look like it's planned work so will probably remain pending). I'm fine with that, but I'd be happier if at least one other person approves before merge.

PR #3580, opened 2025-09-24
Support adding iiif v3 manifests to exhibits

### LINKED ISSUE #3482: Add support for adding IIIF v3 manifests
Example manifests from Columbia:
- https://dlc.library.columbia.edu/iiif/3/presentation/10.7916/d8-aqk2-np13/manifest
- https://dlc.library.columbia.edu/iiif/3/presentation/10.7916/d8-fkvf-hb22/manifest
- Additional objects can be found here: https://dlc.library.columbia.edu/catalog?search_field=all_text_teim&q=bologna

Current behavior when adding these IIIF v3 manifests to spotlight:

Image

Clicking “Add IIIF items” completes and returns the user to the Curation / items list. However, the item is not added, the item list is not updated, and the Columbia material does not appear in the exhibit.

hackartisan:
Allow items to be added to exhibits via iiif v3 manifests.

This PR adds a sort of factory into Spotlight::Resources::IiifService so that it may return either a v2 or a v3 manifest object (via the iiif-presentation gem). A new manifest v3 class is split out and inherits from the previously-existing manifest class. The metadata class is configurable and both manifest classes use the same configuration. So rather than add a new metadata class for v3 I made it more flexible -- it already had notes about handling v3-type json-ld values. However, if we wanted to add an additional configuration for the metadata class, we could split that into 2 different classes.

closes #3482

screenshot of a successfully ingested item in my dev instance:

Screenshot 2025-09-25 at 09-34-36 A Map of the British and French settlements in North America - IIIF V3 - Blacklight

### DIFF
--- a/app/models/spotlight/resources/iiif_manifest/metadata.rb
+++ b/app/models/spotlight/resources/iiif_manifest/metadata.rb
@@ -1,0 +1,158 @@
+module Spotlight
+  module Resources
+    ###
+    #  A simple class to map the metadata field
+    #  in a IIIF document to label/value pairs
+    #  This is intended to be overriden by an
+    #  application if a different metadata
+    #  strucure is used by the consumer
+    class IiifManifest::Metadata
+      def initialize(manifest)
+        @manifest = manifest
+      end
+
+      def to_solr
+        metadata_hash.merge(manifest_level_metadata)
+      end
+
+      def label
+        return unless manifest&.label
+
+        Array(json_ld_value(manifest.label)).map { |v| html_sanitize(v) }.first
+      end
+
+      private
+
+      attr_reader :manifest
+
+      def metadata
+        manifest&.metadata || []
+      end
+
+      def metadata_hash
+        return {} if metadata.blank?
+        return {} unless metadata.is_a?(Array)
+
+        metadata.each_with_object({}) do |md, hash|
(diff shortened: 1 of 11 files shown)


sarah-cul:
After pulling this branch and using the url: [iiif manifest ](https://figgy.princeton.edu/concern/scanned_maps/3c184604-3f83-4b42-a1f5-96e417a9cd39/manifest) I had 56 items imported and didn't notice any errors. It seems to work!

app/models/spotlight/resources/iiif_manifest/metadata.rb
```diff
@@ -0,0 +1,161 @@
...
+module Spotlight
+  module Resources
+    ###
+    #  A simple class to map the metadata field
+    #  in a IIIF document to label/value pairs
+    #  This is intended to be overriden by an
+    #  application if a different metadata
+    #  strucure is used by the consumer
+    #  override with Spotlight::Engine.config.iiif_metadata_class
+    class IiifManifest::Metadata
```
corylown, line 12: Instead of disabling the `Style/ClassAndModuleChildren` cop, would it make sense to move this class up a level and rename it `IiifMetadata` or `IiifManifestMetadata`? It should be okay to rename this class as long as you also change the default configuration for the class that's set in engine.rb. What do you think?

PR #3591, opened 2025-09-29
[blacklight 9] Pass search_state object to SearchService instead of params hash

### LINKED ISSUE #3590: Maintenance: Blacklight 9 SearchService no longer accepts user_params argument
### Description

search_state object must be used instead of user_params

See: https://github.com/projectblacklight/spotlight/blob/6d10822e88ebb0aa2fb9ecf90587d448d067d805/app/controllers/concerns/spotlight/search_helper.rb#L6-L9

corylown:
Fixes #3590

### DIFF
--- a/app/controllers/concerns/spotlight/search_helper.rb
+++ b/app/controllers/concerns/spotlight/search_helper.rb
@@ -3,10 +3,15 @@
 module Spotlight
   # ...
   module SearchHelper
+    # @param [Hash] user_params the query parameters used to reconstitute the search state
+    #          from saved searches. If not provided, will use the search_state from the view context
+    #          or an empty hash if that is not available.
+    # @return [Object] An instance of the configured search service
     def search_service(user_params = respond_to?(:search_state, true) ? search_state.to_h : {})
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/spotlight/search_helper.rb
```diff
@@ -4,9 +4,10 @@ module Spotlight
   # ...
   module SearchHelper
     def search_service(user_params = respond_to?(:search_state, true) ? search_state.to_h : {})
+      search_state = Blacklight::SearchState.new(user_params, blacklight_config)
```
jcoyne, line 7: Can't you just grab the search_state from the view context?  Why do you have to initialize it yourself?

corylown: That change alone isn't enough. Spotlight is relying on the behavior in that `search_service` method elsewhere. And it's tangled up with how search state is being managed to be aware of the current exhibit scope. More work should be done to clean this up, but it's beyond the scope of what I can do right now.

hackartisan:
Agreed to close during PR review ensemble -- we'll reference the issue on another try in the future

PR #3598, opened 2025-10-08
Fix Blacklight version check

corylown:
`Blacklight.version` no longer works.

Screenshot 2025-10-03 at 11 42 50 AM


dl-maura (approved):
Worked in my local

PR #3606, opened 2025-12-04
Remove obsolete browser specific flexbox properties

jcoyne:


app/assets/stylesheets/spotlight/_multi_up_item_grid.scss
```diff
@@ -53,11 +49,6 @@
     img {
       max-width: 100%;
     }
-    -webkit-box-flex: 1;
-    -moz-box-flex: 1;
-    -webkit-flex: 1;
-    -ms-flex: 1;
     flex: 1;
```
taylor-steve:
```suggestion
    flex: 1 1 125px;
```
Do we need to persist the values from the removed `-webkit-flex: 1 1 125px;` from the line below? On my end it seems to make a difference. It looks like the entire `.item-grid-admin` class is unused, added in https://github.com/projectblacklight/spotlight/pull/281/files and since deleted. I'll do a follow up if you don't want to deal with that here.

PR #3618, opened 2026-01-08
Preconfigure the test app to use Solid Queue/Mission Control - Jobs

taylor-steve:
### Description
This pre-configures Solid Queue/Mission Control - Jobs for the test app. This could be a potential quality of life improvement for developers. It's been brought up in the past that the behavior of jobs (e.g., indexing & indexing progress) can be confusing when using the test app for development without an Active Job backend installed & configured.

This is only for the test app in the development environment. It does not impact [template installs](https://github.com/projectblacklight/spotlight?tab=readme-ov-file#installation).

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ The `spec:prepare` task generates css assets for use by the tests. You can delet
 
 Once you are set up, you can also run `rubocop` to enforce consistent coding style.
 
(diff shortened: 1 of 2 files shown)


jcoyne (commented):
I would encourage the community to skip this one.  Although it does successfully show that Spotlight can work with solid queue, it adds more dependencies into testing and generally makes it more complicated.  I think we can all agree that the engine is hard to maintain, so we should strive for simplicity.

PR #3663, opened 2026-05-29
Update search_service to use Blacklight 9's search_state (no user_params)

### LINKED ISSUE #3590: Maintenance: Blacklight 9 SearchService no longer accepts user_params argument

corylown:
Fixes #3590 

Possibly better fix for #3590 over #3591 

Blacklight 9 removed `user_params` from `SearchService#initialize` and now relies on `search_state`. This PR updates `SearchHelper#search_service` to accept a `Blacklight::SearchState` and builds the default from the view context.

`PagesController#search_results` is the only place that passed a hash and it now wraps it in a Blacklight::SearchState first.

Note: `Spotlight::SearchState` is a `SimpleDelegator`, not a `Blacklight::SearchState` subclass, so it fails `is_a?` checks in `SearchBuilder#with` so the `.to_h` conversion in the default value is intentional.

### DIFF
--- a/app/controllers/concerns/spotlight/search_helper.rb
+++ b/app/controllers/concerns/spotlight/search_helper.rb
@@ -3,10 +3,16 @@
 module Spotlight
   # ...
   module SearchHelper
-    def search_service(user_params = respond_to?(:search_state, true) ? search_state.to_h : {})
+    # @param [Blacklight::SearchState] state the search state to use. If not provided, builds one
+    #   from the search_state available in context (controller or model), or an empty state if
+    #   neither is available. Note: Spotlight::SearchState cannot be passed directly because it is
+    #   a SimpleDelegator (not a Blacklight::SearchState subclass), and Blacklight's
+    #   SearchBuilder#with performs an is_a? check.
+    # @return [Object] An instance of the configured search service
+    def search_service(state = Blacklight::SearchState.new(respond_to?(:search_state, true) ? search_state.to_h : {}, blacklight_config))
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/spotlight/search_helper.rb
```diff
@@ -3,10 +3,16 @@
...
   # ...
   module SearchHelper
-    def search_service(user_params = respond_to?(:search_state, true) ? search_state.to_h : {})
+    # @param [Blacklight::SearchState] state the search state to use. If not provided, builds one
+    #   from the search_state available in context (controller or model), or an empty state if
+    #   neither is available. Note: Spotlight::SearchState cannot be passed directly because it is
+    #   a SimpleDelegator (not a Blacklight::SearchState subclass), and Blacklight's
+    #   SearchBuilder#with performs an is_a? check.
+    # @return [Object] An instance of the configured search service
+    def search_service(state = Blacklight::SearchState.new(respond_to?(:search_state, true) ? search_state.to_h : {}, blacklight_config))
```
jcoyne, line 12: This is quite complicated.  Could you instead pass the search_state in every time?

If you do keep this, why not keep the existing search state (if it exists) rather than creating a new one?

PR #3675, opened 2026-06-09
Decouple admin/tabs.js from jquery

jcoyne:


taylor-steve:
I think moving bootstrap to global in rollup is correct, but it breaks some bootstrap functionality in the esbuild test app (dropdowns don't work, facets expand but don't collapse). Moving bootstrap in `package.json` from dependencies to peerDependencies seemed to solve this for me.

taylor-steve (approved):

PR #3680, opened 2026-06-14
Add AGENTS.md

jcoyne:
### Description
So that LLM programming agents understand some context about spotlight.


hackartisan (approved):
We looked at this in PR review ensemble. It seems like a reasonable starting point that might be helpful for some folks. We can always adjust it or remove it if it's not working well.

PR #3686, opened 2026-06-24
Revert "Add AGENTS.md"

hackartisan:
This reverts commit 78ac1b50204e1e7135842c71e0d43d071353d45f.

After some reflection, I want to propose reverting this commit. My reasoning is that the opinions expressed on our call were:
- I don't know if this is useful
- I will have my own agents delete this file
- I wouldn't use this

I'm thinking that's maybe not the level of support that would be appropriate for adding a file, even a documentation file, to our codebase that we'd then need to maintain. I propose we remove it unless / until we have enthusiasm to use and maintain it.


jcoyne:
I'm doing an experiment to see if this has any value. Using Opencode + GLM-5.2

Prompt:
> Rebuild JavaScript bundle

With AGENTS.md
9,059 tokens
10.8s

Without AGENTS.md
9,237 tokens
35s


When the "with AGENTS.md" was tested, the reasoning was:
> The user wants me to rebuild the JavaScript bundle. According to the AGENTS.md instructions, the JavaScript bundle is generated by rollup via npm run prepare, which builds bundles from app/javascript/spotlight/ into app/assets/javascripts/spotlight/spotlight.js{,.esm.js}.
> 
> Let me run npm run prepare.


I do think it is useful in terms of time saved and tokens expended.  Criticism is deserved for low number of iterations, variety of tasks, etc.

tpendragon:
[Recent research](https://arxiv.org/pdf/2602.11988) seems to show that it's largely not useful, and may increase expense and lower success of tasks.

Can't we just document these things? The harnesses are very good at pulling this info if it needs it. I know there are people that generally just delete these as a matter of practice.

Oh, also, this file is a pretty common attack surface now, I've heard several people who delete them as a matter of course.

jcoyne:
@tpendragon I'm fine with deleting this, but I'll certainly be keeping a copy for myself.  I think that study is extremely flawed as it relies on SWE Bench which suffers from severe data contamination (i.e the models are trained on it).  It also indicates you get better testing and custom tool use.

tpendragon:
> @tpendragon I'm fine with deleting this, but I'll certainly be keeping a copy for myself. I think that study is extremely flawed as it relies on SWE Bench which suffers from severe data contamination (i.e the models are trained on it). It also indicates you get better testing and custom tool use.

Fair criticism, and you should definitely run the tools you want, it just doesn't seem the win here is clear. I think a lot of the documentation is good, I just want that documentation to be as available for people as it is for agents, and I think the README would do fine in that way.

tpendragon:
@jcoyne Thanks for engaging with this, if you're cool with it can you approve? I don't really wanna same-org merge this. 😂

hackartisan:
I pulled from this documentation that was new into the README.

PR #3700, opened 2026-08-15
Add Japanese (ja) locale

nakamura196:
Adds a Japanese locale.

### What this changes

- `config/locales/spotlight.ja.yml` (new) — translates the same **20 keys** that the existing `de` / `es` / `fr` / `it` / `zh` locales cover: the contact form, the browse search box, the feedback link, and the honeypot explanation.
- `lib/spotlight/engine.rb` — registers `ja: '日本語'` in `config.i18n_locales`. Without this the locale file has no effect, because `config.i18n.available_locales` is set from that hash.

### Checks

- **Key parity**: compared against `spotlight.de.yml` — same 20 keys, nothing added or omitted.
- **Interpolations preserved**: `%{count}`, `%{browse_query}`, `%{expand_search_url}`, `%{search_size}`, `%{parent_search_count}`, `%{clear_search_url}`, and the markup inside the `_html` keys.
- **Pluralization**: Japanese has no grammatical plural, so `item_count.one` and `item_count.other` carry the same string — the same approach `spotlight.zh.yml` takes.
- **Ordering**: the new entry is placed after `zh: '中文'`, following the existing ordering of `i18n_locales` by display name.
- `bundle exec rubocop lib/spotlight/engine.rb` — no offenses.

### Scope

Deliberately limited to the public-facing subset the other non-English locales use; the admin UI is left in English, consistent with them. Happy to extend it if you would rather have fuller coverage.

The strings were checked against a running Spotlight instance with a Japanese-language exhibit.

### DIFF
--- a/config/locales/spotlight.en.yml
+++ b/config/locales/spotlight.en.yml
@@ -180,6 +180,7 @@ en:
     fr: French
     hu: Hungarian
     it: Italian
+    ja: Japanese
     nl: Dutch
     pt-BR: Portuguese - Brazil
     sq: Albanian
--- a/config/locales/spotlight.ja.yml
+++ b/config/locales/spotlight.ja.yml
@@ -1,0 +1,40 @@
+ja:
+  activemodel:
+    attributes:
+      spotlight/contact_form:
+        email: メールアドレス
+        message: お問い合わせ内容
+        name: お名前
+  helpers:
+    action:
+      cancel: キャンセル
+    submit:
+      contact_form:
+        create: 送信
+        created: 送信しました。ご意見ありがとうございます。
+  spotlight:
+    about_pages:
+      contacts:
+        header: 連絡先
+    browse:
+      search:
+        item_count:
+          one: "%{count} 件"
+          other: "%{count} 件"
+      search_box:
+        label: このカテゴリ内を検索
+        placeholder: 検索…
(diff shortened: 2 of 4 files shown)


jcoyne:
It looks like spec/helpers/spotlight/languages_helper_spec.rb is now failing.

nakamura196:
Thanks — fixed in 13c2541.

Registering `ja` in `config.i18n_locales` was not enough on its own. `LanguagesHelper#add_exhibit_language_dropdown_options` labels each locale with `t("locales.#{locale}")`, and the `locales:` map in `spotlight.en.yml` had no `ja` entry, so the new option had no usable label and the helper spec's `contain_exactly` list no longer matched.

- `config/locales/spotlight.en.yml`: added `ja: Japanese` (alphabetical, after `it`)
- `spec/helpers/spotlight/languages_helper_spec.rb`: added `['Japanese', :ja]` to the expected list

nakamura196:
The helper spec passes now. The one remaining failure — `spec/features/javascript/browse_group_admin_spec.rb:16` — looks unrelated to this change: the last three CI runs on `main` failed in the same area (`browse_group_admin_spec.rb:16`, `browse_group_admin_spec.rb:26`, `feature_page_spec.rb:139`), all Selenium-driven JS feature specs.

Happy to rebase if you'd like a fresh run.

